PR #34, opened 2025-11-28
Bug FIxes

TemplarGeneral:
Some small fixes. Its in the changelog


TemplarGeneral:
Says that spaces are used instead of tabs but i fixed that.

PR #37, opened 2025-11-29
Suggestions for vehicle tweaks and improvements

AngriestBird:
* Adds 2045 Battlecruiser + 2045 Battleship
* Improvements to the Russian OOB 
* Cleaned up MTG references and standardized them
* Reduces the of Kamikaze drones to 45 for base with 15% increase per generation
* Reduces the manpower of Kamikaze drones to 1


MrP0tter (approved):
Looks good to me, totally bird's fault if anything breaks

PR #45, opened 2025-12-07
Initial Zombie Mode

AngriestBird:
# Zombie Game Mode

- Initial zombie mode 
  - Currently, it's the old hardcoded one from the Doomsday series. Will eventually expand it to have more content but this is good enough for now


XCezor (approved):
Good to go remember that game rule

PR #128, opened 2025-12-30
Refactor Equipment References

AngriestBird:
* Refactored equipment references to ensure these are all compatible with the ideas


XCezor (approved):
gtg 2

Issue #99, opened 2025-12-23
Bug in the investment system


WaIter-Heisenberg:
A troublesome bug has been noticed in the investment system. While playing as Poland, I kept receiving investment offers from EU countries. The problem is that, at the moment, I had, for example, level 3 infrastructure and already had 2 active, accepted investment projects for building 2 infrastructure units. So, by the time these projects were completed, there should have been 5 out of 5 infrastructure units in the region. Nevertheless, I continued to receive offers from other countries to build infrastructure in the same region. Accepting them would be pointless because I couldn’t get any more infrastructure, yet I would lose a small amount of money, and the investing country would gain influence over me. This bug occurred not only with infrastructure but also with other buildings.

PR #162, opened 2026-01-02
 Fix issue #99 - the investment system bug

4RH1T3CT0R7:
I updated the AI investment scoring for the 5 main building types:  

1. Civilian Factories (Type 1)
  2. Military Factories (Type 2)
  3. Dockyards (Type 3)
  4. Infrastructure (Type 4) - improved existing logic
  5. Offices (Type 5)

  For each building type, the AI now:
  1. Counts pending projects in that state for that building type
  2. Calculates available slots = current free slots - pending projects
  3. Only proceeds with scoring if there are actual slots available after pending projects complete

  For infrastructure specifically, I also fixed it to properly prevent investment when:
  - Current level = 5 (max)
  - Current level + pending projects >= 5 (would reach max)

Issue #99

REVIEW REQUIRED!
Are there any other game level restrictions that should be considered in this system?

### DIFF
--- a/common/scripted_effects/99_AI_investment_scripted_effects.txt
+++ b/common/scripted_effects/99_AI_investment_scripted_effects.txt
@@ -494,7 +494,7 @@ AI_get_infrastructure_score = {
 		if = { limit = { check_variable = { building_level@infrastructure = 5 } }
 			multiply_temp_variable = { AI_score = 0 }
 		}
-		else_if = { limit = { check_variable = { differential_infra >= 5 } }
+		else_if = { limit = { check_variable = { differential_infra > 4.999 } }
 			multiply_temp_variable = { AI_score = 0 }
 		}
 		else = {


AngriestBird:
No, the system is smart enough to handle everything else. The AI just didn't have a differentiation to determine if someone was adding 3 infrastructure so it would keep doing it in the same because it was trying to use the best choice investment model. This should be really good for spreading out investments farther then where they were now though. 

This was causing the AI to try to overstack states because it was the best place for that building.

common/scripted_effects/99_AI_investment_scripted_effects.txt
```diff
@@ -443,17 +489,22 @@ AI_get_infrastructure_score = {
...
-		}
-		subtract_from_temp_variable = { v1 = v2 }
-		add_to_temp_variable = { AI_score = v1 }
 
-		# Check to ensure they do not invest in infrastructure
+		# Check to ensure they do not invest in infrastructure when at max or when pending projects will reach max
 		if = { limit = { check_variable = { building_level@infrastructure = 5 } }
 			multiply_temp_variable = { AI_score = 0 }
 		}
+		else_if = { limit = { check_variable = { differential_infra >= 5 } }
```
AngriestBird: Does the >= sequence actually work for HOI syntax? I am not familiar with seeing this used anywhere.

4RH1T3CT0R7: definitely, thanks

AngriestBird (approved):

PR #174, opened 2026-01-03
Import Venezuela X from Daniel

AngriestBird:


AngriestBird:
@DanielDobles I see this old fork with a rather large amount of available content improvements for venezuela. Is there any issues with us potentially importing this and fixing out any balance issues? It's quite a bit old so not sure if this is still something you were interested in since I can see you closed the old version way back in November. 

If not we can import it onto a branch and modernize it so the work doesn't go wasted.

PR #188, opened 2026-01-03
Iraqi flag change 

oscylat:
This originally is a submod I made for the mod but I wish my changes to be in the main mod 
All it does is make the Iraqi flag changeable and adds a few historical flags that are already in the mod but in wrong ideologies so i will be happy if you would look into them


Blazer135:
Can you Also show Some Examples of these Flags if Possible i am Really Interested Iraqi here

PR #276, opened 2026-01-09
Shifted Idelogical Powers to On Demand + Fixed an issue with the ruling party effect

AngriestBird:
* Fixed the disable elections confusing developers
* Fixed Vadim in PMR not correctly retaining his place 
* Shifted ideological powers to on demand calls + monthly back up if need be (tentative for now to ensure they're always set albeit a bit delayed)


XCezor (approved):
approve

XCezor (approved):
approve

PR #300, opened 2026-01-13
AI Combat Adjustments Pt. 1 

AngriestBird:


common/ai_navy/taskforce/USA_taskforce_templates.txt
```diff
@@ -1 +0,0 @@
-# Intentionally left blank
```
AngriestBird: Did we wanna replace paths to drop these or did you wanna keep this for now?

PR #348, opened 2026-01-18
Party Changes

ccfox:
Implemented a few parties that were in the loc but not scripted loc and added an entry for every single country (unless I missed one) to have a named ruling party at start.

Additionally I made some small changes to ruling parties for a few nations:
Uganda, the correct ruling party was in the loc, but not set in the history files. I changed it from neutral conservative to neutral autocracy.
Madagascar, changed from western conservative to emerging Left-wing radical.
Somaliland, like Uganda the correct party had loc but wasn't set as the ruling party.
Aceh, again the correct party had loc but wasn't set as the ruling party.
A few small islands had parties that fit better into other ideologies, mostly switching around western ideologies.

A note that for a few nations they operated under a non-partisan system, or had non-partisan ruling administrations; I opted to just go for Pro-Western or Pro-Russian Independents. The exception is Nauru, which seems like it has informal parties - so I used that one instead.

I think I remember seeing somewhere that MD was planning an update within the next few weeks - and since I might be busy I wanted to go ahead and try to contribute this. There might be some errors, but hopefully it works fine.

### DIFF
--- a/localisation/english/MD_subideology_parties_l_english.yml
+++ b/localisation/english/MD_subideology_parties_l_english.yml
@@ -4879,7 +4879,7 @@
  ### USA ###
  USA.conservatism: "£USA_conservatism_small Republican Party"
  USA.conservatism_icon: "£USA_conservatism_small"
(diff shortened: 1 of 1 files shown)


AngriestBird (approved):

AngriestBird:
Just to note we are not updating until May so if you wanna keep doing these go for it!

PR #487, opened 2026-02-06
Lebanese-Armenian Changes

ccfox:
Need a little help, as I can't get the Armenian alternative parties to show up in-game.

I was thinking of ways to solve the issue of an Armenian puppeted Lebanon (happening most often, I'd imagine, as part of Armenia's focus tree) electing parties that would be openly hostile to Armenian domination in Lebanon. This doesn't fully solve the issue, but the inclusion of multiple parties increases the chance of subsequent elections producing pro-Armenian results. Lebanon is one of the few nations in which this is possible, so I figured it would be neat.

I did alter one party as well. I changed the Promise Party, which appears to be a conservative party, out for the Arab Democratic Party - which is a Ba'athist organization with ties to Syria. The Promise Party was in the LWR slot, but is a splinter of a pretty right-wing party and maintains a lot of that platform, with no ties to the left that I can see.

I also briefly considered touching up Syria's parties, but I'm a bit confused by the localization so I left it alone - only altering a few to be in line with Lebanon's sub-sub ideologies.

### DIFF
--- a/common/scripted_localisation/00_subideology_scripted_localisation.txt
+++ b/common/scripted_localisation/00_subideology_scripted_localisation.txt
@@ -13388,8 +13388,8 @@ defined_text = {
 	text = { trigger = { original_tag = LAT date > 2007.8.25 date < 2013.12.15 } localization_key = LAT.liberalism_alt_desc }
 	text = { trigger = { original_tag = LAT date > 2013.12.15 } localization_key = LAT.liberalism_alty_desc }
 	text = { trigger = { original_tag = LBA has_country_flag = LBA_democratic_reform } localization_key = LBA.liberalism_desc }
-	text = { trigger = { original_tag = LEB NOT = { overlord = ARM } } localization_key = LEB.liberalism_desc }
-	text = { trigger = { original_tag = LEB overlord = ARM } localization_key = LEB.liberalism_arm_desc }
+	text = { trigger = { original_tag = LEB NOT = { is_subject_of = ARM } } localization_key = LEB.liberalism_desc }
+	text = { trigger = { original_tag = LEB is_subject_of = ARM } localization_key = LEB.liberalism_arm_desc }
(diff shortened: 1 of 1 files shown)


common/scripted_localisation/00_subideology_scripted_localisation.txt
```diff
@@ -13386,6 +13388,8 @@ defined_text = {
 	text = { trigger = { original_tag = LAT date > 2007.8.25 date < 2013.12.15 } localization_key = LAT.liberalism_alt_desc }
 	text = { trigger = { original_tag = LAT date > 2013.12.15 } localization_key = LAT.liberalism_alty_desc }
 	text = { trigger = { original_tag = LBA has_country_flag = LBA_democratic_reform } localization_key = LBA.liberalism_desc }
+	text = { trigger = { original_tag = LEB NOT = { overlord = ARM } } localization_key = LEB.liberalism_desc }
```
AngriestBird: NOT = { is_subject_of = ARM } will be a better trigger here as OVERLORD can spam errors if you are not a subject.

AngriestBird (commented):
As for the concern you mentioned, if the parties get banned they can't be elected. In this example, is_subject_of is a better trigger.

common/scripted_localisation/00_subideology_scripted_localisation.txt
AngriestBird: What was the concern w/ the Syrian parties?

ccfox: There's a cluster of parties that are all the 'Arab Democratic Party', and that party as far as I can tell (It is the name of a real party) isn't any of those ideologies. There's a comment that says that its due to the focus tree and they are special parties. I'd have to replay Syria and see what exactly they are meant to be.

I suppose I could just ignore them and do only the 'actual' parties, though that would leave an odd gap for whatever circumstances those parties appear in.

PR #524, opened 2026-02-13
Big Robbo proposed UK OOB changes and Name File

BigRobbo919:
This is a custom UK OOB file I have worked on as well as a name list for units. I have sycned the starting Units to the list so new brigades shouldn't duplicate names.

I have based the OOB on the 1998 SDR white paper which outlines the UK as having 2 divisions, One with 3 Armor Brigades and 3 Mobile Brigades. This UK has 3 Armor 3 Mechanised and 3 Mobile Infantry. I have also made the 16th Air Assault Brigade have vehicles (reflecting their real deployments) and have made the 3 Commando Brigade Mobile to make them more useful as a front line division.

Apologies if I haven't got the correct formatting and syntax down but these files do currently work ingame.


AngriestBird:
Yeah WAS and SCO. The idea is we would like to have as many as we can for all of the tags and since you're doing here if you don't mind cooking some up for those that would be great!

Issue #63, opened 2025-12-12
[FLAVOUR] Unit namelists


AngriestBird:
Just a reminder issue that some countries are missing unit namelists. Not critical, but a nice to have. A couple of countries have had their namelists filled.

Requires adding namelists *and* doing some work on the OOBs.

#Europe 
* [x] ALB = "countries/Albania.txt"
* [x] AUS = "countries/Austria.txt"
* [x] BEL = "countries/Belgium.txt"
* [ ] BLR = "countries/Belarus.txt"
* [ ] BOS = "countries/Bosnia.txt"
* [ ] BUL = "countries/Bulgaria.txt"
* [ ] CAT = "countries/Catalonia.txt"
* [ ] CRE = "countries/Crete.txt"
* [ ] CRO = "countries/Croatia.txt"
* [ ] CYP = "countries/Cyprus.txt"
* [x] CZE = "countries/Czech_Republic.txt"
* [ ] DEN = "countries/Denmark.txt"
* [ ] ENG = "countries/United Kingdom.txt"
* [x] EST = "countries/Estonia.txt"
* [x] FIN = "countries/Finland.txt"
* [ ] FRA = "countries/France.txt"
* [ ] FYR = "countries/Macedonia.txt"		#Macedonia
* [ ] GER = "countries/Germany.txt"
* [ ] GRE = "countries/Greece.txt"
* [ ] HOL = "countries/Holland.txt"
* [ ] HUN = "countries/Hungary.txt"
* [ ] ICE = "countries/Iceland.txt"
* [ ] IRE = "countries/Ireland.txt"
* [ ] ITA = "countries/Italy.txt"
* [ ] KOS = "countries/Kosovo.txt"
* [x] LAT = "countries/Latvia.txt"
* [x] LIT = "countries/Lithuania.txt"
* [ ] LUX = "countries/Luxembourg.txt"
* [ ] MLT = "countries/Malta.txt"
* [ ] MLV = "countries/Moldova.txt"
* [ ] MNT = "countries/Montenegro.txt"
* [ ] NCY = "countries/Northern Cyprus.txt"
* [x] NOR = "countries/Norway.txt"
* [ ] POL = "countries/Poland.txt"
* [ ] POR = "countries/Portugal.txt"
* [ ] ROM = "countries/Romania.txt"
* [ ] SCL = "countries/Sicily.txt"
* [ ] SCO = "countries/Scotland.txt"
[...]

PR #534, opened 2026-02-14
Proposed unit name file for Wales and Scotland

BigRobbo919:
As requested I knocked up a couple unit name files for Scotland and Wales, both have been tested and work in game.

Testing did expose a weird bug where a part of the Sweden Unit name file was missing an argument that made it avaliable to released nations. I have added the correct : " for_countries = { SWE } " argument to the syntax.

Both unit names reflect how I imagine a newly independent Scotland and Wales might identify their military brigades. With Scotland focusing on Castles and Clan names. While Wales focuses on Welsh location names. I did add a bit of a gag name that I am happy to remove :

11th Cymru Infantry (Llanfairpwllgwyngyllgogerychwyrndrobwllllantysiliogogogoch)

These changes have been added to the Changelog too.

### DIFF
--- a/Changelog.txt
+++ b/Changelog.txt
@@ -338,6 +338,7 @@ v2.0.0
   - Fixed an issue where SHORAD sites were never actually being properly factored into your Military Spending
   - Fixed an issue with Iranian Aid not properly giving you income as it should for the second tier of the idea
   - Fixed the USA focus "Army Knowledge Priority" not properly giving generals army size (replaced w/ giving 6 general skilled staffer that do not have it + increased the personnel cost by 1%)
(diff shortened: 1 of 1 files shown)


AngriestBird:
Thank you, my friend! These are greatly appreciated!

AngriestBird:
If you are bored @BigRobbo919 we have a whole list of these that could use some refinement: https://github.com/MillenniumDawn/Millennium-Dawn/issues/63

No pressure at all but you can see some of the tasks we have pending either in there that could be nice haves, but like I said absolutely no pressure. Thank you for the ones you've already given us!

PR #536, opened 2026-02-15
Added UK Scientists

BigRobbo919:
I've added 5 potential Scientists to the UK Pool for the start of the game, These will need to be added to the country history file with a recruit command.

They have been given levels ranging from 2 - 3 and traits that should suit them.

4 Of the Scientists are based on approximate 2000s experts who might have fulfilled specific roles.

I reused Michael Boyce who also serves as an Advisor for the Navy.

All scientists have portraits, large and small.

Testing shows them ingame but I have no idea if they balance with the current mod direction.

### DIFF
--- a/common/characters/ENG.txt
+++ b/common/characters/ENG.txt
@@ -1,35 +1,6 @@
 characters = {
-	ENG_michael_walker = {
-		name = "Michael Walker"
-		portraits = {
-			army = {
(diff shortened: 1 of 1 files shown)


BigRobbo919:
Follow up I noticed I made a syntax error and duplicated some of the Scottish Names into the Welsh file when I made them earlier have fixed and re added

TemplarGeneral (approved):

PR #556, opened 2026-02-18
Fix missing CSS in wiki

kotru21:
Added Ruby setup and dependency installation for Jekyll build.


kotru21:
I think it'll do the job

PR #582, opened 2026-02-20
Fix Armenia/Artsakh integration bugs

4RH1T3CT0R7:
- **Bug 1**: Fixed `ARM_has_artsakh_idea` persisting after NKR annexation. The cancel condition now also triggers when NKR no longer exists as a country. Added `remove_ideas = ARM_has_artsakh_idea` as a safety net in all 9 Karabakh negotiation event options and the `NKR_armenia_annex_nkr` decision where ARM annexes NKR.
- **Bug 2**: Added `ARM_open_borders_with_AZE` decision and `karabakh.30`/`karabakh.31` events, allowing Armenia to propose opening borders with Azerbaijan after the Karabakh conflict is resolved (`karabakh_regulated_flag`). Includes English and Russian localisation.

### DIFF
--- a/common/decisions/Armenia.txt
+++ b/common/decisions/Armenia.txt
@@ -1359,6 +1359,34 @@ ARM_ministry_of_external_category = {
 			factor = 355
 		}
 	}
+
+	ARM_open_borders_with_AZE = {
+		icon = GFX_decision_negotiate_icon
+		cost = 50
+		fire_only_once = yes
+		days_remove = 30
+
+		visible = {
+			has_country_flag = karabakh_regulated_flag
+			country_exists = AZE
+			NOT = { has_war_with = AZE }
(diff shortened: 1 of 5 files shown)


common/decisions/Armenia.txt
```diff
@@ -1359,6 +1359,34 @@ ARM_ministry_of_external_category = {
...
+	ARM_open_borders_with_AZE = {
+		icon = GFX_decision_negotiate_icon
+		cost = 50
+		fire_only_once = yes
+		days_remove = 30
+
+		visible = {
+			has_country_flag = karabakh_regulated_flag
+			country_exists = AZE
+			NOT = { has_war_with = AZE }
```
AngriestBird: redundant check here. If you want the check to be visible then you can just omit this one

AngriestBird (approved):

PR #585, opened 2026-02-20
Name Files for Canada and UK

BigRobbo919:
I've added historic ships to the UK Name pool for Type 26 and 31 including Pennant Numbers. The same for Carriers, LPDs, Type 45, Astute Subs and Patrol Boats. (also fixed a slight issue that the last Type 23 in Production used a different name format to the already built ships)

I created a unit name file for Canada and linked it to the OOB so existing units draw from that file and new units players recruit will draw from these pools. Names no longer have the "(Reserve)" tag since the initial brigade being reserves makes this redundant and allows players to upgrade them to frontline brigades without losing the flavor.

Changelog updated to reflect this.


BigRobbo919:
Ok I don't want to step on the toes of the guys working on the UK but I noticed that Type 32 was refered to as "Duke Class" this clashes with the Type 23 Duke Class. I figured this might be slightly confusing to some players so I have uploaded a proposed Type 32 Guardian Class with two names that match this.

This is my last proposed change for a little while I plan to go back to working on some name files. I can as mentioned to Bird earlier, add some potential names for UK ships to the name file but wanted to ask the folks working on this side of things how they want me to do it.

I can leave "Historic Names" to ships that really existed and either create new lists for potential ships or add ship names to existing lists e.g. Counties or Mythic Figures/

KianGhk1530 (dismissed):

AngriestBird:
@BigRobbo919 Can you fix the conflict and can you make an issue for all your feedbacks that you're not directly suggesting or providing a fix for it?

PR #592, opened 2026-02-22
New og images

kotru21:
Yoo guys wdyt about new auto-generated OG images?
image
WIP


kotru21:
So ok I came up w/this
image

PR #699, opened 2026-03-06
Fix energy load sharing negative consumption when recipient has surplus (#690)

### LINKED ISSUE #690: [BUG] Energy Load Sharing | Energy in consumption negatives.
**Describe the bug**

> I don't know if the energy consumption should be -1.28 like it says, I am playing on the 2.0.0V mod and didn't know where to post this, but I think its from the energy contracts i made with Latvia beacuse when I gave them energy it worked but after they didn't need anymore it just went into the negatives.

Doolittle suggested a basic clamp but I'll do some digging to see if I can fix the root issue just in case.

[Link to Discord](https://discord.com/channels/210890013334831104/1477017537874231397)

**To Reproduce**
Steps to reproduce the behavior.

**Screenshots**

**Operating System**

- [ ] Windows
- [ ] Linux
- [ ] Mac

**Additional context**

AngriestBird:
## Summary

- Fixes #690 — Energy consumption going negative when an energy contract recipient (e.g. Latvia) no longer needs energy

## Root Cause

In `calculate_energy_load_sharing`, the inner loop sets `energy_lack` to the recipient's `energy_difference_variable`. When that value is **positive** (recipient has a surplus), `energy_load_sharing_gain_var` was being set to a **negative** value (after multiplying by -1).

That negative gain flows directly into the `energy_dynamic_modifier`'s `energy_gain` field, causing the displayed energy consumption to go negative.

## Fix

Clamp `energy_lack` to `max = 0` before using it. This ensures only actual deficits are processed — if a recipient has a surplus, their gain is 0 rather than negative.

```
clamp_temp_variable = { var = energy_lack max = 0 }
```

## Test plan

- [ ] Set up an energy load sharing contract with a neighbour
- [ ] Play until the recipient's energy balance improves to a surplus
- [ ] Confirm energy consumption display no longer goes negative for the recipient


XCezor (approved):
okie

PR #726, opened 2026-03-08
Localization Tweaks and minor Druze Flavor

ccfox:
A few things changed for consistency and added here and there!

I gaze Druze a few names based on existing paramilitary and militia groups, with the naming convention for militias based on the existing National Guard group. I think leaving everything else in-line with Syria seems appropriate to me. I also added a party for the Druze State.

Also I notice the conditions for the neutral communist description for Tajikistan (which didn't even exist until I ported over the existing emerging communist description) is very...strange. Is there any reason why its like that or is it safe to remove?

### DIFF
--- a/localisation/english/MD_subideology_parties_l_english.yml
+++ b/localisation/english/MD_subideology_parties_l_english.yml
@@ -1931,7 +1931,7 @@
 
  #Salafist
  PER.Kingdom: "£BLC_jondollah (PRMI) Jundallah"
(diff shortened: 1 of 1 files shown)


AngriestBird:
Can you clarify on the TAJ issue? I am good with the changes though.

AngriestBird:
I'm going to guess for some national content. If you search the flag there is an event.

ccfox:
> I'm going to guess for some national content. If you search the flag there is an event.

Fair enough, I just found it odd! If its fine I'm fine to leave it be. There's an actual description now for it.

PR #778, opened 2026-03-14
New pages structure

kotru21:
Now every page have a new short description component

image

image

image


kotru21:
Gazillion changes because of prettier

kotru21:
Decided to move lib to shatre/lib bcuz it's more familiar to me as a react dev. Also, its more like a FSD approach. Also added a little bit of fixes related to wcag recommendations

kotru21:
> @kotru21  Could you make the dev diary sorting the opposite so the newest is always on top?
> image
> 

Yeah, I'll do it in a few hours

PR #800, opened 2026-03-16
Laos, Indonesia, Minor Improvements

ccfox:
Changed around parties for Indonesia and added new ones - some parties were 100% in the wrong slot, like the former far-right dictatorship's main political vehicle Golkar was in Socdem. The Shia party is kind of a stretch, since its technically just an advocacy group. There are allegedly parties or political orgs in both Malaysia and Indonesia, but we must endure the nebulousness of thinktanks and their ominous chittering.

I added full parties for Laos as well, with descriptions as well. I was going to do Malaysia, but that'll wait because Github has seemingly decided that I am the target for the totality of its wrath.

This is a draft because I have no clue what is even going on anymore. Github demanded I pull from the main - which included Algeria's changes. It then forced me to manually go through and copy-paste every individual change in these two files (I even found bugs like random removed parties such as Iraq's occupation government and some extra spaces where they weren't needed) only to now tell me that I seemingly hallucinated that entire thing and that Algeria was never changed. I do not have the energy at the moment to unfix all the fixes I made it make it happy - so it'll be a draft just to show that I have added things to some nations.


ccfox:
Can't do the discord, sorry. I'll have to go through it at some point and fix everything and then try to figure out why it pulled Algeria's changes.

ccfox:
@AngriestBird Alright, should be fixed now. This doesn't include the formatting fixes or the restoration of missing parties (still unclear if those were removed by accident or not - such as Iraq's occupation government, which is still in the localization files but now removed from scripted loc - and one other that I can't remember now).

PR #971, opened 2026-04-05
crash error fix

Raun139:
I had Codex analyze the crash, and these are the fixes it produced.

Below is Codex’s summary:
-------------------
During the April 5, 2026 CTD investigation, the crash dump itself only showed a generic C0000005 access violation, so I traced the last successful scripted action in the logs instead. game.log stops immediately after EU_update_vars_mission fires at 21:00 on 2001-09-01, which led me to EU_update_vars in common/scripted_effects/99_eu_scripted_effects.txt. That effect divides by global.var_EUpopTotal without checking whether the value is greater than zero. If the EU member array is empty or inconsistent, the denominator can become 0, which is a plausible crash trigger. I added a zero-denominator guard so those divisions only run when global.var_EUpopTotal > 0; otherwise the computed values stay at 0 and the update completes safely.


AngriestBird:
This won't directly fix a crash but supports a theoretical error in the log.

AngriestBird (approved):

PR #985, opened 2026-04-05
Carrier Raids Fix

hhx-ux:
First of all, sorry If I uploaded the file inappropriately. I don't know anything about modding 
-When an air raid is selected from the aircraft carrier, this prevents the planes from being automatically sent to a nearby airbase. As a result, the raid is launched directly from the carrier. 
-Created with the help of AI
-AI can't use this idk why, It is player only fix
-There is localisation problem because in-game raid names appear weird like "raid_type_state_power_strike_air_carrier" etc.
-Raid icon appears as  a question mark so there is also a graphic issue.

### DIFF
--- a/md_carrier_direct_raids/common/raids/carrier_raids.txt
+++ b/md_carrier_direct_raids/common/raids/carrier_raids.txt
@@ -1,0 +1,2304 @@
+types = {
+
+	# =============================================
+	# FUEL RAID - CARRIER DIRECT
+	# =============================================
+	fuel_production_storage_raid_air_carrier = {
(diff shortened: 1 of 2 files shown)


TemplarGeneral:
@hhx-ux Did you manage to find out how to make a game rule? Otherwise we need to close it

hhx-ux:
> @hhx-ux Did you manage to find out how to make a game rule? Otherwise we need to close it

I actually tried but couldn't. 
So it means that it's not gonna be in the mod?

AngriestBird (approved):

PR #990, opened 2026-04-06
Mercosur rework

Raun139:
・Converted MERCOSUR / UNASUL into a joint tree
・Revised the focus tree, including checks for historical consistency, removal of the diplomacy branch, and unlocking decisions
・Added a new invitation/recruitment mechanic
・Added dynamic modifiers that grow as the number of member states increases


common/national_focus/00_generic.txt
```diff
@@ -4,6 +4,10 @@
 	continuous_focus_position = { x = 5 y = 830 }
 
 
+
+
+
+
```
XCezor: what are those empty lines for?

events/mercosur.txt
```diff
@@ -0,0 +1,2733 @@
+﻿add_namespace = mercosur
+add_namespace = mercosur_news
+
+country_event = {
+	id = mercosur.09
```
XCezor: The file starts with id mercosur.09, then jumps to 02, then again to 18. Could you re-order the events in the file? It's a bit chaotic

Raun139:
I fixed the issues that were pointed out.

common/dynamic_modifiers/05_MERCOSUR_dynamic_modifiers.txt
```diff
@@ -0,0 +1,37 @@
...
+	production_speed_industrial_complex_factor = mercosur_member_state_production_speed_industrial_complex_factor
+	production_speed_offices_factor = mercosur_member_state_production_speed_offices_factor
+}
+
+unasul_member_state_modifier = {
+	enable = { always = yes }
+	icon = GFX_idea_unasur_member
+
+	political_power_factor = global.MER_unasul_political_power_factor
+	stability_factor = MER_unasul_member_state_stability_factor
```
AngriestBird: Should this be global ``stability_factor = MER_unasul_member_state_stability_factor``

history/countries/MEX - Mexico.txt
```diff
@@ -8,7 +8,7 @@ capital = 835
 	add_to_array = { global.uncdf_recipients = THIS }
 	add_dynamic_modifier = { modifier = un_aid_subsidies_modifier }
 	set_variable = { gdp_per_capita = 17.943 }
-
+	set_country_flag = mercosur_joint_tree_member
```
AngriestBird: IMO we should covert this to an array so we can replace the any_country = { } calls you do with any_of_scopes = { array = global.mercosur_member_state triggers... }. It is more performance friendly and it's a little simpler=.

common/dynamic_modifiers/05_MERCOSUR_dynamic_modifiers.txt
```diff
@@ -0,0 +1,37 @@
...
+	enable = { always = yes }
+	icon = GFX_idea_mercosur_member
+
+	political_power_factor = mercosur_member_state_political_power_factor
+	stability_factor = mercosur_member_state_stability_factor
+	productivity_growth_modifier = mercosur_member_state_productivity_growth_modifier
+	tax_gain_multiplier_modifier = mercosur_member_state_tax_gain_multiplier_modifier
+	foreign_influence_modifier = mercosur_member_state_foreign_influence_modifier
+	interest_rate_multiplier_modifier = mercosur_member_state_interest_rate_multiplier_modifier
+	total_workforce_modifier = mercosur_member_state_total_workforce_modifier
```
AngriestBird, line 11: Are these intentionally not global. like unasul?

common/ideas/Brazilian.txt
AngriestBird: Are these used as well? I think these might be orphaned since you rebuilt that whole tree

common/decisions/categories/MERCOSUR_decision_categories.txt
```diff
@@ -0,0 +1,56 @@
+MER_mercosur_diplomacy = {
+
+	allowed = { has_country_flag = mercosur_joint_tree_member }
+
+	visible = {
+		has_country_flag = mercosur_joint_tree_member
```
AngriestBird:
```

visible = {
	has_country_flag = mercosur_joint_tree_member
	OR = {
		has_dynamic_modifier = mercosur_member_state_modifier
		has_dynamic_modifier = unasul_member_state_modifier
		AND = {
			is_ai = no
			is_subject = no
			OR = {
				AND = {
					NOT = { has_idea = BRA_idea_associate_states }
					any_country = {
						NOT = { tag = ROOT }
						has_country_flag = mercosur_joint_tree_member
						has_dynamic_modifier = mercosur_member_state_modifier
						is_subject = no
						has_at_least_regional_power_status = yes
						has_global_flag = MER_mercosur_requests_enabled
					}
				}
				AND = {
					has_idea = BRA_idea_associate_states
					any_country = {
						NOT = { tag = ROOT }
						has_country_flag = mercosur_joint_tree_member
						has_dynamic_modifier = mercosur_member_state_modifier
						is_subject = no
						has_at_least_regional_power_status = yes
						has_global_flag = MER_mercosur_full_membership_enabled
					}
				}
				AND = {
					NOT = { tag = MEX }
					NOT = { tag = PAN }
					has_global_flag = MER_unasul_expansion_enabled
				}
			}
		}
	}
}
```

There's a good deal of over complication here I think. WE could probably simplify this quite a bit but we'd need the array. I think the easiest is extracting some of the common logic in some form but not sure exactly how we would do that without the array I mention later in the review.

common/decisions/mercosur_decisions.txt
```diff
@@ -0,0 +1,4144 @@
...
+			effect_tooltip = { custom_effect_tooltip = MER_mercosur_full_invite_reject_tt }
+			custom_effect_tooltip = TT_IF_THEY_ACCEPT
+			effect_tooltip = { custom_effect_tooltip = MER_mercosur_full_invite_accept_tt }
+		}
+
+		ai_will_do = { base = 1 }
+	}
+
+	MER_invite_PRU_associate = {
+		icon = GFX_decision_generic_diplomacy
```
AngriestBird, line 69: Add an ``allowed = { NOT = { original_tag = PREU } }``. This will allow you to get rid of some of the TAG ROOT triggers you do in the visible. I would also suggest reviewing the visible and available. You have a lot of duplication.

I would suggest extracting the logic you think makes sense for it to show up and then when you obviously want a player to see. Right now since they are the same a player will only see if all the conditions are met.

common/decisions/mercosur_decisons.txt
```diff
@@ -0,0 +1,4144 @@
...
+
+		remove_effect = {
+			log = "[GetDateText]: [Root.GetName]: Decision MER_invite_CHL_full_member"
+			country_event = { id = mercosur.05 days = 0 }
+			CHL = { country_event = { id = mercosur.06 days = 1 } }
+			newline = yes
+			custom_effect_tooltip = TT_IF_THEY_REJECT
+			effect_tooltip = { custom_effect_tooltip = MER_mercosur_full_invite_reject_tt }
+			custom_effect_tooltip = TT_IF_THEY_ACCEPT
+			effect_tooltip = { custom_effect_tooltip = MER_mercosur_full_invite_accept_tt }
```
AngriestBird: effect_tooltip isn't needed since you have a custom_effect_tooltip. It only matters if they have an actual in game effect.

common/decisions/mercosur_decisons.txt
```diff
@@ -0,0 +1,4144 @@
...
+				has_idea = BRA_idea_associate_states
+				NOT = { has_dynamic_modifier = mercosur_member_state_modifier }
+				NOT = { has_country_flag = mercosur_chl_accession_pending }
+				is_subject = no
+			}
+		}
+
+		remove_effect = {
+			log = "[GetDateText]: [Root.GetName]: Decision MER_invite_CHL_full_member"
+			country_event = { id = mercosur.05 days = 0 }
```
AngriestBird: Use the short hand on this. No need to define days = 0 when you can just do country_event = mercosur.05

common/decisions/mercosur_decisons.txt
```diff
@@ -0,0 +1,4144 @@
...
+				NOT = { has_country_flag = mercosur_col_accession_pending }
+				is_subject = no
+				has_elections = yes
+				NOT = { has_country_flag = partyall_banned }
+			}
+		}
+
+		target_array = global.mercosur_potential
+
+		target_trigger = {
```
AngriestBird: These are really bad on performance. You are effectively looping through all of the potential mercosur members but they can only ever be one tag. I would suggest leaning entirely into the targeted decisions or make it unique decisions per.

common/decisions/mercosur_decisions.txt
```diff
@@ -0,0 +1,4144 @@
+MER_mercosur_diplomacy = {
```
AngriestBird, line 1: General comment about the decisions and the AI for the focuses. Try to remember to set AI so the AI actually remembers to do these otherwise it'll just do it whenever it maybe ever gets the PP at some distant point with base = 1.

AngriestBird (changes requested):

AngriestBird (approved):

PR #1131, opened 2026-04-23
Feat/cleanup or wrappers

AngriestBird:


common/decisions/categories/99_CHI_decision_categories.txt
```diff
@@ -195,11 +195,9 @@ arunachal_war_decisions_category = {
 			is_claimed_by = CHI
 			is_controlled_by = RAJ
 		}
-		OR = {
-			CHI = {
-				has_completed_focus = CHI_Strike_Across_the_Himalayas
-				owns_state = 475
-			}
+		CHI = {
```
XCezor, line 198: OR should be inside the CHI instead of outside, developer misplaced it (I very think so)

XCezor (approved):
Approved

Blazer135 (approved):

PR #1168, opened 2026-04-29
Strait of hormuz

KianGhk1530:
Added Kharg and Hormuz related content for Iran
Adjusted some stuff for the civil war
Flavor
Fixes


TemplarGeneral (approved):
@KianGhk1530 your add_to_variable doesn't have a tooltip. Is that how you want it? Everything is approved

Issue #1247, opened 2026-05-10
Refactor: deduplicate plane-loss and building-damage blocks across drone strike raids


AngriestBird:
## Background

`common/raids/drone_strike_raids.txt` contains 7 raid types. The first one (`drone_swarm_strikes`) was deduplicated against the new `apply_drone_swarm_damage` scripted effect in `common/scripted_effects/00_raid_scripted_effects.txt`, saving ~130 lines and giving a single point of truth for drone-swarm damage tuning.

Two larger duplication patterns remain across the other 6 raids and are good candidates for the same treatment.

## Opportunity 1: plane-loss block (high confidence, ~480 lines saved)

The `actor_effects` plane-loss flow appears in **5 raids × 4 tiers = 20 nearly-identical blocks**:
- `drone_target_recon`
- `targeted_drone_strike`
- `drone_airfield_radar_strike`
- `drone_power_grid_strike`
- `drone_rocket_site_strike`

Variance is very small:
| Tier | random max | war multiplier | non-war chance max | special handling |
|---|---|---|---|---|
| failure | 8 | 1.5 | 2 | if rolled 2, subtract 1 |
| limited_success | 5 | 1.25 | 1 | none |
| success | 3 | 1.15 | 1 | if not 1, set 0 |
| critical_success | 2 | 0.95 | 1 | none |

Per-raid variance is limited to the non-war PP penalty (`-100` for `drone_power_grid_strike` and `targeted_drone_strike/failure`; `-300` for the rest).

**Suggested approach:** add four per-tier helpers (`apply_raid_failure_losses`, `apply_raid_limited_success_losses`, `apply_raid_success_losses`, `apply_raid_critical_success_losses`) to `common/scripted_effects/00_raid_scripted_effects.txt`, where tier-specific knobs are baked in and the only parameter is `non_war_pp_loss`. Each call site collapses to ~2 lines.

## Opportunity 2: building-damage block (medium confidence, ~400 lines saved)

The `victim_effects` building-damage flow appears in **4 building-strike raids × 3 tiers = 12 blocks**:
- `targeted_drone_strike`
- `drone_airfield_radar_strike`
- `drone_power_grid_strike`
- `drone_rocket_site_strike`

Variance is larger and would need careful per-raid verification:
- Base value: `1.25` vs `1.0`
- Multiplier: 25 / 30 / 40 / 50 / 60 / 70 (varies per raid AND per tier)
[...]

PR #1253, opened 2026-05-10
Deduplicate raid boilerplate (non-drone raids) — refs #1247

AngriestBird:
Refs #1247. Deduplicates ~2200 lines from the raid system by extracting four repeated patterns into shared scripted effects/triggers. Fixes 17+ NOT-AND logic traps in AI modifiers and adds an adversarial-review skill.

Changes:
- Helpers: calculate_aa_defense_value, apply_raid_failure_losses, apply_raid_building_damage_limited/success/critical
- Logic fixes: 17+ NOT-AND traps in AI modifiers; clamped aa_defense >= 1 upstream
- Validator: extended check_common_mistakes.py with raid-specific checks
- Docs: added adversarial-review skill; updated review-branch to use it

Files: 7 raid files, shared scripted_effects/triggers, game_rules, localisation, new lint tool.
Net: -3958/+1867 lines. Drone raids skipped (separate branch).


AngriestBird:
## Code Review — PR #1253 (Deduplicate raid boilerplate)

Solid refactor. ~2200 net lines removed and the helpers are well-documented in the file header. Most call sites are mechanical substitutions and read cleanly. A few items worth addressing or considering as follow-ups.

---

### Bugs / Correctness Issues

**None blocking** — the substitutions are faithful. The PR description's "behaviour normalisation" note is accurate: pulling the `aa_defense = 0 → 1` clamp upstream into `calculate_aa_defense_value` and then unconditionally dividing is safe at every site **except one** I want to flag:

- **`common/raids/cruise_missile_strike.txt` `anti_missile_strikes/limited_success` (around L135–L175)** — this block was intentionally left bespoke (per PR description) for the `rocket_mult = 1.5` scaling. Pre-PR it had an unguarded `divide_temp_variable = { building_damage_by_missile = aa_defense }` (one of the 9 div-by-zero edge cases the PR acknowledges). The PR now routes its building-damage flow through `apply_raid_building_damage_limited`, which still divides unconditionally — **but the bespoke pre-helper block does not call `calculate_aa_defense_value` and is missing the final `if = { limit = { check_variable = { aa_defense = 0 } } add_to_temp_variable = { aa_defense = 1 } }` clamp**. The `aa_defense = 0 → rocket_mult` fallback inside the `rocket_mult != 0` branch is preserved, but in the case where the target state has zero AA buildings AND `rocket_mult` evaluates to 0, `aa_defense` stays at 0 and the helper divides by zero. Same edge case as pre-PR (`anti_missile_strikes` targets `rocket_site > 0` so realistically rocket_mult is rarely 0, but damaged rocket sites push `non_damaged_building_level@rocket_site` to 0). Recommend either:
  - Adding `if = { limit = { check_variable = { aa_defense = 0 } } add_to_temp_variable = { aa_defense = 1 } }` after the bespoke AA composition, or
  - Folding the bespoke `rocket_mult = 1.5` scaling into a parameterised helper

---

### Style / Standards Violations

**None.** Helper file uses tabs consistently, brace style matches surrounding code, comments are placed where invariants are non-obvious (the docblock above each helper documenting required caller-set temp vars is exactly the kind of "why" comment `general-rules.md` endorses).

One nit:

- **`common/scripted_effects/00_raid_scripted_effects.txt` L29–L31** — `calculate_raid_aa_defense` body is now just `calculate_aa_defense_value = yes` + `multiply_temp_variable`. Trivially fine, but worth a comment noting that this exists for actor-side use; the docblock above (L25–L27) covers this already.

---

### Simplification Opportunities

1. **The three building-damage helpers (`apply_raid_building_damage_limited`/`_success`/`_critical`) are identical except for the `missile_luck_mod` random range.** Could be collapsed into a single helper that takes `missile_luck_min`/`missile_luck_max` as caller-set temp vars:

   ```
   # Caller sets missile_luck_min, missile_luck_max, building_damage_by_missile, raid_damage_mult.
   apply_raid_building_damage = {
       set_temp_variable_to_random = {
           var = missile_luck_mod
           min = missile_luck_min
           max = missile_luck_max
       }
       multiply_temp_variable = { building_damage_by_missile = missile_luck_mod }
       divide_temp_variable = { building_damage_by_missile = aa_defense }
       multiply_temp_variable = { building_damage_by_missile = raid_damage_mult }
   }
   ```

   Three helpers → one. Tuning the luck curve becomes a one-liner per call site instead of a helper-rename. Trade-off: 2 extra `set_temp_variable` lines per call site (~96 lines total) vs. ~30 lines of helper duplication today. Net is roughly even, but the single source of truth wins long-term.

2. **The four plane-loss helpers (`apply_raid_failure_losses` / `_limited_success_losses` / `_success_losses` / `_critical_success_losses`)** could collapse similarly if you parameterise `raid_planes_max`, `war_loss_mult`, and the per-tier "dead planes" handling. The fail/critical tiers' special branches (`if = { limit = { check_variable = { random_chance_on_dead_planes = 2 } } subtract... }` and the inverse on `success`) make this trickier than the damage case, but a single helper with a `tier` index + lookup arrays would still be cleaner than four 30-line duplicates. Reasonable to defer this to a follow-up.

3. **Stale `if = { limit = { check_variable = { aa_defense > 0 } } divide... }` guards in `divisions_cruise_missile_strike` and the `canal_bombing_raid` luck-roll sites** (the ones the PR intentionally didn't migrate). Now that `calculate_aa_defense_value` clamps `aa_defense >= 1`, these guards are dead code. Drop them or migrate those sites to a `apply_raid_unit_damage_*` helper. (Not blocking — current behaviour is correct.)

4. **`common/raids/cruise_missile_strike.txt` L135–L139 and similar sites** — the caller does its own `var:target_state = { ... }` block to read `aa_tgt`/`radar_tgt`/`airbase_tgt`, then `calculate_aa_defense_value = yes` opens **another** `var:target_state` internally. Two scope switches to the same state. Minor (var: scopes are cheap once the state is resolved) but a follow-up helper variant that accepts pre-read AA values would avoid the second switch entirely.

---

### Performance Concerns

None. The refactor is net-positive: fewer parsed tokens, identical runtime semantics, no new hot-path overhead. The clamp-before-divide change (helper guarantees `aa_defense >= 1`) is correct per `.claude/docs/performance-patterns.md` "Clamp Before Division".

---

### LGTM Items

- Scope chain `PREV.raid_pp_loss` inside `var:actor_country = { ... }` resolves correctly (confirmed via `common/scripted_effects/00_scripted_effects.txt:4303` precedent).
- Brace balance is preserved (162 open / 162 close in helper file).
- Indentation is tabs throughout, matching project style.
- Helper docblocks explicitly call out caller-set temp-var contract — exactly the kind of invariant-doc that `general-rules.md` says to write.
- `EH_raid_portal` correctly left inline (PR description acknowledges this).
- `anti_missile_strikes` bespoke `rocket_mult = 1.5` correctly preserved (modulo the clamp issue above).
- `divisions_cruise_missile_strike` and `canal_bombing_raid` unit-damage paths correctly left inline (different damage variable, as PR notes).

---

### Recommended Action

1. **Address the `anti_missile_strikes/limited_success` div-by-zero edge case** (add the `aa_defense = 0 → 1` clamp after the bespoke composition). Single-line fix.
2. Consider consolidating the three building-damage helpers as noted above — clean follow-up PR if not in scope here.
3. The four loss helpers consolidation and stale `aa_defense > 0` guard cleanup are good follow-ups but not blocking.

TemplarGeneral (approved):

TemplarGeneral:
@AngriestBird approved but you need to update and then you can merge

PR #1262, opened 2026-05-11
Fix Helsing MIO

JonathanWoollett-Light:
- Event can now trigger based on date and air focus completion (previously it never triggered).
- Logo updated to be larger and more readable within the frame
Screenshot 2026-05-11 193218
image


TemplarGeneral:
@JonathanWoollett-Light Okay so we don't want to use the on_montly like that. You could change the focus so it launches this focus or use the scripted effect yearly effects to let the event launch on the correct year.

AngriestBird (approved):

PR #1345, opened 2026-05-20
Simplify random-construction effects (+ fix), subideology loops, and drought flags

AngriestBird:
## Summary

Performance/simplification pass across scripted effects, plus a correctness fix to the random-construction helpers.

### Random construction effects (`00_scripted_effects.txt`)
The per-quantity building helpers (`one_/two_/three_/four_random_*` and their `_state` fallbacks) were collapsed onto a shared selection pattern. The first attempt used `every_owned_controlled_state` + `random_select_amount = build_qty` — but **`every_owned_controlled_state` is not a real effect, and `random_select_amount` only accepts an integer literal, not a variable.** This is now corrected to `every_controlled_state` with literal amounts and an `is_owned_by` guard to preserve owned+controlled semantics. `replace_civ_with_arms_factories` likewise uses `every_controlled_state` + literal `5`.

### Reusable build-location triggers (`00_scripted_triggers.txt`)
Extracted the duplicated "valid random build state" limit into scripted triggers:
- `free_shared_building_slots` — one check covering every shared-slot building, since they all draw from the same pooled state building slots.
- Per-building triggers for the non-pooled buildings (infrastructure, dockyard, nuclear reactor, air base, radar station, anti-air, internet station, fuel silo, fossil powerplant, renewable energy).

### De-duplication across focuses & decisions
Replaced **48 inline `random_owned_controlled_state` build limits** that exactly matched the pattern with the new triggers, across `Iran`, `mercosur_decisions`, `05_bashkiriya`, `05_india`, `05_nigeria`, `05_tatarstan`, `spain`, and `00_CHI_scripted_effects`. Pure de-duplication — no build or payment effects were changed. Sites that don't match the pattern (e.g. Singapore's `size > 1` check) were intentionally left untouched.

### Subideology steal_popularity loops (`00_subideology_scripted_effects.txt`)
Collapsed repeated `set_temp_variable = { steal_party_index = N }` sequences into `for_each_loop` blocks (`value = _steal_idx`).

### Drought events (`agricultural_events.txt`)
Replaced the four mutually-exclusive `AGRI_*_drought` country flags with a single `AGRI_drought_severity` variable (0–3).

**Net:** ~700 lines added, ~3,500 removed.

### Simplification validator (`tools/validation/validate_simplifications.py`)
New auto-discovered validator that flags inline random-build location limits which exactly match one of the shared triggers, and any use of the non-existent `every_owned_controlled_state` effect. Warnings only, non-blocking. Running it surfaced 8 more matching limits in `events/` (San Marino, Indian, Serbia, Internal Faction Events, Econ_events), now converted to the triggers as well.

## Test plan

- [x] Building rewards (arms/civ factory, infrastructure, dockyard, offices, etc.) construct correctly for player and AI, in home and owned-controlled states
- [x] `replace_civ_with_arms_factories` still converts 5 civilian factories to arms factories
- [x] Deduplicated focus/decision build rewards (mercosur accession, India, CHI infrastructure, etc.) build the same count and charge the same cost as before
- [x] Subideology steal_popularity still targets the correct party indices
- [x] Drought events set and read severity correctly across all four tiers


ducksnacks (approved):
I cant see anything

Blazer135 (approved):

PR #1364, opened 2026-05-21
Britsh oob and production rework 

sayu-fr:
Add Air OOB and rework ground force, missing production, i add him latter


TemplarGeneral:
@sayu-fr is this ready to review now?

Blazer135 (approved):

PR #1384, opened 2026-05-22
Event Horizon Finish p1

XCezor:
## Event Horizon
- Chimera AI is now more aggressive, should push the frontline with more units instead of holding army in the Portals
- You cannot longer setup new Randallite mines if the amount of the Randallite will exceed 300 units
- Finished adding icons for the new focus tree
- Optimalized electric traps decisions
- Reduced Chimera max. division amount from 800 to 750
- Implemented 2 new Chimera units and added them to their battalions
- Fixed Bosphorus Defence Line destruction
- Randallite effects on the country are now reduced
- Mech production speed can no longer be negative
- Fixed United States of North America not detecting Alaskan Portal
- Randallite income from mines is reduced by 1 (so instead of chance to get from 4 to 6, you can get from 3 to 5 units)
- Fixed satellite disruption static screen visibility on larger screen resolutions

## CZE
- Fixed monarchist decisions visibility
- Slightly optimized Petr Pavel's GUI

## POL
- Some communist decisions from ZKP-P path will no longer repeat itself after being completed


common/decisions/categories/EH_decision_category.txt
```diff
@@ -125,8 +125,6 @@ EH_start_scenario_cheat_category = {
 	icon = GFX_decision_generic_arms_trade
 	priority = 1000
 
-	allowed = { has_global_flag = GAME_RULE_event_horizon_enabled }
```
AngriestBird, line 128: Why remove?

common/on_actions/99_EH_on_actions.txt
```diff
@@ -486,6 +477,39 @@ on_actions = {
 
 	on_weekly = {
 		effect = {
+			if = {
+				limit = { ROOT = { has_war_with = ALN } }
```
AngriestBird: We could likely make this more performance friendly if we flip this to an on_weekly_ALN and check every_enemy_country rather then looping for every nation on_weekly in a standard game.

AngriestBird (changes requested):

common/decisions/categories/EH_decision_category.txt
XCezor: This decision has `visible` block which makes it visible only when scenario is enabled so there's no need for additional `allowed` block with similar trigger

common/on_actions/99_EH_on_actions.txt
XCezor: Moved to ALN block

AngriestBird (approved):

PR #1387, opened 2026-05-22
Blazer validator fixes

Blazer135:
Validator Fixes


localisation/english/events_l_english.yml
```diff
@@ -1763,6 +1763,14 @@
  diplomatic_response.4.d: "We knew that some [From.GetAdjective] people are idiots, but their top envoys had to join the rank of idiots in the last few days of negotiations. They defied us, rejected our demands and sent our envoys home.\n\nThese fools."
  diplomatic_response.4.o1: "They will know the consequences."
 
+ # annex_game_rule — hidden game-rule annexation relay events
+ annex_game_rule.1.t: "Annexation Relay Triggered"
+ annex_game_rule.1.d: "Routing annexation to the correct country scope."
```
AngriestBird, line 1768: If the hidden event has titles and descriptions they can be just removed and moved all effects into immediate. No reason to have an option or interactive element or localization for something a player is never going to see.

localisation/english/MD_focus_IRQ_l_english.yml
```diff
@@ -2270,3 +2271,38 @@
  IRQ_war_with_shias: "War with Shias"
  IRQ_western_engineers_in_iraq: "Western Engineers in Iraq"
  IRQ_western_aid_to_iraq: "Western Aid to Iraq"
+
+ # iraqi_MNF hidden state-flag initialisation events (18-25, 250, 251)
+ iraqi_MNF.18.t: "Initialising Hunt for Saddam"
```
AngriestBird, line 2276: As I stated before, any hidden event with options should instead not have localization and then instead have everything in immediate. This is just localization for the sake of localization. Comments should convey the interaction and purpose.

AngriestBird (changes requested):

PR #1418, opened 2026-05-24
Validator suite: false-positive cleanup + performance overhaul

AngriestBird:
#### Validation

- **Eliminated ~270 false-positive validator warnings.** Local pre-commit was producing more noise than signal, pushing contributors into content workarounds rather than fixing the gates. Root causes patched across the suite:
  - `validator_common.py` loc-key loader: regex now includes hyphens. Hyphenated focus IDs (e.g. `SOV_Monopoly_in_the_post-Soviet_space`) were silently truncated, producing ~60 spurious "missing loc key" warnings.
  - `validate_focus_tree.py`: recognises `joint_focus` as a top-level focus definition (previously only `shared_focus` matched).
  - `validate_on_actions.py` duplicate-event check: computes a gate signature from enclosing `if` / `else_if` / `random` / `random_list` / numeric-weighted / country-tag / iterator-scope blocks, so refs in mutually-exclusive branches are no longer flagged. 127 warnings → 0.
  - `validate_localisation.py` color-syntax check: auto-detects keys referenced via `$KEY$` substitution and skips the per-line section check for them.
  - `validate_gfx_references.py`: auto-discovers a vanilla HoI4 install (Steam paths + `HOI4_PATH` env var) and unions its sprite definitions with the mod's. 44 errors → 5.
  - `validate_events.py`: removed the incorrect `news_event` `fire_only_once` + `major` combination check. Vanilla `NewsEvents.txt` uses this combination throughout. 72 errors → 0.
  - `validate_scripted_localisation.py`: in staged mode, always scans the whole repo for usages even when only staged definitions are checked.

- **Validator bug fixes from a follow-up audit:**
  - `validator_common._pool_map` operator-precedence fix — missing parentheses around `(self._pool is None and len < 10)` made multiprocessing silently fall back to sequential when workers > 1 and pool not yet initialised.
  - `validate_on_actions._scan_on_action_block` now receives the right `line_offset`, fixing the coordinate mismatch that made duplicate detection unreliable.
  - `validate_events.count_event_ids_in_file` only counts IDs that actually appear in cleaned text, preventing false positives from substring matches inside log strings.
  - `validate_localisation.get_all_colors` logs a warning when `core.gfx` is missing or unparseable so fallback behaviour is visible.

- **Regression test suite** (`tools/validation/tests/validator_bug_regressions_test.py`) — 10 tests lock in the false-positive fixes so they cannot silently regress.

#### Performance

Full-suite wall time on this mod: **4:17.76 → 1:29.99 cold / 0:43.92 warm** (65% / 83% faster). Issue counts and structural issue list (file + line + severity + category) are byte-identical to the pre-branch baseline — verified across both runs.

- **`BaseValidator.get_full_path` basename index** — collapses the per-flag `glob.iglob("**/*.txt")` full-tree walk that dominated `validate_variables` (34s → 8s standalone). `find_line_number` now reads via `FileOpener` so iterating many lookups against the same file only hits disk once.
- **`FileOpener` cache** switched from full-clear-on-overflow to `OrderedDict` LRU, raised to 8192 entries.
- **`compute_line_offsets` + `line_for_offset`** turn the `raw.count("\n", 0, m.start())` per-match line counting in `validate_gfx_references` from O(N²) to O(log N).
- **`validate_cosmetic_tags`** collapses three sequential pool scans into one batched worker.
- **`validate_decisions`** parallelises the previously-sequential cross-tree activation scan; module-level patterns shared across pool tasks.
- **`validate_unused_textures`** hoists its regexes to module level (32s → 17s cold) and disk-caches per-game-file path extractions.
- **`validate_set_variables` and `validate_variables`** persist per-file regex results under `.validation_cache/` keyed by `(filename, mtime_ns, size)`. The `set_variables` cache namespace also folds in a hash of the tracked-variable set so the alternation-regex output is only reused when the input set is unchanged.

#### Disk cache + CI

- **`tools/validation/disk_cache.py`** — new module exposing `per_file_cached` and `aggregate_cached`. Pool workers call these directly; atomic writes via `os.replace` stay safe under concurrent tasks. `.validation_cache/` is gitignored.
- **`.github/workflows/coding-pipeline.yml`** restores the cache via `actions/cache@v5.0.5` keyed on the PR base SHA plus a hash of `tools/validation/`. Restore-keys walk back to any prior cache for the same validator so cold starts on a new base SHA still get partial reuse.

#### Cleanup

- Removed the `--include-slow` gate and the empty `SLOW_VALIDATORS` set; all validators now run by default in `run_all_validators.py`.
- Deleted the legacy `tools/validation/run_all_validators.sh` wrapper (out of date and no longer referenced).
- Removed ~300 lines of dead per-operation flag/target workers in `validate_variables.py` (`process_file_for_flags`, `_get_flags`, the `get_all_used/set/cleared_*` wrappers, and their `EventTargets` twins) — superseded by single-pass workers added earlier on this branch.
- Trimmed verbose docstrings throughout to match the new comment style.

#### Content

- **[ZOM] Unified zombie game rules under `MD_CUSTOM_SCENARIO_RULES` group** so all custom-scenario toggles appear together in the rules screen, instead of split across `RULE_ZOMBIE` and the rest.
- Removed a duplicate game-rule definition from `common/game_rules/00_game_rules.txt`.

#### Docs

- `tools/COMMENT_STYLE.md` — Python comment rules: default to none, only add one when the **why** is non-obvious, never restate the code.
- `tools/validation/DISK_CACHE.md` — how the cache works, layout, CI wiring, clearing it locally.
- `/validate` skill now calls `run_all_validators.py` directly.

### Test plan

- Run `python3 tools/validation/run_all_validators.py` from the repo root → confirm all validators complete with the same error/warning counts as `main` and the run finishes in well under two minutes cold (well under one minute warm).
- Run `python3 -m pytest tools/validation/tests/ -q` → all 22 regression tests pass.
- Toggle a zombie game rule in the custom-game setup screen → verify it appears under the unified custom-scenario group.
- Touch a single `.txt` file under `common/` and re-run the suite → verify the warm cache only re-scans that file (per-validator timing summary should show near-zero scan time on the slow validators).


XCezor (approved):
gtg

PR #1706, opened 2026-06-07
Add post-raid damage reports for carrier-based raids (carrier_raids.txt) 

hhx-ux:
This modification improves carrier-based raid feedback in MD by displaying damage results after each completed raid. The feature helps players better evaluate operational impact and track the effectiveness of carrier-based air strikes. 

So with this change, the amount of damage dealt to the enemy can now be seen when a raid is successful. Previously, it wasn't visible, which seemed to be a bug.

### DIFF
--- a/common/raids/carrier_raids.txt
+++ b/common/raids/carrier_raids.txt
@@ -1,1134 +1,1257 @@
 types = {
 
-	# =============================================
-	# FUEL RAID - CARRIER DIRECT
-	# =============================================
-	fuel_production_storage_raid_air_carrier = {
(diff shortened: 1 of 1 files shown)


AngriestBird:
Noting that the merge conflicts need to be resolved prior to getting a good review. 

@hhx-ux for future note with more work that you do, you should operate on a branch on your fork and try to merge that. You end up with weird conflicts like this which give you more headache then it is worth here. Let us know if you need help resolving the conflict but my recommendation is to take your code from yours, update your fork from our main and then reapply your changes on a branch. I know you're not super familiar with this but it would help us a ton if you could try to do that.

TemplarGeneral:
@hhx-ux any update on this?

PR #1737, opened 2026-06-09
Summaries

JonathanWoollett-Light:
## This isn't ready to merge yet

Makes it a bit easier for people to pick paths by making the affects more easily readable.

This is quite an opionated change I can see some people might disagree with this or find it clutters the view.

# Some examples

image

image

image
 
Screenshot 2026-06-09 053540


AngriestBird:
@JonathanWoollett-Light While I think this is helpful it seems a bit superfluous since you can view individualistic and just kinda supports meta playing which I am not sure where we want to go. Thoughts on pros and cons on this?

TemplarGeneral:
@JonathanWoollett-Light did you looked at this yet?

JonathanWoollett-Light:
While an interesting idea, it seems a little too difficult to implement nicely.

PR #1739, opened 2026-06-09
Limit modifier tooltip indent to dynamic-modifier entries

JonathanWoollett-Light:
PR #1736 added a leading two-space indent to every entry in `replaced_from_vanilla_modifiers_l_*.yml`. That was wrong, it should only be on localisations used for dynamic-modifier tooltips. This rolls back a lot of the spaces which where added, some may be re-added in the future if we notice they need them.

# Before

image

# After

image


ducksnacks (approved):
Go to bed dawg

PR #1754, opened 2026-06-10
Wiki

kotru21:
## Summary

- **Single H1 per page:** layouts render the page title; duplicate `# Title` lines removed from content. `check:content-html` fails on hero collections when body H1 matches frontmatter `title`.
- **Markdown pipeline refactor:** remark/rehype plugins moved to `markdown-processor.ts`; sanitize schema extracted to `markdown-sanitize-schema.ts` with `target`/`rel` allowed on anchors so external links keep `target="_blank"`.
- **Accessibility:** post-build check flags links labeled "(opens in new tab)" without `target="_blank"`.
- **SEO:** safe JSON-LD serialization; changelog Article schema gets `datePublished` and `imageUrl`; KaTeX OG fonts vendored locally (`katex` dep removed).
- **Misc:** searchable-index pagination height fix; lightbox backdrop click suppressed after drag/pinch; CONTRIBUTING H1 guidelines; Bun/Astro/MDX/tooling bumps.


AngriestBird:
@kotru21 This was marked ready for review. Good for me to look it over?

kotru21:
> @kotru21 This was marked ready for review. Good for me to look it over?

Yeah

docs/src/shared/lib/markdown/markdown-processor.ts
```diff
@@ -0,0 +1,33 @@
+import { unified } from "@astrojs/markdown-remark";
```
AngriestBird, line 1: Blocking: `@astrojs/markdown-remark` does not export `unified`, so `bun run build` fails immediately (`does not provide an export named 'unified'` → `Unable to load your Astro config`). The whole `ci` chain dies here, so none of the post-build checks run.

Beyond the import, this standalone `unified()` processor is the wrong shape for Astro's content pipeline (see astro.config.ts:52). Fix: drop this file and wire the plugins back through `markdown.remarkPlugins` / `markdown.rehypePlugins`.

docs/astro.config.ts
```diff
@@ -50,22 +49,6 @@ export default defineConfig({
...
-          rel: ["noopener", "noreferrer"],
-          content: { type: "text", value: " (opens in new tab)" },
-        },
-      ],
-      rehypeTableScope,
-      rehypeTableWrapper,
-      rehypePreWrapper,
-      rehypeTailwindContent,
-    ],
+    processor: markdownProcessor,
```
AngriestBird, line 52: Blocking: `markdown.processor` is not an Astro option. Astro builds its own processor via `createMarkdownProcessor`, which only accepts `remarkPlugins`/`rehypePlugins`/`syntaxHighlight`/`shikiConfig`/etc. The unknown `processor` key is silently stripped by the Zod schema (no error), so every plugin is disabled: country directives, base-path link rewriting, `rehypeExternalLinks` (target=_blank / "(opens in new tab)"), image dimensions, table wrappers, tailwind injection, and the new sanitize. Hard regression vs `main`.

Fix: restore the supported keys:
```ts
markdown: {
  syntaxHighlight: ...,
  shikiConfig: { langs: [hoiscriptLanguage] },
  remarkPlugins: [remarkDirective, remarkCountryDirectives, [remarkRootRelativeToBase, SITE_BASE_PATH]],
  rehypePlugins: [rehypeImgAlt, rehypeImageDimensions, [rehypeExternalLinks, {...}], rehypeTableScope, rehypeTableWrapper, rehypePreWrapper, rehypeTailwindContent],
}
```
Gate the PR on a passing `bun run build`.

docs/scripts/check_accessibility_basics.py
```diff
@@ -36,6 +41,25 @@ def handle_starttag(self, tag: str, attrs: list[tuple[str, str | None]]) -> None
...
 
+        if tag == "a":
+            self._in_anchor = True
+            self._anchor_text = ""
+            self._anchor_target = (attr_map.get("target") or "").strip() or None
+            self._anchor_line = self.getpos()[0]
+
+    def handle_endtag(self, tag: str) -> None:
+        if tag == "a" and self._in_anchor:
+            if "(opens in new tab)" in self._anchor_text and self._anchor_target != "_blank":
```
AngriestBird, line 52: This check passes vacuously given the config issue above: `rehypeExternalLinks` no longer runs, so built HTML has neither `(opens in new tab)` nor `target="_blank"`, and nothing is ever flagged. The logic itself is fine. Once the pipeline is restored it'll actually do its job. Add a fixture/test with a known external link so a future pipeline break is caught here instead of silently passing.

docs/src/pages/feed.xml.ts
```diff
@@ -19,8 +20,10 @@ export async function GET(context: APIContext) {
 
   const items = [
     ...visibleChangelogs.map((entry) => mapItem(entry.data.title, entry.data.description, getChangelogPath(entry))),
-    ...devDiaries.map((entry) => mapItem(entry.data.title, entry.data.description, getDevDiaryPath(entry))),
-  ];
+    ...devDiaries.map((entry) =>
+      mapItem(entry.data.title, entry.data.description, getDevDiaryPath(entry), entry.data.date),
+    ),
+  ].sort((a, b) => (b.pubDate?.getTime() ?? 0) - (a.pubDate?.getTime() ?? 0));
```
AngriestBird, line 26: Side effect: changelog items have no `pubDate`, so they all collapse to `0` here, sink below every dated dev diary, and their mutual order goes unstable. On `main`, changelogs led the feed. Fix: give changelog items a date, or sort within groups and keep changelogs first.

AngriestBird (commented):
Reviewed the code paths and ran `bun run build` on this branch. Two blocking issues: the build crashes on a bad import, and the markdown refactor is wired to a config key Astro doesn't read, which silently disables the whole remark/rehype chain `main` had working. Inline details below. SEO/JSON-LD, CSP, lightbox, and pagination changes look solid.

AngriestBird:
Take the above with a grain of salt. Just some quick Claude ception. I would verify that the above is correct or incorrect. If they are then we should likely look at integrating the build step but no deploy as part of the PR verification when work on the site to ensure its' functional there.

kotru21:
> Take the above with a grain of salt. Just some quick Claude ception. I would verify that the above is correct or incorrect. If they are then we should likely look at integrating the build step but no deploy as part of the PR verification when work on the site to ensure its' functional there.

No worries, I've already set this up some time ago when was migrating the wiki from Jekyll to astro
image

kotru21:
As for other AI findings, I'll check it. i'm not an Astro dev specifically, so they may be real

AngriestBird:
Approved then. Thank you! Claude getting cranky over nothing.

AngriestBird (approved):

kotru21:
Wait bro i didnt verify the ai findings lol. i need a few mins

AngriestBird:
I didn't see the comment when I merged it already. Totally sorry about that. 

Edit: I assumed when you said the build was there it was fine haha.

docs/src/shared/lib/markdown/markdown-processor.ts
kotru21: Couldn't reproduce. @astrojs/markdown-remark@7.2.0 (pinned in bun.lock) re-exports unified as Astro's processor factory, and a clean bun run build succeeds (144 pages). unified() here returns the { name, createRenderer, … } processor descriptor that Astro's schema expects (markdown.processor defaults to () => unified()), so this is the intended shape, not the standalone unified library.

My guess is the failing build was against stale node_modules — on main @astrojs/markdown-remark isn't even a dependency and astro is ^6.4.4, so without a fresh bun install after checkout you'd get an older markdown-remark without the unified export. After bun install against the lockfile it builds clean. Happy to add a CI gate on bun run build to make this explicit(I'll open a new PR w/that soon).

docs/astro.config.ts
kotru21: processor is a supported key in this Astro version — see astro/dist/core/config/schemas/base.js (processor: z.object({ name, options, createRenderer, createMdxRenderer }).default(() => unified())), so Zod doesn't strip it. Verified the full chain runs in the built output on both .md and .mdx:

rehypeTailwindContent: heading scroll-mt-[calc(var(--header-height)…] markers present in changelog and dev-diary HTML.
rehypeTableWrapper: div.table-wrapper in tutorials and .mdx dev-diaries.
rehypeExternalLinks (+ rehypeSanitize, which whitelists target/rel): e.g. troubleshooting-guide → …rel="noopener noreferrer" target="_blank"…<span> (opens in new tab)</span></a>.
So no plugins are disabled and it's not a regression vs main. If you'd prefer the conventional remarkPlugins/rehypePlugins keys for clarity 

I can switch back, but functionally the current config is equivalent.

docs/scripts/check_accessibility_basics.py
kotru21: Good call on the fixture. The check isn't actually vacuous here (built HTML does carry target="_blank" + (opens in new tab), see the config thread), but to guard against a future pipeline break I added a --self-test mode with fixtures — including an external link with and without target="_blank" — and wired it into docs-quality.yml before the baseline run. Passes locally (3 fixtures).

docs/src/pages/feed.xml.ts
kotru21: Fixed — real regression, thanks. 
changelogSections has no date field, so the global descending sort sank every changelog below the dated dev-diaries (on main changelogs led). Now changelogs keep their lead and the date sort is scoped to dev-diaries only. Verified in dist/feed.xml: changelogs first, then dev-diaries newest-first. **(Minor: Array.sort is stable in V8/Bun, so the changelogs' mutual order wasn't actually unstable — but the sinking was real.)**

kotru21:
Ok so one there's actually one thing that'll be good to fix.
AI sometimes hallucinates — for better code review check the context7 MCP server with up-to-date docs.

AngriestBird:
Yeah I use context7 MCP. I just misunderstood what you meant haha

kotru21:
> Yeah I use context7 MCP. I just misunderstood what you meant haha

Heads-up: Dependabot and I only update the tracked lockfile in git. Your local env stays stale until you run bun install in docs/ after pulling. Use bun update when you intentionally want to bump deps locally.

And tell the guys who sometimes do touch the docs to do this from time to time so you all would have exactly the same env.
Bcuz after I'll update something major wiki can break on local machines. That's also may be the issue why Claude hallucinated some bugs — behaviour of some Astro features can change

PR #1769, opened 2026-06-11
Compatible Patch

TemplarGeneral:


common/ideas/indonesian.txt
TemplarGeneral, line 1: The tabbing is off

common/buildings/00_buildings.txt
TemplarGeneral, line 1: Building needs pricing

TemplarGeneral: $15.50

common/ideas/indonesian.txt
TemplarGeneral: Fixed

Blazer135 (approved):

Issue #1761, opened 2026-06-10
[TASK] Singapore Improvements


TemplarGeneral:
- [x] BoP for something
- [ ] Balance the tree
- [x] run the audit
- [ ] standardize all files

Blazer135:
Audit is Done this had a lot of Major issues especially with checks and missing ideas and so on, i am not going to add new content.
1. made new scripted trigger to simplify code just like the HKG content 
2. removed the Every_country checks, swapped it with the already available loop for SCO members
3. GUI dirty counter added
for now this is the major points that got fixed

Blazer135:
BOP Values Slashed to half

PR #1787, opened 2026-06-12
Task1761 singapore

Blazer135:
Fixed all the Tasks Except Focus tree rebalance honestly i am not sure how to balance it without gutting the focus tree 

reference Task #1761


Blazer135:
i did a review but there is to much fuckups [in the branch] i might have missed something from like the to many white spaces or AI

AngriestBird (approved):

PR #1797, opened 2026-06-13
Energy GUI rework

JonathanWoollett-Light:
Reworks the energy GUI the same way the economy/tax GUI was reworked.

# After

image

# Before

image


TemplarGeneral:
Il close it as mentioned in ui-overhaul channel. This needs to be integrated with Markspider and having a new location.

PR #1882, opened 2026-06-16
Macabre

ducksnacks:
Add national spirt for UK about you protecting BP


TemplarGeneral:
Could you fix the validators? Otherwise its good 👌🏻

AngriestBird (approved):

AngriestBird:
@TemplarGeneral The two that failed are broken validators. I need to fix those.

PR #2105, opened 2026-06-27
Redefined the LUA for Carrier Deck sizes to reflect new airwing sizes of 12

BigRobbo919:
I noticed in the latest version of the BETA that the air wings have been resized but that carrier Decks remain in multiples of 10 I dug into the LUA and found the stat to change so that carrier decks increase in sizes of 12. Hope this helps


AngriestBird:
@BigRobbo919 The fix is going out for this tonight already :)

BigRobbo919:
Oh... fair enough hahaha!

PR #2303, opened 2026-07-06
Fix PER_attack_ikr: release IKR before declaring war

kiarash6k:
## What
`PER_attack_ikr` never actually starts the war against Iranian-Kurdistan (IKR).

## Why
The decision is `visible` while IKR is Iran's subject (`autonomy_federal_state_iran`), and its `remove_effect` does `declare_war_on { target = IKR }` (plus `war_with_on_remove = IKR`). HOI4 does not allow an overlord to be at war with its own subject, so both the `declare_war_on` and `war_with_on_remove` silently no-op. The player only receives the `add_stability = -0.15` and the state cores — no war ever begins.

## Fix
Add `end_puppet = IKR` before the war declaration so IKR is released from subject status first, allowing the war to start as intended.

```
PER = {
    end_puppet = IKR
    declare_war_on = { target = IKR  type = annex_everything }
}
```

Reproduced on current `main` via the Iran post-revolution civil-war path (IKR becomes a federal-state subject, then the "attack IKR" decision does nothing).


common/decisions/Iran.txt
```diff
@@ -4054,6 +4054,10 @@ PER_domestic_politics = {
 		remove_effect = {
 			log = "[GetDateText]: [Root.GetName]: Decision PER_attack_ikr"
 			PER = {
+				# Release IKR from subject status first — an overlord cannot
```
Blazer135: these are not needed the comments only keep the end_puppet = IKR and make sure to wrap it in a check to check for IKR exist and is a puppet

kiarash6k: Done — removed the comments and wrapped it in `if = { limit = { country_exists = IKR  IKR = { is_subject_of = PER } } end_puppet = IKR }` so it only releases IKR when it exists and is Iran's subject. Thanks for the review!

AngriestBird (approved):

PR #2304, opened 2026-07-06
Fix CI: restore missing localisation BOM and whitelist 3 unused regional flags

kiarash6k:
## What
Fixes two pre-existing CI failures on `main` (both surface on unrelated PRs, e.g. #2303):

### 1. Missing UTF-8 BOM
`localisation/english/MD_focus_NKO_l_english.yml` lost its UTF-8 BOM, failing **Structural Lint → Check localisation UTF-8 BOM**. Re-added with the repo tool:
```
python3 tools/linting/validate_localization_encoding.py --fix localisation/english/MD_focus_NKO_l_english.yml
```

### 2. Unused country flags
**Validate Variables** flagged three set-but-unused flags in `common/scripted_effects/00_country_group_flags.txt`:
- `somali_nation_flag` (line 38)
- `sudanese_nation_flag` (line 41)
- `iranian_nation_flag` (line 94)

These are long-standing sub-regional infrastructure flags (added in #1216), parallel to the ~20 other regional flags in that file — they simply have no `has_country_flag` consumer yet. I added them to `FALSE_POSITIVES_COUNTRY_UNUSED` (the mechanism already used for other intentional set-but-unused flags) rather than deleting them, to preserve the infrastructure.

**Alternative:** if maintainers would rather drop the dead flag-sets entirely, I can remove the three `if`/`set_country_flag` blocks instead — just say the word.

## Verification
Both validators pass locally:
```
validate_localization_encoding.py  → 276 valid, 0 errors
validate_variables.py              → NO ISSUES FOUND
```

### DIFF
--- a/tools/validation/validate_variables.py
+++ b/tools/validation/validate_variables.py
@@ -775,6 +775,12 @@ class Validator(BaseValidator):
             "_QMV_voted",
             "_EP_approval",
             "recognised_opponent_",
+            # Regional infrastructure flags set in 00_country_group_flags.txt
+            # (parallel to the other sub-regional flags) that currently have
+            # no has_country_flag consumer; kept for runtime/future use.
+            "somali_nation_flag",
+            "sudanese_nation_flag",
+            "iranian_nation_flag",
         ]
 
         for flag_type, fp_cleared, fp_missing, fp_unused in [
--- a/tools/validation/validate_variables.py
+++ b/tools/validation/validate_variables.py
@@ -775,12 +775,6 @@ class Validator(BaseValidator):
             "_QMV_voted",
             "_EP_approval",
             "recognised_opponent_",
-            # Regional infrastructure flags set in 00_country_group_flags.txt
-            # (parallel to the other sub-regional flags) that currently have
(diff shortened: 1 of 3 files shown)


tools/validation/validate_variables.py
```diff
@@ -775,6 +775,12 @@ def run_validations(self):
             "_QMV_voted",
             "_EP_approval",
             "recognised_opponent_",
+            # Regional infrastructure flags set in 00_country_group_flags.txt
+            # (parallel to the other sub-regional flags) that currently have
```
AngriestBird: We can just remove these flags I don't think there is any issues with removing these. If we need them we can add them back.

AngriestBird (changes requested):

tools/validation/validate_variables.py
kiarash6k: Done — removed the three unused flag blocks from `00_country_group_flags.txt` and reverted the whitelist addition, so the validator passes on the actual content. Only the BOM fix remains. Thanks!

AngriestBird (approved):

PR #2392, opened 2026-07-11
fix: Malta crash (nested else syntax) and starting microchip/composite tech modifiers

### LINKED ISSUE #2385: [BUG] Microchip Plant Research not adding increased production of microchips
> *Reported by **discord-sync** via [Discord](https://discord.com/channels/210890013334831104/1525460790818115664) on 2026-07-11*
> *Tags: BETA, General Bug*

**Describe the bug**
HOI4 version  v1.19.2.0.a729, all DLCs enabled
MD Beta  v2.0.0
Checksum b1e3

Playing as USA with latest MD Beta v2.0.0 available on Steam and all DLCs enabled, noticed that the research line under Electronics that should increase the production of microchips per plant (usually adding +1 per research completed) does not actually increase it. 
Added screenshot from Texas where i have 2 Microchip plants constructed which technically should produce 36 microchips (32 base + 4 additional chips) but just produces the base.
You can see that for Composites it does increase it correctly. Screenshot from Texas with 2 Advanced Composites constructed which produces 40 (32 base + 8 additional since I have 3 researches completed).

**To Reproduce**

**Screenshots**

**Operating System**

**Additional context**

hasan4938:
Closes #2385

### Summary of Changes

This branch addresses several bugs related to technology modifiers, startup scope warnings, and building damage effects:

1. **Microchip & Composite Tech Updates (`00_microchip_effects.txt`, `00_on_actions.txt`):**
   - **Fixes #2385**: Replaced hardcoded multipliers in `microchip_update` and `composite_update` with dynamic loops that scale based on the country's researched tech levels.
   - Restored starting technology checks on game startup (`microchip_startup`) to ensure starting modifiers (e.g., USA's initial tech levels) are correctly applied to buildings.

2. **Raid Categories Scope Warning (`raid_categories.txt`):**
   - Fixed a startup scope warning (`Invalid Scope: None`) in the `EH_portal_raids` visibility block. Inlined the game rule checks directly instead of calling the `EH_scenario_enabled` scripted trigger in a global (None) scope.

3. **Building Damage Warnings in Turkey Events (`events/Turkey.txt`):**
   - Resolved recurring monthly warnings (`Failed to find building for damage effect`) during the PKK insurgency events. 
   - Wrapped `damage_building` effects for `offices` and `industrial_complex` in conditional blocks (`offices > 0`, `industrial_complex > 0`) to ensure the game only attempts to damage buildings that actually exist in states 935 and 162.


AngriestBird:
The nested else in the if statement is valid syntax. Did you test that this actually fixes the crash? 

The microchip part. Did you confirm this actually fixed the issue? I'm fairly certain the microchip one the composites was reported as working. I think part of the issue here is that the updats may not be properly accounting for the right amount of microchips in total based on the tech.

hasan4938:
Thanks for the feedback! 

1. **Nested Else & History Files:**
You are completely right. The nested `else` is indeed valid syntax. Reformatting them was a mistake on my part. I have reverted all modifications to the country history files and restored them to their original state. 

2. **Microchip startup fix:**
The bug occurs because the USA starts with pre-researched microchip technologies (`microchip_production_1`, `2`, and `3`) in their history file. HOI4 does not execute the `on_research_complete` effect blocks of pre-researched starting technologies when initializing the game, meaning the USA never received the `modify_building_resources` modifier for microchip plants on startup. Calling `microchip_startup` on game start fixes this by manually applying the modifiers for starting techs.

3. **Dynamic Updates Math:**
I've updated `microchip_update` and `composite_update` to dynamically calculate the per-plant production based on the researched technology level (`microchip_production_X` / `composite_production_X`), instead of using hardcoded multipliers. This ensures the update variables correctly account for the exact microchips produced on the map.

I cleaned up the branch and force-pushed. The PR now only contains clean changes to the two relevant files:
- `common/on_actions/00_on_actions.txt`
- `common/scripted_effects/00_microchip_effects.txt`

AngriestBird:
Haven't forgotten about this. I will be looking at this in depth later to see if there's a better solution. It might be the only way but I would rather see if there's something else where we don't need to maintain two sources of truth for these modifications.

AngriestBird (approved):

PR #2526, opened 2026-07-20
Fix non-focus validation errors

AngriestBird:
## Summary

Fixes the non-focus validation failures left after the focus-tree validation branch was merged.

- Replaces deprecated faction creation with faction templates.
- Converts membership scans to maintained arrays.
- Replaces runtime nation triggers with initialization flags.
- Corrects tautological trigger conditions.

## Validation

- `check_common_mistakes.py`: passed.
- Non-focus style checks: passed.
- Full style output contains only focus-tree-standard warnings.


AngriestBird:
Superseded by #2527, which is based directly on main and contains only the six intended files.

PR #2631, opened 2026-07-28
Sat automation

Blazer135:
Added Automation to allow player to Automate Satellite building and launching, did not include Killsat
fixed a couple of bugs 
1. Constellation was not upgrading and it stayed on the basic tier so now it is automatically swapped 
2. Space production price and duration are now read from the tier of the selected model rather than its position in the display list, so skipping a generation no longer prices a new satellite as an old one
3. The ctrl-click and shift-click variants of the space production amount and parallel-line buttons now respect an in-progress batch, matching the plain buttons.
4.[PMC] The Halo Corporation can now be hired by the United States and the United Kingdom - the sponsor check was always true, so both were forced down the path requiring each country to hold an opinion of itself


common/scripted_effects/00_missiles_scripted_effects.txt
```diff
@@ -420,6 +348,13 @@ calculate_production_vars_13 = {
 	clamp_variable = { var = var_gui_prod_13_days min = 1 }
 	clamp_variable = { var = var_gui_prod_13_price_total min = var_gui_prod_13_price_unit }
 	clamp_variable = { var = var_gui_prod_13_amount min = 1 }
+
+	set_variable = { var_space_week_13 = var_gui_prod_13_days }
+	divide_variable = { var_space_week_13 = 7 }
+	clamp_variable = { var = var_space_week_13 min = 1 }
+	set_temp_variable = { temp_week_price = var_gui_prod_13_price_total }
+	divide_temp_variable = { temp_week_price = var_space_week_13 }
+	set_variable = { var_space_week_13 = temp_week_price }
```
AngriestBird:
```
set_variable = {
	var_space_week_13 = {
		value = var_gui_prod_13_price_total
		divide = {
			value = var_gui_prod_13_days
			divide = 7
			clamp = { min = 1 max = 1000000 } # magic number for the AI to not be stupid
		}
	}
}
```

common/scripted_effects/00_space_automation_effects.txt
```diff
@@ -0,0 +1,1853 @@
...
+	}
+	else_if = {
+		limit = {
+			check_variable = { temp_pid > 160 }
+			check_variable = { temp_pid < 173 }
+		}
+		set_temp_variable = { temp_fam = 2 }
+		set_temp_variable = { temp_tier = temp_pid }
+		subtract_from_temp_variable = { temp_tier = 160 }
+		set_temp_variable = { temp_idx = temp_tier }
```
AngriestBird: Use the math expression format from before to optimize this

common/scripted_effects/00_space_automation_effects.txt
```diff
@@ -0,0 +1,1853 @@
...
+	add_to_array = { space_auto_orbit_array = temp_cur }
+	add_to_array = { space_auto_target_array = temp_max }
+	set_temp_variable = { temp_short = temp_max }
+	subtract_from_temp_variable = { temp_short = temp_cur }
+	clamp_temp_variable = { var = temp_short min = 0 max = 10000 }
+	add_to_array = { space_auto_short_array = temp_short }
+}
+
+space_auto_refresh_comsat = {
+	set_temp_variable = { temp_tier = 0 }
```
AngriestBird, line 277: Recommend a simplify pass on this and potentially look at setting a variable somewhere else as this adds a lot if else cehcks for minimal gain. 

Interpolating something like (if has_tech or on_research_complete should set a variable so it's gnss_level = x and then you can omit all these if else if checks

common/scripted_effects/00_space_automation_effects.txt
```diff
@@ -0,0 +1,1853 @@
...
+					subtract_from_temp_variable = { temp_fit_mi = 1 }
+					set_temp_variable = { temp_fit_mass = global.sat_mass_array^temp_fit_mi }
+					if = {
+						limit = { check_variable = { var = temp_capacity value = temp_fit_mass compare = greater_than_or_equals } }
+						set_temp_variable = { temp_fit_searching = 0 }
+					}
+				}
+				if = {
+					limit = { check_variable = { temp_fit_searching = 0 } }
+					set_temp_variable = { temp_fit_id = temp_fit_base }
```
AngriestBird: more math expressions. Highly recommend we review for these since we can eliminate unneeded temp variables since we are using them for filler and jsut directly do the math in line with set_variable etc

common/scripted_effects/00_space_automation_effects.txt
```diff
@@ -0,0 +1,1853 @@
...
+		subtract_from_temp_variable = { temp_owed = var_space_paid_11 }
+		clamp_temp_variable = { var = temp_owed min = 0 max = 1000000 }
+		add_to_temp_variable = { temp_committed = temp_owed }
+	}
+	if = {
+		limit = { has_active_mission = missile_construction_12 }
+		set_temp_variable = { temp_owed = var_gui_prod_12_price_total }
+		subtract_from_temp_variable = { temp_owed = var_space_paid_12 }
+		clamp_temp_variable = { var = temp_owed min = 0 max = 1000000 }
+		add_to_temp_variable = { temp_committed = temp_owed }
```
AngriestBird: Use the example I gave earlier for more of this. I would also set a constant at the root and reference @<value_of_constant>

AngriestBird (changes requested):

common/scripted_effects/00_missiles_scripted_effects.txt
```diff
@@ -395,6 +327,17 @@ calculate_production_vars_12 = {
 	clamp_variable = { var = var_gui_prod_12_days min = 1 }
 	clamp_variable = { var = var_gui_prod_12_price_total min = var_gui_prod_12_price_unit }
 	clamp_variable = { var = var_gui_prod_12_amount min = 1 }
```
AngriestBird, line 329: Can you do the math expressions for the entire production scripted effect?

AngriestBird (changes requested):

AngriestBird (approved):

AngriestBird:
Math expressions improvements will be deferred to follow on work.

PR #2684, opened 2026-08-02
Preview the acceptance outcome on focuses that send an offer abroad

AngriestBird:
### Changes
- Focuses that send an offer to another nation now preview what you receive if it is accepted, covering 248 focuses across 40 country trees
- Removed 24 event IDs that were being fired with no event behind them, so the affected focuses, decisions and event chains no longer stop dead partway through
- Reworded the effects-from-event tooltip header to read as potential effects, since it fronts an event where only one of several reactions ever lands
- Fixed the Swiss Eurofighter and Rafale purchases handing the jets and the 6bn bill to Germany and France instead of Switzerland
- The Swiss Rafale import focus now shows the 6bn the purchase costs
- Fixed the Volhynian referendum granting autonomy to Lviv instead of Volhynia
- Our Old Realm now applies its diplomatic cost and the Spanish reaction directly, instead of routing them through an event that was never written
- Replaced four one-off "if they accept" tooltip keys with the shared one and removed the orphaned localisation entries
- The event validator now flags any fire at an event ID no event file defines, so this cannot silently come back
- The focus validator no longer reports four classes of false positive: a bankruptcy guard whose cost is paid by the event it fires, an event whose options are a choice of reaction rather than accept or decline, an event whose options are pure flavour, and a fire inside an effect_tooltip that only previews what a decision does

Four focuses are deliberately left without an acceptance tooltip because their accept branch has no outcome to preview. Filed separately rather than papered over: #2686, #2687, #2688, #2689


Changelog.txt
TemplarGeneral, line 1: SWI and HKG changes shoudn't be mentioned because they sre not on main version of steam

TemplarGeneral, line 1: And ALG

common/national_focus/05_netherlands.txt
```diff
@@ -2845,6 +2844,7 @@ focus_tree = {
 			}
 			add_to_variable = { HOL_economy_microchip_export_multiplier_modifier = 0.15 tooltip = microchip_export_multiplier_modifier_tt }
 			newline = yes
+			custom_effect_tooltip = TT_IF_THEY_ACCEPT
```
TemplarGeneral: The current tooltip already says that. So the HOL tooltip should be replaced then with the effect_tooltip

common/national_focus/05_netherlands.txt
```diff
@@ -4737,6 +4735,7 @@ focus_tree = {
 			log = "[GetDateText]: [Root.GetName]: Focus HOL_delta_rhine_energy_corridor"
 			GER = { country_event = HOL_economy.024 }
 			newline = yes
+			custom_effect_tooltip = TT_IF_THEY_ACCEPT
```
TemplarGeneral: The current tooltip already says that. So the HOL tooltip should be replaced then with the effect_tooltip

common/national_focus/05_netherlands.txt
```diff
@@ -5605,6 +5620,7 @@ focus_tree = {
 					country_event = HOL_economy.015
 				}
 				newline = yes
+				custom_effect_tooltip = TT_IF_THEY_ACCEPT
 				custom_effect_tooltip = HOL_European_Railway_Development_bel
```
TemplarGeneral: The current tooltip already says that. So the HOL tooltip should be replaced then with the effect_tooltip

TemplarGeneral (changes requested):

TemplarGeneral (approved):

PR #2707, opened 2026-08-03
Validator Fixes

TemplarGeneral:
### Validator & tooling
- `validate_unused_scripted.py`: suppresses false positives on preset/convention libraries. Adds `set_partial_draft_effect` to `FALSE_POSITIVE_PATTERNS` and 14 whole-file exemptions to `FALSE_POSITIVE_FILES` (`00_alert_triggers.txt`, `00_continent_triggers.txt`, `00_cyber_triggers.txt`, `00_debt_ratio_triggers.txt`, `00_economic_triggers.txt`, `MD_antarctica_station_module_triggers.txt`, `00_budget_effects.txt`, `00_ct_effects.txt`, `00_economic_system_utilities.txt`, `00_generic_ideas_scripted_effects.txt`, `00_influence_scripted_effects.txt`, `00_internal_faction_effects.txt`, `00_pp_scripted_effects.txt`, `00_sanctions_scripted_effects.txt`)
- `validate_focus_tree.py`: the cross-country tooltip check now also accepts `TT_IF_THIS_ACCEPTS` and `TT_IF_EACH_ACCEPTS` alongside `TT_IF_THEY_ACCEPT` / `TT_EFFECTS_FROM_EVENT`. `TT_IF_THIS_ACCEPTS` renders `[THIS.GetNameWithFlag]`, so it sits inside the target's scope block rather than beside the `effect_tooltip`; the matcher is scope-agnostic and clears both placements. Disk cache key bumped `v5` → `v6`, plus two new cases in `focus_cross_country_tooltip_test.py`
- `validator_common.py`: drops `mtg_usa_congress_remove_state_tt` from `KNOWN_VANILLA_LOC_KEYS`, referenced nowhere in the mod

### Unused script & orphaned loc removal
- Removes unused scripted effects and scripted triggers flagged by the validator, plus the orphaned localisation keys that went with them
- Notable removals: the `FSA_death_over_humiliation` idea (never added by any script), `ENG_increase_agitation_3/4/5`, `SWE_modify_royal_acceptance` and its `SWE_royal_acceptance_status` history seed, `PER_AB_T`, `PER_reform_majlis`, `syrian_renewable_effect`, and the dead blocks in the ALG / ARM / BRA / CZE / ISR / NKO / USA / LAR / religion / middle-east / formable-ratio files

### error.log fixes
- Fixes `clamp_variable Unknow variable 'jap_karoshi_wellbeing' / 'jap_galapagos_openness'` firing every month from game start in `on_monthly_JAP` — `clamp_variable` errors on a variable that was never created, and both trackers are only ever written by focuses. Karoshi and cultural-change clamps now sit next to the effect that creates them; the Galapagos clamp is guarded on `has_variable`
- Fixes `MD_Country_Groups_Triggers.txt:1611: and: Invalid Scope ... provided: None` (4 errors on every load) — `MD_special_countries` was called bare inside the `allowed` blocks of five scripted diplomatic actions, which are global gates evaluated at init with no current scope. Now wrapped in `ROOT = { }`, matching the file's existing convention
- Fixes `construct_building_in_random_province: Can't construct building: Civilian R&D Facility in state: Mittelland` on `SWI_the_large_hadron_collider` — `nuclear_facility` shares a one-per-state cap with land/air/naval facilities and 2 of Switzerland's 3 states are already occupied. Facility and reactor construction are now guarded on `can_construct_building`, with a `CAT_nuclear_reactors` research bonus as fallback so the focus is never a dead reward
- Fixes the recurring `Failed to find building for damage effect` — `damage_building` was called on buildings the state does not have. Guarded the `intercept_drug_convoys` removal effect (state 931 ships with `infrastructure = 0`, so it errored every 14 days), the five cyberwarfare event sites, and the eight special-forces sabotage calls in `SF_marine_modern_demolition_effect` / `SF_marine_logistics_raiding_effect` / `SF_PARA_sabotage_effect`
- Fixes `clamp_variable Unknow variable 'var_space_auto_flights'` in `00_space_automation_effects.txt` — the variable is only ever written by the AI-only branch of `space_auto_ai_monthly`, so a player who completes the space program and opens the launching ledger hits `space_auto_preview` with it unset. `space_auto_preview` now seeds the default (2) when the variable does not exist; the duplicate clamp in `space_auto_toggle` is dropped since the very next line calls `space_auto_preview`
- Fixes `clamp_variable Unknow variable 'voj_separ'` in `99_SER_scripted_effects.txt` — `modify_voj_separ_support` wrote `THIS.voj_separ` but clamped `ROOT.voj_separ`, so every caller that scopes in (`VOJ = { ... }` from the Serbia events and decisions) clamped a variable SER never had. Both this and `modify_voj_integr_support` now clamp the scope they wrote to. Also fixes `MNT_traditional_vl_mon`, which called the effect in Montenegro's own scope and silently built a dead `MNT.voj_separ`
- Fixes `add_equipment_to_stockpile: adding an archetype equipment without equipment existing in country` on `BRM_singaporean_armour` — Myanmar creates no flame-chassis (IFV) design, so the bare archetype resolved to nothing while the focus still charged 5bn. Now grants `medium_tank_flame_chassis_2`, matching the 1995 `IFV_4` in the legacy branch and the pairing used by the UK and Czech focuses. Liechtenstein's `LIC_invest_apc` / `LIC_invest_ifv` had the same bug and no equipment variants at all
- Fixes `States are controlled by same county` / `error in start_border_war` from `Egypt.14.b` — the border war used state 209 (Gaza) as both attacker and defender, so it could never start. Egypt now attacks out of Sinai (213), guarded on both controllers, with a fallback to `Egypt.16` so the option is never a no-op. `BRM_operation_NHN` had a related defect: an `attacker` block and no `defender` block at all, now 505 → 930 to mirror the reverse decision
- Fixes `compliance_modifiers.txt: state: Invalid Scope, supported: State, provided: Country` (3 errors) — `compliance_80` is a `core_compliance_modifier`, so `on_enable` runs with the occupier as ROOT and the occupied country as FROM, and no state is in scope. The Xinjiang / Tibet / Inner Mongolia carve-out is now a tag check on FROM (`ETK` / `TIB` / `OMG`)
- Fixes `MD_on_actions.txt:71: remove_from_array: Invalid index to remove` — the weekly subsidy cleanup removed from `subject_subsidization` while `for_each_scope_loop` was still walking it, shifting every later entry. Removals are now staged into a second array and applied after the loop, matching the pattern in `99_COM_on_actions.txt`
- Fixes `Could not find sprite type [GFX_scorched_earth_state]` — MD's `countrystateview.gfx` fully overrides vanilla's and dropped this sprite, but `countryconstructionsview.gui` still references it. Re-declared against the same vanilla texture
- Adds controller guards to the unguarded 165 → 630 (Iraq/Kurdistan) and 206 ↔ 209 (Israel/Gaza) border wars, which produce the same "same country" error once either side occupies the other's state. Also corrects `iraq_md.81`, which used the Iraqi-victory event for both attacker and defender `on_win`

Smaller fixes of the same kind:

| Fix | File |
| --- | --- |
| `create_unit` referenced division templates the country may not own. `Piechota` and `Dywizja Kawalerii Pancernej` (NSB and non-NSB variants) are now defined on demand | `common/decisions/Poland.txt` |
| Same, for the `Territorial Defense Brigade` template Algeria never defines (only Belarus does) | `common/decisions/Algeria.txt` |
| `add_timed_idea = CZE_skoda_superb_iii_bought_idea` referenced a nonexistent idea. Corrected to `_iv_` across all 14 gen-4 Skoda export events | `events/Czech Republic.txt` |
| `change_tag_from = PUK` takes a scope, not a tag literal. Now `ROOT` | `events/Kurdistan.txt` |
| `belarus_another.19` ran `set_autonomy` in the event's own scope with a hardcoded `target = POL`. Now `FROM = { set_autonomy = { target = ROOT ... } }`, so the real overlord puppets the real recipient | `events/belarus.txt` |
| `add_tech_bonus` used the nonexistent name `HOL_space_tech` (twice), and a `random_list` bonus had no `name` at all. Both now `HOL_corporation_space` | `events/05_netherlands.txt`, `common/national_focus/05_netherlands.txt` |
| Alawite / Kurdish / Druze status math divided by `state_population` and `gdp_per_capita` without guarding against 0. Also `kurdish_gdp_per_capita` used `add_to_variable` where it should seed with `set_variable`, so it accumulated on every run | `common/scripted_effects/99_SYR_scripted_effects.txt` |
| `PER_majlis_elections` branched on the `PER_reformed_majlis` flag, which nothing sets since `PER_reform_majlis` was dead. Now `has_completed_focus = PER_weaken_guardian_council`, and the complementary `else_if` collapsed to `else` | `common/scripted_effects/99_PER_scripted_effects.txt` |
| `one_random_synthetic_refinery` is a country-scope effect but was called inside `595 = { }` on two focuses | `common/national_focus/hongkong.txt` |
| `resource_storage_system` ran its per-resource math on unset variables when the setup never happened (game rule toggled on mid-game, or an older save). Now runs `resource_storage_system_setup` when `steel_in_storage` is missing | `common/scripted_effects/00_resource_storage_effects.txt` |
| `MNT_traditional_vl_mon` is now guarded on `country_exists = VOJ` and scoped into VOJ | `common/national_focus/05_montenegro.txt` |
| `BNL_treaty_of_union`'s accept tooltip previewed `FROM = { puppet = ROOT }`, which is undefined at tooltip time in a joint focus. Now explicitly HOL puppeting BEL and LUX. `BNL_benesam` also gained the naval grand-doctrine prerequisite its reward assumes | `common/national_focus/03_benelux_shared.txt` |
| `HOL_authority_in_information` recreated an existing intelligence agency. Now guarded on `has_intelligence_agency = no` | `common/national_focus/05_netherlands.txt` |
| Nine FSA mission decisions were gated on `has_global_flag = FSA_free_syrian_army_active_flag`, a flag never set anywhere, so they could never activate | `common/decisions/Free_syrian_army.txt` |
| `ISR_raid_beit_leiha` and `HAM_start_raid_on_isr` now require controlling their own state and not the target's, so raids stop firing once territory flips | `common/decisions/Israel.txt` |
| `israel_economy.163.a` swept `every_other_country` for `has_idea = AU_member`. Replaced with `for_each_scope_loop` over `global.AU_member` plus a proper tooltip | `events/Israel_events.txt` |
| `[? VAR]` with a stray space does not render. Fixed on three Welsh/Scottish keys | `localisation/english/MD_focus_ENG_l_english.yml` |
| `MODIFIER_SCIENTIST_RESEARCH_BONUS_FACTOR` was referenced by `scientist_research_bonus_factor_tt` but never defined | `localisation/english/replace/replaced_from_vanilla_modifiers_l_english.yml` |
| `AFG_join_axis_of_resistance` and `TAJ_axis_of_resistance` fired their contact events even when Iran leads no faction, is at war, or already has us. Now gated on `PER = { is_faction_leader = yes }` and friends, with a bypass | `common/national_focus/05_afghanistan.txt`, `common/national_focus/tajikistan.txt` |
| `ContactIran.4.a` ("Iran accepts") had no effect at all. It now dissolves or leaves the accepter's faction and runs `FROM = { add_to_faction = ROOT }` | `events/Afghanistan.txt` |
| Swaps the multi-tag `is_horn_of_africa_nation` trigger call for the cached `horn_of_africa_nation_flag`, matching what the politics-view triggers already do | `common/scripted_guis/00_MD_politicsview_scripted_guis.txt` |

Cross-country focuses that fire an event at another country now front the outcome preview the validator asks for: `TT_IF_THEY_ACCEPT` on the Afghan, Belarusian and Tajik focuses, `TT_EFFECTS_FROM_EVENT` plus a new `ETH_eritrea_assembly_parties_tt` key on `ETH_eritrea_assembly`.

### Dutch House of Orange-Nassau
The Netherlands' Western Autocracy party now draws **Monarchist** internal-politics powers instead of Autocrat ones, reflecting that its "autocrat" slot is the monarchy rather than a strongman.

- Two new triggers in `01_political_triggers.txt`: `western_autocrats_have_autocrat_powers` (western autocrats in power, not a technocrat leader, not HOL) and `western_autocrats_have_monarchist_powers` (western autocrats in power and HOL)
- `00_scripted_localisation.txt`: `Western_Autocracy_scripted_loc` now excludes HOL, and a new `HOL_Western_Autocracy_scripted_loc` carries it into the Monarchist powers list in `IP_monarchist_powers_desc`
- `00_MD_politicsview_scripted_effects.txt`, `00_MD_politicsview_scripted_guis.txt` and `subparty_policies.txt`: the autocrat gates move to the new trigger and the Kingdom/Monarchist gates gain the monarchist one. This covers the idea configuration, the corruption update, both ideology glows, and the sub-party policy `cancel` blocks
- `non_technocrats_western_autocrats_are_in_power` is untouched and still used by the Italian content

### Balance & content
- `delta_wing_design` had `start_year = 1990` while sitting in the 2010 folder row. Corrected to 2010
- `HOL_international_peace_city_idea`: `stability_factor = 0.10` replaced with `foreign_influence_modifier = 0.05`
- `HOL_dutch_defense_doctrine`: dropped the `bypass = { NOT = { has_idea = the_military } }` and added `has_any_grand_doctrine = special_forces` to the doctrine count, which the completion reward already branched on
- `SWI_import_eurofighters` now requires `has_tech = gen_4_medium` under By Blood Alone, since the reward builds a gen-4 airframe variant
- Wires `ALG_update_anti_poaching` into all five law-budget ideas alongside the existing Botswana hook
- Localisation polish: Benelux integration tooltips gain colour tags and `GetNameWithFlag`, and `HOL_daf_organization` moves out of the vanilla-replace file into `MD_focus_HOL_l_english.yml`, shortened to "DAF"


common/scripted_effects/99_BRA_scripted_effects.txt
```diff
@@ -131,45 +131,6 @@ BRA_wild_acreage_effect = {
 	BRA_calculate_amazon_system = yes
 }
 
-# function: BRA_target_acreage_effect
-# purpose: Adjust the targeted total number of acreage
```
AngriestBird, line 135: These should be left for now for Brazil

common/scripted_effects/99_NKO_scripted_effects.txt
```diff
@@ -270,26 +270,6 @@ NKO_spa_initialise = {
 	}
 }
 
-NKO_spa_open_session = {
```
AngriestBird, line 273: I notified Blazer about this so he needs to redo this mechanic. Revert this one

common/scripted_effects/99_USA_main_scripted_effects.txt
```diff
@@ -229,72 +222,6 @@ USA_congress_large_support = {
 		USA_house_large_support = yes
 	}
 }
-USA_election_senate = {
```
AngriestBird, line 232: These need a new issue to implement for the American mid term elections. Restore these.

common/scripted_effects/99_SWE_scripted_effects.txt
```diff
@@ -1,10 +1,4 @@
 # Royal Stuff #
-# set_temp_variable = { modify_royal_acceptance = +- }
```
AngriestBird, line 2: Confirm this shouldn't be integrated into the existing Swedish content

localisation/english/MD_countries_cosmetic_l_english.yml
```diff
@@ -2820,19 +2820,19 @@
  HBL_fascism_ADJ: "Netherlands"
  HBL_fascism_DEF: "the Caliphate of the Low Countries"
  #Benelux (autocratic/monarchist Netherlands-led)
- HBL_REB_democratic: "United Kingdom of the Netherlands"
+ HBL_REB_democratic: "United Provinces"
  HBL_REB_democratic_ADJ: "Netherlands"
- HBL_REB_democratic_DEF: "the United Kingdom of the Netherlands"
- HBL_REB_communism: "United Kingdom of the Netherlands"
+ HBL_REB_democratic_DEF: "the United Provinces"
+ HBL_REB_communism: "The Netherlands"
```
AngriestBird: Shouldn't this be "The United Provinces" all the way down?

tools/validation/validate_unused_scripted.py
```diff
@@ -55,6 +55,9 @@
     re.compile(
         r"^should_(not_)?activate_active_crypto_bonuses$"
     ),  # Vanilla engine crypto hooks — engine-read override points
+    re.compile(
+        r"^set_partial_draft_effect$"
+    ),  # Conscription-law preset kept for parity with its three called siblings
```
AngriestBird: We should streamline this so it's not add a new re.compile but an array we can loop over for every element

tools/validation/validate_unused_scripted.py
```diff
@@ -83,6 +92,14 @@
         "MD_regional_triggers.txt",
         "00_scripted_effects.txt",
         "00_law_blocking_effects.txt",
+        "00_budget_effects.txt",
+        "00_ct_effects.txt",
+        "00_economic_system_utilities.txt",
+        "00_generic_ideas_scripted_effects.txt",
```
AngriestBird: Did you confirm this? This is a lot of space here to flag as false positives.

AngriestBird (changes requested):

tools/validation/validate_unused_scripted.py
TemplarGeneral: They were for me atleast mechanical triggers/effects for MD mechanics for devs to just use instead of remake but its up to you what to do with it

AngriestBird (approved):

PR #2767, opened 2026-08-05
perf(startup/gui/on_actions): eliminate redundant startup loops, zero-assignments, shift daily ticks, and decouple national GUIs (#2765)

### LINKED ISSUE #2765: [TASK] On Startup | See if we can eliminate the "every_country" for the missile startup
**Description**

There is an every_country in the on startup taht just sets some variables to 0. This can probably be removed and it will help boot time for weaker computers. 

**Additional context**

hasan4938:
Resolves #2765.

### Performance Optimizations Included:
1. **Startup Optimization (#2765)**:
   - Merged the secondary duplicate `every_country` loop inside `on_startup` (`common/on_actions/00_on_actions.txt`) directly into the primary `on_startup` loop.
   - Removed redundant `set_variable = { protest_strength = 0 }`, `set_variable = { protest_radicalisation = 0 }`, and `set_variable = { anti_bully_wars = 0 }` assignments from the startup loop across 200+ nations.

2. **On-Actions Tick Optimization**:
   - Shifted NEPAD Africa cleanup check in `common/on_actions/99_FRA_on_actions.txt` (`on_daily_FRA`) and Badakhshan resistance check in `common/on_actions/99_TAJ_on_actions.txt` (`on_daily_TAJ`) to weekly ticks, eliminating daily tick overhead.

3. **National GUI Decoupling**:
   - Replaced `dirty = global.md_gui_dirty` (which forced daily UI re-evaluations across all nations) with nation-specific dirty variables (`dirty = global.update_TAG_ui`) across all national scripted GUIs (USA, SOV, TUR, UKR, ENG, ITA, ISR, LBA, PER, AFG, ALG, ARM, BOS, CUB, CZE, EH, GRE, HEZ, IRQ).
   - Added startup guard for Counter-Terrorism initialization (`counter_terror_initialized`).

### DIFF
--- a/common/on_actions/00_on_actions.txt
+++ b/common/on_actions/00_on_actions.txt
@@ -1673,12 +1673,6 @@ on_actions = {
 					multiply_variable = { debt = 0.0125 }
 				}
 
-				#Init protest variables, this should stop error spam
-				set_variable = { protest_strength = 0 }
-				set_variable = { protest_radicalisation = 0 }
-
-				set_variable = { anti_bully_wars = 0 }
-
 				# Sets Global vars for investment then resizes array
 				set_variable = { project_build_amount^-1 = 1 }
 				set_variable = { constructable_buildings = 1 }
@@ -1751,6 +1745,12 @@ on_actions = {
 				set_variable = { last_quarter_gdp_total = gdp_total }
 				set_variable = { last_quarter_gdpc_total = gdp_per_capita }
 
+				calculate_starting_mio_size = yes
+
+				if = { limit = { is_ai = yes }
+					yearly_investment_targets_routine = yes
+				}
+
 				if = { limit = { is_ai = no }
 					add_to_array = { global.player_nations = THIS.id }
 					graph_initial_setup = yes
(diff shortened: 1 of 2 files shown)


common/on_actions/00_on_actions.txt
```diff
@@ -1728,12 +1673,6 @@ on_actions = {
 					multiply_variable = { debt = 0.0125 }
```
AngriestBird, line 1673: you are missing the one for sats I think. I think that can be removed in generally since that one itself is just setting values to 0. I want to say it was on_startup_sat  or something. Your previous PR properly guarded it.

common/on_actions/99_TAJ_on_actions.txt
AngriestBird, line 6: I would check with Droid that this does not cause any content issues.

common/scripted_guis/99_AFG scripted_gui.txt
```diff
@@ -4,7 +4,7 @@ scripted_gui = {
 		context_type = selected_state_context
 		window_name = "OTT_Infopanel"
 		parent_window_token = selected_state_view
-		dirty = global.md_gui_dirty
```
AngriestBird, line 7: Switching is a good idea, but if these have any interactions at all they will be forever blocked by the fact these values are never incremented.

AngriestBird (changes requested):

hasan4938:
Everything should be correct now.

PR #2777, opened 2026-08-06
Egypt, Estonia and energy event cleanup + date-scheduling validator

TemplarGeneral:
### Changes
- Estonia's Victory Column unveiling, the 2000 diplomat expulsions, the founding of Res Publica, and the founding of the Estonian United Left Party now fire on their historical dates
- Fixed Estonia's ETV2 broadcast event never firing, it was scheduled inside Spain's country scope instead of Estonia's
- Removed ten Estonian events that had no text and nothing to fire them
- Removed the unreachable uranium enrichment demand event chain, its three opinion modifiers, and its localisation
- Fixed declining Israel's offer in the Egyptian chain firing the acceptance event instead of the rejection
- Nations importing Egyptian Evergreen fertilizer can now cancel the contract through a new decision
- Fixed the civilian industry loss and relocation notifications never reaching the player, and the notification events re-running the industry change they were meant to report
- Added a validator check that flags events carrying a date guard that nothing schedules from 00_yearly_effects.txt


tools/validation/validate_unused_scripted.py
```diff
@@ -81,6 +81,10 @@
         "enable_debt_rate_payments",
         # 00_ct_effects.txt
         "add_new_org",
+        # 00_economic_effects.txt - industry loss/relocation library, deliberately
+        # callerless; a caller sets reduction_factor then invokes one of these.
+        "remove_civilian_industry_effect",
```
AngriestBird: Just delete these. WE dont' use these effects at all and never have.

Changelog.txt
```diff
@@ -157,6 +157,8 @@ v2.0.0
   - Added a new La Resistance mission "Lobby Government" that lets you temporarily disrupt the target nations ability to combat your influence
   - Focuses that send an offer to another nation now preview what you receive if it is accepted, across 40 country trees
   - [HEZ] Added two focuses for Asymmetric Warfare
+  - [EST] The Victory Column unveiling, the 2000 tit-for-tat diplomat expulsions, the founding of Res Publica, and the founding of the Estonian United Left Party now fire on their historical dates
```
AngriestBird: IMO we don't need to add changelogs for all this.

AngriestBird (changes requested):

AngriestBird (approved):

PR #2772, opened 2026-08-05
France Rework: new focus tree, regime paths and counter-terrorism

Blazer135:
### Changes
- Rebuilt France's political focus tree around the real party system, with a shared opener branching into the Socialists, the UDF/LREM centre, the Republicans, the Gaullist RPF, the National Front, France Unbowed, the Greens and the Communists, growing the tree from 200 to 382 focuses
- Removed the old Egalite and Fraternite branches and the neo-colonial staging paths for Quebec, Louisiana, Indochina, North and West Africa, the Horn, the Caribbean and the Pacific bases, replaced by the new party branches
- Added four ways to end the Fifth Republic: the French Commune, a Sixth Republic, a restored Kingdom under a Bourbon or Orleanist claimant, and a Bonapartist Empire
- Added an extremist infiltration mechanic on the far-left and far-right paths, ending in a choice between arresting the movement or letting it seize power, with a Republic Is Safe route for either crackdown
- Added a counter-terrorism system with Vigipirate levels, states of emergency, Operation Sentinelle, intelligence surges and anti-terror raids, where attack severity scales with the intelligence and police budgets
- Added a green extraction ban with its own events, idea and reversal decision
- Replaced France's static political ideas with political, economic and social balance dynamic modifiers, plus separate agenda modifiers for the communist, royalist, Bonapartist and world revolution paths
- Added Joint Weapon Projects: the SCAF and EMBT programmes with Germany and the ATOM IFV programme with Russia
- Added a Liberal Arms Export Policy focus with arms contract events for Saudi Arabia, the UAE, Serbia, Greece, Egypt and India
- Added EU assistance decisions to fund civilian industry, infrastructure and the wider economy in Germany, Greece, Italy, Belgium, Spain and Poland, with a separate green-investment set
- Added foreign policy decisions for rejoining NATO's integrated command, seeking a Russian partnership, recognising Palestine, condemning the invasion of Ukraine, vetoing EU federalisation, a non-aligned initiative and a Frexit campaign
- Added expansion and war decision sets for the Commune, Empire and Kingdom paths, including Red Europe, the Eighth Coalition, the Frankish Empire and the crusade chains
- Added the Paris Accord faction template and a hidden Francosphere faction for French-aligned blocs
- Added 303 French events covering the party branches, the regime changes, terrorist attacks, pension reform and the new flavour chains
- Rewrote every French party description in the politics view and added the Party for the Proletarian Revolution, the National Rally and the House of Orleans as selectable parties
[...]

Issue #2755, opened 2026-08-04
[TASK] Ukrainian space MIO (Yuzhnoye/Pivdenmash)


Qwerty483LOL:
### Background

The original ask was for two scripted effects so a Ukrainian MIO could make OLVs cheaper and about 15 days faster to build. Per @TemplarGeneral's comment, this should be a modifier rather than per-MIO scripted effects.

Half of that already existed. The space system (OLV, GNSS, COMSAT, SPYSAT, KILLSAT) has build-time modifiers - `olv_production_speed_modifier` and its four siblings in `common/modifier_definitions/missile_modifier_definitions.txt` - folded into the days math in `calculate_production_vars_10..14` (`common/scripted_effects/00_missiles_scripted_effects.txt`).

There was no equivalent for price: `var_gui_prod_NN_price_total` was unit price x quantity with no country modifier applied, so nothing in the mod could make a launch vehicle or satellite cheaper.

### Done in #2778

Five country modifiers mirroring the existing speed modifiers, applied in the same five price formulas (MD cost convention: `color_type = bad`, negative = cheaper), plus token registration, English localisation and doc entries:

- `olv_production_cost_modifier`
- `gnss_production_cost_modifier`
- `comsat_production_cost_modifier`
- `spysat_production_cost_modifier`
- `killsat_production_cost_modifier`

### Remaining

The Ukrainian space MIO itself. `common/military_industrial_organization/organizations/MD_UKR_organizations.txt` has no Yuzhnoye/Pivdenmash entry - that is a separate design task (branches, traits, icons, localisation).

Once it exists, the trait tree can grant both halves directly, no scripted effects needed:

```
olv_production_cost_modifier = -0.15   # 15% cheaper
olv_production_speed_modifier = 0.15   # 15% fewer days
```

Same pattern for the `gnss_`, `comsat_`, `spysat_` and `killsat_` families.

TemplarGeneral:
I think we should make a modifier for it

AngriestBird:
I added this with the #2780 so this task is done.

PR #2778, opened 2026-08-06
Add space production cost modifiers

TemplarGeneral:
### Changes
- Added five space production cost modifiers so content can make launch vehicles and satellites cheaper to build, matching the build-time modifiers that already existed for OLV, GNSS, COMSAT, SPYSAT and KILLSAT
  - Part of #2755
- Fixed the "OLV Contruction Speed" modifier name typo

The France orphaned event cleanup that was originally bundled here has been reverted per review, to avoid conflicting with the France Rework (#2772).


common/national_focus/france.txt
```diff
@@ -2351,6 +2351,21 @@ focus_tree = {
 				leave_faction = yes
 			}
 			create_faction_from_template = faction_template_paris_accord
+			for_each_scope_loop = {
```
AngriestBird: This is the same as if doing a every_cuontry without the loop. It would be more performance friendly to user every_other_country here.

AngriestBird (commented):
Most of these are fine but I would usggest reverting the France changes. France rework is getting ready for merge and this would cause unneeded conflicts.

AngriestBird (approved):

PR #2812, opened 2026-08-07
UN Fixes

MrP0tter:
Changes a large part of the UN math to use math expressions

Tweaked one scaling multiplier, so that bonuses from subsidies will be actually useable now, and not just non-existant.


common/scripted_effects/00_money_system.txt
```diff
@@ -6960,18 +6960,28 @@ update_optional_contributions = {
...
+	}
+
+	set_variable = {
+		global.uncdf_budget = {
+			value = 0
+			add = base_budget_number
+		}
+	}
 
 	for_each_loop = {
```
AngriestBird: If we convert the united nations to a collection we could in theory optimize this further by iterating. We'd need to benchmark but this *could* be faster.

common/scripted_effects/01_international_systems_effects.txt
AngriestBird, line 1678: We could fold this check in with the values if we want to get these in here using if subsidies bonus x > greater than 0 and we could look at getting riid of the subsidy flags.

AngriestBird (dismissed):

AngriestBird (approved):

PR #2817, opened 2026-08-08
Even More Bits of Performance

MrP0tter:
Part two of the expressions work


common/scripted_effects/99_neigbour_system_effects.txt
```diff
@@ -1,4 +1,12 @@
 # Author(s): Angriest Bird
+refresh_country_and_neighbors_effects = {
```
AngriestBird: This one runs n x n x n. I would revert this back. The data staleness here is acceptable.

common/scripted_effects/00_space_automation_effects.txt
```diff
@@ -1677,7 +1677,6 @@ space_auto_tech_watch = {
 }
 
 space_automation_monthly = {
-	space_automation_weekly = yes
```
AngriestBird, line 1680: Need to re-add this back.

common/on_actions/99_LBA_on_actions.txt
```diff
@@ -680,7 +637,50 @@ on_actions = {
...
+			200 = LibyaCivilWarContent.20	#Jihadist militias
+			200 = LibyaCivilWarContent.21	#Islamist militias
+			200 = LibyaCivilWarContent.22	#Secularist militias
+			200 = LibyaCivilWarContent.23	#Reformist militias
+		}
+	}
+
 	on_weekly_LBA = {
+		#These happen when the "Anti-Gaddafi Protests" is active
+		random_events = {
```
AngriestBird, line 666: Move this back to a regular on_weekly. This was in an on_weekly due to an issue with the tags. we can fix this later with a tag alias.

AngriestBird (changes requested):

common/on_actions/01_tfv_on_actions.txt
```diff
@@ -292,6 +299,10 @@ on_actions = {
 			set_temp_variable = { influence_target = ROOT.id }
 			change_influence_percentage = yes
 			research_slot_init_for_country = yes
+			update_neighbors_effects = yes
+			if = { limit = { is_in_array = { global.arabic_countries = THIS.id } }
+				AS_register_arab_spring_country = yes
```
AngriestBird: This can be unified into the ingame_startup_nations_effect I think. We can have this set here so we can reduce the dupliaction of this in these files since this is used all over the place.

AngriestBird (changes requested):

common/scripted_effects/99_neigbour_system_effects.txt
MrP0tter: Was reverted

AngriestBird:
Independent review of the full diff against main. No blocking bugs found.

Worth noting: the naval-hegemon rework fixes a live bug on main. The weekly staggered `assign_power_ranking_ideas` calls `remove_power_ideas` on any tier change, which strips `naval_power_of_the_continent` until the next quarterly pass. Dropping the naval idea from that list closes the hole.

**Verified equivalent**

- `update_state_variables`: all 27 accumulators in the new `every_collection_element` loop match the old per-variable sums, including the `maximum_renewable_energy_var = 1` seed and `last_overall_productivity` captured before zeroing. `num_missile_silo` / `num_anti_air_building` temps are still set before their only consumer (`ingame_update_setup` runs `update_state_variables` then `update_military_rate`).
- `recalculate_resources_price`: the `PREV.sum_*` temp accumulation matches the established pattern (landmark tourism income in `00_money_system.txt`, `!_energy_effects.txt`).
- Naval tracking edge cases: old saves bootstrap via the in-loop `has_idea` check; region membership is exclusive (if/else chain in `register_in_power_ranking_array`) so no dual-region holder can be stripped by one region while owed by another; nothing removes a live country from a regional array, so the recorded holder is always visible to the bootstrap; `remove_ideas` on a dead id is a no-op. The `largest > 0` guard also fixes the old `var:0` scope-in when a region has no deployed navies.
- Regional powers: top-N counts unchanged (7/3/3/7/4), clear/set cadence unchanged, and the new `clear_temp_array` guards are required and present now that the partial sort no longer drains `pr_sort_*`.
- Libya: all of `LibyaCivilWar.1-14` are gated on `original_tag = LBA` plus the protests mission, so the `on_weekly_LBA` move preserves probability exactly. Keeping the content events global is right, since civil-war splinters are not tag LBA.
- Space gate: the flag OR is exactly the union of `space_automation_weekly`'s internal gates (`space_auto_reconcile_lines` needs a line flag, the body needs `space_auto_active`). The AI path goes through `space_auto_ai_monthly`, which bypasses the on_action gate.
- `update_neighbors_effects`: the monthly `every_country` refresh in `on_monthly` is untouched, so the new event-driven calls only shorten stale windows after border changes. Scopes are correct, including the `FROM` exists-guard on `on_peaceconference_ended`.

**Findings (low severity)**

1. `common/on_actions/01_tfv_on_actions.txt`: `on_release_as_puppet` and `on_release_as_free` already run `update_neighbors_effects` via `ingame_startup_nations_effect` (`00_scripted_effects.txt:2398`). The new explicit calls re-run the full neighbor recompute in the same chain. Idempotent so harmless, but dead weight in a perf PR. Same for the `on_liberate` path that takes the startup-effect branch.
2. `common/scripted_effects/00_power_ranking_effects.txt`: array tails beyond the sorted top-N prefix are now in registration order. No consumer indexes by rank, but order-sensitive scans exist, e.g. the cyber AI `look_and_add` walk with `break` (`00_cyber_ai_effects.txt:93`) now considers tail candidates in registration order instead of contribution order. Behavior note rather than a bug; flagging so the AI-preference shift is a conscious choice.
3. Cosmetic: the six temp-var zero resets at the end of `power_ranking_evaluate_all_regions` are unnecessary. Nothing later in any calling chain reads them, and each sort clears its own state.

Pre-existing and unchanged by this PR: the standalone `update_military_rate` calls at `00_on_actions.txt:2622/2639` run without `update_state_variables` earlier in the chain, so the silo/AA temps read 0 there. Same before and after.

AngriestBird (approved):

PR #2820, opened 2026-08-08
event cleanup

KianGhk1530:
### Changes
Removed useless events from the following countries: Russia, Karabakh, Central Asia (Shared), Palestine, Gulf states (Shared), Tajikistan

Fixed bugged events for the following countries: Tajikistan


KianGhk1530:
and also algeria

TemplarGeneral (approved):

PR #2886, opened 2026-08-11
Mandate fix

Blazer135:
Fixed the Mandate and re-wrote how the system work ish, now it also reacts to world events like the American Invasion of iraq and also Afghanistan

### DIFF
--- a/common/decisions/05_north_korea.txt
+++ b/common/decisions/05_north_korea.txt
@@ -916,8 +916,6 @@ NKO_arduous_march = {
 			}
 		}
 
-		cost = 50
-
 		custom_cost_text = cost_0_5
 		custom_cost_trigger = { NKO_spa_can_afford_money_05 = yes }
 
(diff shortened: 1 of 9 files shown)


common/decisions/05_north_korea.txt
```diff
@@ -916,41 +916,19 @@ NKO_arduous_march = {
...
-			}
 		}
 
 		ai_will_do = {
-			base = 100
-			modifier = {
-				factor = 0
-				has_idea = NKO_spa_bonus_industry
-			}
+			base = 0
```
AngriestBird, line 931: Intentional to disable the AI here?

common/on_actions/99_NKO_on_actions.txt
```diff
@@ -38,6 +70,73 @@ on_actions = {
...
+
+	on_peaceconference_ended = {
+		effect = {
+			if = {
+				limit = {
+					ROOT = { original_tag = USA }
+					FROM = { original_tag = IRQ }
+					country_exists = NKO
+					NKO = {
+						has_country_flag = NKO_spa_iraq_war_response_fired
```
AngriestBird, line 108: Wouldn't this be easier to add this with the other check in the Iraq war stuff rather then making your own? Seems needless in this file and should migrate to the other file where this is initially triggered..

PR #2889, opened 2026-08-12
feat: Purchase fuel automatically when low

### LINKED ISSUE #2460: [TASK] Automation | Deficit-aware automatic fuel purchase
**Description**
`energy_automatic_fuel_purchase` re-buys every 28 days unconditionally once queued. Make it deficit-aware so it only fires (or scales its amount) when the country actually runs short.

Split from #2370.

**Additional context**
- Mission: `energy_automatic_fuel_purchase` (`common/decisions/!_energy_decisions.txt`, timeout re-activates itself)
- Purchase effect: `buy_fuel_from_the_market_effect` (`common/scripted_effects/!_energy_effects.txt`)
- Deficit trigger: `has_an_energy_deficit` (`common/scripted_triggers/05_energy_scripted_triggers.txt`)

### Task List
- [x] Gate or scale the timeout purchase on deficit
- [x] Tooltip explaining the condition

SBeausoleil:
### Changes
- Only buy fuel automatically IF the country is at less than 20% fuel capacity
- Doubled the amount of fuel bought per `buy_fuel_from_the_market_effect` effect

Closes #2460


common/decisions/!_energy_decisions.txt
```diff
@@ -25,25 +24,25 @@ decision_category_energy = {
...
-
 		days_mission_timeout = 28
 		fire_only_once = no
 		selectable_mission = yes
 		is_good = yes
 
 		timeout_effect = {
-			log = "[GetDateText]: [Root.GetName]: Decision energy_automatic_fuel_purchase"
-			buy_fuel_from_the_market_effect = yes
+			if = { limit = { fuel_ratio < 0.2 }
```
AngriestBird, line 40: Actually, I think we can do this a bit better by moving this into on actions. AT the moment this doesn't retrigger if you have over 20% fuel ratio. We should have it set where on_weekly it checks whether it has the automatic flag + whether the fuel ratio is there. In this example the automatic trigger here stops if you are greater then 20% fuel ratio with no automatic retrigger.

AngriestBird (changes requested):

AngriestBird (approved):

PR #2907, opened 2026-08-13
Add satellite deorbit controls

### LINKED ISSUE #2643: [BUG] Satellite Not Being Destroyed
> *Reported by **discord-sync** via [Discord](https://discord.com/channels/210890013334831104/1532038544107765790) on 2026-07-29*
> *Tags: General Bug*

**Describe the bug**
Playing til 2034, reached Satelite cap of 3000. How do i destroy old satelites to make room for more?

**To Reproduce**

**Screenshots**

**Operating System**

**Additional context**

Blazer135:
## Summary

- add owner-only single-satellite deorbiting with a confirmation click
- add bulk deorbiting for the selected satellite model, showing the affected count before confirmation
- reuse the existing satellite removal and system-refresh paths, including SPYSAT mission safeguards
- enforce the 3,000 active-satellite limit from live constellation totals so deorbiting frees launch capacity

Closes #2643.

## Validation

- [x] staged diff contains only four missile/orbit files
- [x] scripted GUI bindings and localisation references validate
- [x] brace/style, localisation, variable-use, common-mistake, and commit-stage content validators pass
- [x] English localisation remains UTF-8 with BOM
- [x] in-game: deorbit an owned GNSS/COMSAT/SPYSAT/KILLSAT satellite and confirm counts/system values refresh
- [x] in-game: bulk-deorbit one model and confirm only that model is removed
- [x] in-game: verify foreign satellites cannot be deorbited and active SPYSAT missions remain protected
- [ ] in-game: reach the 3,000 active-satellite limit, deorbit one satellite, and launch its replacement
image

### DIFF
--- a/Changelog.txt
+++ b/Changelog.txt
@@ -82,6 +82,7 @@ v2.0.0
     - Orbit view now has filter buttons to make finding and targeting specific satellites easier
     - Nation selection UI has been adjust so that it now displays the name and flag of the country, and not just the flag
     - Production & Payload menus now display what version a satellite is, along with the name.
+  - Added owner-only single and model wide satellite deorbit controls with confirmation, live affected counts, SPYSAT mission safeguards, and stale ASAT target cleanup
   - Russia's "Annexation of Crimea" is made more flexible so as to increase the options available to the Russian player
   - Power Ranking system now is calculated 2 times a year (May and November) for performance-oriented purposes
   - Extended the Generic tree for more economic and military options
--- a/common/scripted_effects/00_missiles_scripted_effects.txt
+++ b/common/scripted_effects/00_missiles_scripted_effects.txt
@@ -2720,20 +2720,26 @@ remove_satellite_from_system = {
 }
 
 remove_deorbited_satellite_target_reference = {
-	set_temp_variable = { MD_deorbit_target_removed = 0 }
 	for_each_loop = {
-		array = sat_target_array
-		value = v
-		index = i
-		break = MD_deorbit_target_removed
-		if = {
-			limit = {
-				check_variable = { v = var_sat_ID }
-				check_variable = { sat_target_TAG_array^i = ROOT.id }
+		array = global.countries
+		value = targeting_country
+		var:targeting_country = {
+			set_temp_variable = { MD_deorbit_target_removed = 0 }
+			for_each_loop = {
+				array = sat_target_array
+				value = v
+				index = i
+				break = MD_deorbit_target_removed
+				if = {
+					limit = {
+						check_variable = { v = ROOT.var_sat_ID }
+						check_variable = { sat_target_TAG_array^i = ROOT.id }
+					}
+					remove_from_array = { array = sat_target_TAG_array index = i }
+					remove_from_array = { array = sat_target_array index = i }
+					set_temp_variable = { MD_deorbit_target_removed = 1 }
+				}
 			}
-			remove_from_array = { array = sat_target_TAG_array index = i }
-			remove_from_array = { array = sat_target_array index = i }
-			set_temp_variable = { MD_deorbit_target_removed = 1 }
 		}
 	}
 }
--- a/common/scripted_effects/00_missiles_scripted_effects.txt
+++ b/common/scripted_effects/00_missiles_scripted_effects.txt
@@ -2322,6 +2322,7 @@ add_satellite_from_payload = {
 	}
 	set_satellites_gui = yes
 	clear_array = payload_array
+	refresh_bulk_deorbit_state = yes
 }
 
 ###################################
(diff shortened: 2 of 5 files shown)


AngriestBird (commented):
Review focused on maintainability, simplification, and performance. One edge-case bug, the rest cleanup.

## Findings

**1. Medium: stale `MD_bulk_deorbit_count` can bypass the SPY-mission safeguard**
`bulk_deorbit_selected_model` (`common/scripted_effects/00_missiles_scripted_effects.txt`) guards with the count computed at staging time in `stage_bulk_deorbit_selected_model`, but removes all *current* satellites of the model at confirm time. If time passes unpaused between the two clicks, in-flight launches arriving via `add_satellite_from_payload` (on_actions) can raise the model's live count above the staged count. The guard then checks the stale count against `var_SPY_mission_num_available` while the effect removes more satellites than were vetted, which can deorbit SPY sats committed to active missions. The `Confirm ([?MD_bulk_deorbit_count|0])` label is also wrong in that window.
Fix: recompute the count (rerun the counting loop) at the top of `bulk_deorbit_selected_model` before the guard. The unchecked validation item ("active SPYSAT missions remain protected") is the test that would catch this; run it with game time advancing between stage and confirm.

**2. Medium (maintainability): the same 3-line state reset is copy-pasted into 15 click handlers**
`clr_country_flag = MD_confirm_deorbit_satellite` + `clear_bulk_deorbit_state = yes` + set/clr of `MD_satellite_orbit_selection_active` appears in all six `sat_N_button_click` handlers and nine navigation handlers in `common/scripted_guis/00_missiles_scripted_guis.txt`. Extract two scripted effects next to `clear_bulk_deorbit_state` (one that clears all deorbit state, one that additionally sets the selection flag). Removes ~45 lines, and the next handler that needs the reset can't silently forget one of the three lines.

**3. Low (simplification): `<=` emulated with `OR = { < , = }` twice**
Both the guard in `bulk_deorbit_selected_model` and `confirm_bulk_deorbit_model_button_click_enabled` write `check_variable = { MD_bulk_deorbit_count < var_SPY_mission_num_available }` OR'd with `=`. Use the long form instead: `check_variable = { var = MD_bulk_deorbit_count value = var_SPY_mission_num_available compare = less_than_or_equals }`. Same fix in both places; also removes drift risk between the trigger and the effect guard.

**4. Low (performance): one global sweep per satellite in bulk deorbit**
`bulk_deorbit_selected_model` calls `remove_deorbited_satellite_target_reference` inside the per-satellite loop, and that effect iterates `global.countries` and each country's `sat_target_array`. Bulk-deorbiting a few hundred satellites late-game means hundreds of full country sweeps with scope switches in a single click. Hoist it: run one sweep after the removal loop that drops every `sat_target_array` entry whose TAG matches `ROOT.id` and whose orbit slot now reads 0 (removal zeroes slots, so staleness is detectable), or collect removed IDs in a temp array and match in one pass. The single-deorbit path can keep the current form.

**5. Low (simplification): duplicated cleanup tails**
`deorbit_selected_satellite` and `bulk_deorbit_selected_model` end with the same clear-selection/clear-flags/`country_event = satellites.9` sequence; fold into one shared effect alongside item 2.

**6. Trivial: localisation inconsistency**
`TOTAL_SAT_NUM_TT_DELAYED` says "3000" while the reworked `satellites.8.d` says "3,000". Pick one.

Minor UX note: if the specific staged satellite is destroyed (ASAT/KILLSAT) between stage and bulk-confirm, `confirm_bulk_deorbit_model_button_visible` fails while the staged flag still suppresses the stage button, so both buttons vanish until the player re-clicks a satellite. Recomputing state at confirm (item 1) is the natural place to also self-heal this.

## Verified clean

- Bulk-loop index stability: `remove_satellite_from_orbit` zeroes `orbit_array` slots rather than removing them, so iterating while removing is safe, and system-before-orbit ordering is preserved in both new flows.
- `satellites.8` is a `news_event` and `satellites.9` exists (hidden, runs `update_space_system_values` -> `set_satellites_gui`), so counts and system values refresh correctly after deorbit.
- `ROOT` is preserved inside the `global.countries` sweep, and the nested `for_each_loop`s reusing default `v`/`i` temp vars don't produce stale reads.
- The 3,000 gate refreshes `var_*_sat_total` via `set_satellites_gui` before summing, counts additions incrementally, and reports the limit event once per batch. All navigation paths that change `orbit_selected_TAG` now reset the confirm state.

Recommendation: fix item 1 and run the unchecked SPYSAT-protection test before merge; items 2-4 are worth doing here while the code is new.

AngriestBird (changes requested):

AngriestBird (approved):

PR #2678, opened 2026-08-01
Thailand: new focus tree, national content and Preah Vihear rework

cal-chu:
### Changes
- Adds a unique Thailand focus tree: 385 focuses, five political paths from the 2000 start (Thaksinist Populism, Royalist Establishment, Democrat Establishment, Reform Wave and an irredentist Greater Thailand branch), a communist entry, and a shared economic and diplomatic trunk. Thailand previously used the generic tree
- Adds 158 Thai national spirits, 115 decisions across 14 categories and 233 events covering the Tom Yum Kung recovery, the Thaksin years, the coups, the Deep South insurgency and the 2020s protest wave
- Adds the Deep South insurgency: an intensity meter driving a tiered spirit and weekly incidents, answerable with security operations, an autonomy framework, or a mix
- Adds a coup-risk system for populist governments, surfacing as a tiered spirit and a defusable mission
- Adds the Populist Tide, a rural-support meter that grows at a rate set by the economy and stability, lifts the Thaksinite vote against the whole non-aligned bloc, and bites at two tiers. At 100 the red shirts occupy central Bangkok and the government must let them stay or clear them out. The Democrat side answers it with rural spending, and the levers cost real money
- Adds the Kra Canal as a three-phase funded construction project with cost overruns and sabotage on contested ground
- Adds a royalist no-coup path so the network monarchy can rule through the courts, the appointed senate and coalition kingmaking instead of tanks
- Adds a trade diplomacy spirit that scales foreign influence reach with signed trade agreements, deliberately weaker than Singapore's equivalent
- Thailand now starts under Chuan Leekpai's Democrats with Thai Rak Thai surging in opposition, and the 2001 race is won on the campaign rather than starting near the finish line
- The Thaksinite party renames itself TRT, PPP and Pheu Thai across eras, New Aspiration and Chart Pattana merge into it, and the leader rotation grows from 6 to 9 sub-ideology branches so Yingluck, Srettha, Paetongtarn, Pita and Natthaphong arrive through elections
- Each dissolution and re-registration is now an on-screen event rather than a silent rename, and party names are gated on the founding flag instead of a hardcoded date, so a tree that runs ahead of or behind history still shows the right banner. Adds the People's Party as the third incarnation of the reform movement
[...]

PR #2942, opened 2026-08-15
Thailand: remove three unused events, fix 2005 landslide seat count

cal-chu:
## What

Removes three `is_triggered_only` Thailand events with no callers, as flagged by `tools/validation/validate_events.py` (`unreferenced-triggered-only`), and fixes a wrong seat count in the 2005 election content.

**Dead events (`events/Thailand.txt`)**
- `thailand_news.12` "The Canal Is Open" — superseded by `thailand_news.20`, whose own comment says it "replaces the old news.12 fire at the capstone". `SIA_two_oceans_one_kingdom` fires news.20.
- `thailand.230` "Engineers at Songkhla" — the canal-vs-land-bridge fork, superseded by `SIA_kra_gateway` → `thailand.380`. Its stated hook `SIA_the_southern_question` fires `thailand.350` now, and `thailand.380` option c already "folds the land-bridge flavor".
- `thailand.206` "The Red Wave" — superseded by `thailand.223`, which `SIA_the_2005_landslide` fires. The focus already grants the PP and party-popularity payload `.206` duplicated.

All three came in with #2678 and were each replaced by a later event in that same PR, so this is leftover iteration, not missing wiring.

**Also removed**
- The `SIA_kra_route_canal` branch in `thailand.383.a`. That flag was only ever set by `thailand.230.a`, so the cost-overrun surcharge was already unreachable.

**Seat count**

Thai Rak Thai won 377 of 500 seats on 6 February 2005, not 374. Corrected in the focus title `SIA_the_2005_landslide` ("The 374-Seat Sweep" → "The 377-Seat Sweep"), its description, and `thailand.223.d`.

`thailand.223` is retitled "The Network Is Alarmed" → "Three Hundred and Seventy-Seven". The old title named a mood rather than an image, and leaned on McCargo's *network monarchy* term that the event never explains.

English loc only — non-English is left alone per AGENTS.md.

## Why

`.206` was worth a second look rather than a straight delete, because the chain does jump from the focus to the establishment's reaction with no election-night beat in between. It still goes, because its text puts "red-shirted supporters" at a February 2005 election: the UDD formed after the September 2006 coup and red only became the protest colour around 2007. `SIA_red_shirts_mobilized` is where that imagery belongs and already exists. Reviving it as-is would also have stacked its +10% party popularity on the focus's own +5%.

## Validation

Swept `events/Thailand.txt` for unreferenced events after the change — zero remaining out of 198. `thailand.19` still reads as orphaned to a naive grep but is fired from the `random_events` weight list in `common/on_actions/00_on_actions.txt`.

Nothing player-visible changes from the deletions; none of that code could execute.

Closes #<issue number here>


MrP0tter:
Did a cursory review; this looks good. I'm handling the other event cleanups right now, so I'll leave these three alone since you have a PR up for them already. Undraft once youre ready for this to be merged in, and ignore the variable validator issue, I'm handling that as well.

MrP0tter (approved):

PR #3025, opened 2026-08-18
EU Math Collections

### LINKED ISSUE #2274: [TASK] Performance | Update the EU systems to math expressions + collections
**Description**

Updating the math expressions + collections for the European Union to have faster performance per month where needed. Clean up duplicate references. 

**Additional context**

MrP0tter:
Closes #2274

Converts what parts of the EU can be from arrays + normal math to collections & math expressions.

Fixed a display bug.


common/collections/collections.txt
```diff
@@ -1,5 +1,30 @@
 # Collections
 
+eu_potential_countries_collection = {
+	input = constant:country_groups.eu_potential_countries
+	name = COLLECTION_EU_POTENTIAL_COUNTRIES
+}
+
+eu_members_collection = {
+	input = game:all_countries
```
AngriestBird: This should use the constant country groups since we have it and would reduce the amount of collection element refresh.

AngriestBird (commented):
common/collections/collections.txt:6 — blocker — This filters founding members out of the only collection copied into `global.EU_potential` at startup (`99_eu_scripted_effects.txt:812-815`). That array is still the `ai_enabled` gate for the parent EU GUI and both parliament/council AI contexts (`01_european_union_guis.txt:68-86`, `4259-4266`, `4601-4608`), while each `ai_check` requires `EU_member`, so every starting member now fails the automation gate. The same omission also skips founding members in the monthly breach check, parliament cleanup, EU-wide Europeanism changes, and map-mode targets. Later joiners remain in the startup snapshot, making behavior depend on when a country joined. Keep `global.EU_potential` as the full available-country superset, or change all member consumers to `global.EU_member` before filtering this collection.

AngriestBird (approved):

PR #3061, opened 2026-08-19
Standardize Free States focus identifiers

AngriestBird:
## Summary
- Rename legacy Free States focus identifiers to use the `USB_` prefix.
- Update the focus tree and English localization references.

## Motivation
The focus standards validator rejected the lowercase `free_*` prefixes.

## Checks
- Free States focus standards validation
- Focus tree structural validation
- `git diff --check`


2ndLTLuige2010 (approved):
looks good/less dumb lol

PR #3069, opened 2026-08-19
Public sector workforce system

TemplarGeneral:
### Changes
- The public sector now competes for your workforce: government administration, police, healthcare, education and social services each employ workers scaled to their spending law
- Public sector staffing is drawn after industry, so a labour-poor nation understaffs its state before its factories
- An understaffed public sector costs stability, and an understaffed administration reduces political power gain
- The workforce and worker fulfillment tooltips now break out police, education and social services alongside the existing sectors, and show overall public sector staffing
- Higher corruption inflates how many civil servants the administration needs to run at full strength
- Economic regulation laws now also raise or lower the staffing that administration, police, education and social services require, on top of their effect on building workforce needs
- Fixed the Healthcare Workforce Requirement modifier colouring a higher requirement as a good effect


AngriestBird:
Noted internal discussions this will come at a later time.

PR #3157, opened 2026-08-22
Fix UN observer status and UNSC eligibility

### LINKED ISSUE #1469: [TASK] United Nations | Permanent Observer Status + Non State Actors on UNSC
**Description**

The Vatican among other nations are not actually members of the United Nations. They need to be granted something to prevent them from being in the UN but  they are fully recognized countries. 

**Additional context**

MrP0tter:
## Context

Fixes #1469. The Holy See was being added to the United Nations as a full member, and stale Council state could retain ineligible actors.

## What

- Treat the Holy See as a recognized UN observer.
- Exclude unrecognized countries and non-state actors from UNSC voting and seating.
- Clear stale permanent-member state during UN cleanup.


common/scripted_effects/01_international_systems_effects.txt
```diff
@@ -426,7 +426,10 @@ un_recog_setup_check = {
 		}
 	} else_if = {
 		# SWI is a recognized non-member until the 2002 referendum decision (SWI_join_the_united_nations)
-		limit = { NOT = { original_tag = SWI } }
+		limit = {
+			NOT = { original_tag = SWI }
```
AngriestBird, line 434: Noting that we need handling for the new Untied Nations content for Switzerland

AngriestBird (changes requested):

AngriestBird (approved):

PR #3135, opened 2026-08-22
Add Corporate History and OEM Industry Systems

JasonBreen:
## What

This adds 32 Corporate History chains across 13 countries, four independent OEM and compute subsystems, a separately selectable Linux ecosystem, and an American economic and dashboard layer.

- Adds Full, Outcomes Only, and Off/Disabled game-rule controls with later-start reconstruction.
- Turns industry history into player choices, persistent national outcomes, follow-up decisions, crises, and read-only dashboards.

## Why

This contribution addresses the scope requested in [#2027](https://github.com/MillenniumDawn/Millennium-Dawn/issues/2027). It turns the history of computer, semiconductor, telecommunications, finance, and console industries into national capabilities and player decisions instead of isolated flavor events.

Explicit game rules and silent reconstruction make that content predictable in normal campaigns, later starts, and campaigns that reduce or disable authored events.

## Mechanics

| Rule | Script options | Player-visible behavior |
| --- | --- | --- |
| Corporate History | `full`, `outcomes_only`, `disabled` | Full reconstructs passed milestones, schedules current and future events once, and enables decisions, crises, and dashboards. Outcomes Only silently establishes reached outcomes without story popups or crises. Disabled creates no Corporate History state while native systems continue normally. |
| Linux Ecosystem | `full`, `outcomes_only`, `off` | Full enables Linux events, national integrations, programs, and economic effects. Outcomes Only silently reconstructs reached adoption and support milestones. Off creates no Linux state or effects. |

Linux remains independently selectable. All nine rule combinations are valid; disabling Corporate History does not disable native Belt and Road content or force Linux state.

Linux lifecycle work is bounded to Brazil, China, France, Germany, India, Poland, Russia, the United Kingdom, the United States, and Venezuela. Each participant uses its country-specific monthly host; there is no global Linux callback or registry.

```mermaid
flowchart LR
    A["Owner-local monthly driver"] --> B{"Game-rule mode"}
    B -->|"Full"| C["Reconstruct passed milestones"]
    C --> D["Schedule current and future events once"]
    D --> E["Expected / pending / resolved markers"]
    E --> F["Bounded recovery if delivery is interrupted"]
    B -->|"Outcomes Only"| G["Silently reconstruct reached outcomes"]
    B -->|"Off"| H["Create no subsystem state"]
    F --> I["One route and one terminal outcome"]
    G --> I
    I --> J["Dashboards and derived systems read bounded state"]
```

[...]

PR #3149, opened 2026-08-22
Add Ericsson Corporate History for Sweden

JasonBreen:
## What

Adds a Sweden-only Ericsson Corporate History chain with three strategic milestones in 2001, 2012, and 2022. This is the first and smallest tag-by-tag extraction from the closed full-system PR.

- Adds Full, Outcomes Only, and Off controls for Corporate History, including silent later-start reconstruction.
- Connects each Ericsson route to Sweden's existing digital modifier so choices leave persistent national effects.
- Uses existing upstream event art and introduces no new image assets.

## Why

This follows the requested split of #3135 into small, tag-focused contributions and continues the industry-history work proposed in #2027. The chain turns Ericsson's handset partnership, smartphone-era restructuring, and Vonage strategy into choices about specialization, vertical integration, investment, and research instead of a run of passive flavor notifications.

The earlier separate LTE and 5G popups were removed during the de-duplication pass. Their gameplay purpose is represented through the strategic network routes and research outcomes in the three retained decisions.

## Mechanics

| Rule | Player-visible behavior |
| --- | --- |
| **Full** | Presents the three choices at their milestones. A campaign beginning after a milestone silently reconstructs its historical persistent route without replaying the popup or one-time reward. A delivery interrupted by owner collapse gets one bounded retry. |
| **Outcomes Only** | Silently reconstructs reached historical routes and persistent effects. It shows no Corporate History popups and grants no one-time political-power, treasury, or research rewards. |
| **Off** | Creates no Ericsson Corporate History state and schedules no Ericsson events. Native Swedish content remains unchanged. |

| Date | Decision | Lasting trade-off |
| --- | --- | --- |
| **2001** | Form Sony Ericsson or retain Ericsson handsets | Foreign-investment efficiency versus domestic digital productivity |
| **2012** | Exit consumer handsets or reinvest in a controlled device business | Project economics and immediate liquidity versus research capacity |
| **2022** | Build the Vonage API platform or concentrate on radio networks | Office/investment returns versus research and productivity |

One-time treasury, political-power, and research effects make the immediate choices distinct. Persistent effects are idempotent and feed the existing `SWE_digital_modifier`; they do not create a second Swedish economy system.

## How

- The normal path is registered in the existing 2001, 2012, and 2022 yearly effects and executed only in the original Swedish tag.
- A bounded `on_monthly_SWE` driver initializes the selected rule mode, reconciles a milestone missed after the January pulse, and recovers an interrupted expected delivery once. It performs no global country scan.
[...]

Issue #2027, opened 2026-06-22
[SUGGESTION] OEM industry development options


millenniumdawnmod-hub:
> *Reported by **discord-sync** via [Discord](https://discord.com/channels/210890013334831104/1518420911408025691) on 2026-06-22*
> *Tags: Mechanic, Economy, Feature, Minor Content*

**Description**
What does this add?: While the microchips category in the beta is great, it doesnt add anything to make the player as though they're building the modern OEM industry, (HP/Compaq, Dell, Lenovo, etc.), or chip OEMs like NVIDIA or AMD.

Detailed description: add content regarding the history of 2000's computing, and the consequences of compaies such as compaq and packard bell dying. i.e. add decisions, events, etc.

Consequences of adding?: More content for players regarding the advancement of computing after 2000 rather than just have research nodes.

Who will want to use this feature?: compsci turbonerds/historians or just the average md player

Link to why this is historical: https://www.weforum.org/stories/2020/11/heres-how-technology-has-changed-and-changed-us-over-the-past-20-years/

**Additional context**
Auto-created from a Discord forum post.

millenniumdawnmod-hub:
**AI Triage**

Suggestion to add decisions and events about 2000s computing history, including OEM companies like Compaq and Packard Bell, to complement the existing microchips category.

Severity: :green_circle: Low · Suggested labels: `suggestion:content`, `suggestion:economy`, `suggestion:events`, `economy`, `events`, `from-discord`

**AI Work Estimate**
Priority: **Low** · Effort: **Medium**
*Estimated by the bot — maintainers can adjust the Priority/Effort fields on the issue.*

**Could the reporter clarify?**
- Would you like this content tied to the existing microchips tech tree, or introduced as a standalone decision tree/decisions system?
- Should the events focus on company bankruptcies/mergers (e.g. Compaq-Compaq's HP merger, Packard Bell's decline) or broader consumer computing milestones like the rise of pre-built desktops?

millenniumdawnmod-hub:
**Reply from Discord:**

tying it to existing microchip tech tree wouldnt be a bad idea, but there should be economic decisons regarding these, the microsoft court case decision is good example

millenniumdawnmod-hub:
**Reply from Discord:**

also to the second thing the bot asked, yes, the events should focus on company bankruptcies/mergers/expansion, and the rise of nvidia, intel, and amd

millenniumdawnmod-hub:
[...]

PR #3183, opened 2026-08-24
Add Siemens Corporate History for Germany

JasonBreen:
## What

Adds a Germany-only Siemens Corporate History chain covering six milestones: the millennium strategy, the Nokia Siemens Networks venture, the networks crossroads, the Healthineers listing, the Siemens Energy spin-off, and the closing deconsolidation.

Built to the lean shape from your review on #3149 rather than converted to it afterwards.

## Why

Siemens is the one European industrial group whose whole period is a sequence of decisions about what to keep. Between 2000 and 2025 it went from a conglomerate spanning automation, power, transport, and telecommunications to a much narrower company, and every step of that was a choice that could have gone the other way.

The chain also removes a cross-country dependency. In its original form the network venture was a negotiation: Germany waited on Finland's proposal, answered it, and both sides read each other's flags. That made either chain unlandable without the other. The venture is now Germany's own decision, made from German flags, so this PR stands alone and the Finnish chain in #3184 stands alone beside it.

## Mechanics

| Rule | Player-visible behavior |
| --- | --- |
| **Full** | Presents the six choices at their historical dates. |
| **Outcomes Only** | Applies the historical route and its persistent effects silently, without popups or the live events' one-time costs or rewards. |
| **Off** | Creates no Siemens state and schedules no events. Native German content is unchanged. |

| Date | Decision | Lasting trade-off |
| --- | --- | --- |
| **2001** | Reinforce the integrated group or prioritize digital industry | Command and control research versus computing research |
| **2007** | Fold the networks arm into a joint venture or press industrial automation | Internet research versus political capital and automation research |
| **2013** | Sell the networks business or take full control of it | Four billion recovered and computing research versus four billion spent and network depth |
| **2018** | List Siemens Healthineers or keep it integrated | Three billion raised versus tighter control of its research and capital |
| **2020** | Complete the Siemens Energy spin-off or keep the division | Four billion raised versus an unbroken engineering portfolio |
| **2025** | Advance deconsolidation or preserve the integrated portfolio | Resolves the route into one of four permanent shapes |

The closing milestone awards one of four terminal ideas depending on the accumulated route: a focused industrial group, a digital industries company, an automation and control core, or a diversified conglomerate that never broke itself up.

## How

- Events live in `events/Germany.txt` under the existing `germany_yearly` namespace; there is no separate Siemens event file.
- `00_yearly_effects.txt` is the only driver. Each milestone is one effect that either fires the event under Full or writes the historical route under Outcomes Only. There is no monthly driver and no second scheduling path.
- Ten country flags, two per milestone, representing the routes and nothing else. No delivery bookkeeping and no accumulated score variables.
- The awarded idea is the authoritative terminal state, so no parallel completion flag is kept. Outcomes are idempotent and guarded on that idea.
- The rule is read inline; no scripted-trigger abstraction layer for it.
- Four of the six milestones charge the treasury on one route and not the other, so the AI gate zeroes the paying option while a bankruptcy collapse mission is running. The free option's historical-focus zero yields during collapse, so the pair is never both zero. The opening milestone charges the same either way and is therefore left alone.
- Uses existing upstream event art. The two Siemens-specific images from the source branch are dropped in favour of `GFX_generic_factory`, `GFX_stock_market`, `GFX_powerplant`, and `GFX_hospital`.

## One change outside Germany

`MD_game_rules_l_english.yml` gets a two-line correction. The Corporate History rule descriptions that landed with #3149 still promise that the rule "silently reconstructs milestones already passed" and "recovers an interrupted event delivery once". That machinery was removed in the lean rework, so the rule currently documents behaviour it no longer has. The wording now matches what the rule does.

## Checks

- `precommit_validate.py` across the changed files: characters, common mistakes, events, ideas, MIOs, OOB units, style. All passed.
- `validate_variables`, `validate_localisation`, `validate_gfx_references`, `validate_modifiers`, `validate_simplifications`, `validate_scripted_params`, `validate_on_actions`, `validate_unused_scripted`, `validate_technologies`: no findings against the new content.
- Option labels checked against the state each option sets, so no button names a route it does not take.
- `ai_chance` evaluated across all four historical and bankruptcy states: no all-zero option set, and no collapse state that steers the AI onto the paying route.
- Localisation cross-check: all referenced keys resolve, no duplicates, and the capstone's fourth option is named `.e` so it does not collide with the event's own `.d` description key.
- No new image assets.

## Related

Refs #2027. Supersedes the Germany/Siemens portion of closed #3135.


events/Germany.txt
```diff
@@ -21418,3 +21418,381 @@ country_event = {
...
+	trigger = {
+		GER_siemens_outcome_settled = no
+	}
+
+	option = {
+		name = germany_yearly.305.a
+		trigger = {
+			has_country_flag = GER_siemens_healthineers_listed
+			has_country_flag = GER_siemens_energy_spun_off
+		}
```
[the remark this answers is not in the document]
JasonBreen: Confirmed and fixed in e98281f11e. You had it exactly: the capstone read four flags and neither of the 2013 ones was among them, so that milestone could not change what Siemens became.

I took the first option rather than documenting 2013 as non-terminal, because selling the networks business is a divestiture of the same kind as the Healthineers listing and the Energy spin-off. It now joins those two in the test for a focused industrial group, which also matches the history: Siemens sold its Nokia Siemens Networks stake in 2013, listed Healthineers in 2018 and spun off Energy in 2020, and the focused outcome is where that lands.

Checked afterwards that all thirty two combinations of the five routing flags still resolve to exactly one option, so the capstone stays exhaustive and mutually exclusive, and that the historical AI still reaches the focused outcome. The distribution shifts as you would expect, with the focused ending now needing all three divestitures rather than two.

common/ideas/GER_siemens_ideas.txt
```diff
@@ -0,0 +1,55 @@
+ideas = {
+	country = {
+		GER_siemens_focused_industrials = {
+			picture = production_bonus
+			allowed = { original_tag = GER }
```
AngriestBird: Similar cleanup. Allowed's here in this one is something to be removed.

common/scripted_triggers/99_GER_scripted_triggers.txt
```diff
@@ -0,0 +1,9 @@
+# The awarded idea is the authoritative outcome; no parallel flag is kept.
+GER_siemens_outcome_settled = {
```
AngriestBird, line 2: Unify with the existing GER_scripted_triggers or renamed to GER_scripted_triggers.txt

AngriestBird (changes requested):

common/ideas/GER_siemens_ideas.txt
JasonBreen: Done in 4cad6fe0bb. You are right, and this one slipped: I applied the allowed drop to the other seven tags after your review on #3193 but never came back to Germany and Finland. Both are fixed now, so all nine chains are consistent. allowed_civil_war is kept, as on the others.

common/scripted_triggers/99_GER_scripted_triggers.txt
JasonBreen: Done in 4cad6fe0bb, renamed rather than unified, because Germany has no country trigger file on main to unify into. I checked and the closest are 99_FRA_diplomatic_triggers.txt and 99_GRE_scripted_triggers.txt, with no GER file at all.

So it is now 99_GER_scripted_triggers.txt, matching the forty other country files rather than the bare GER_scripted_triggers.txt, since every existing one carries the 99_ prefix.

Two other branches had the same per-company filename and I fixed them at the same time rather than waiting for you to find them: Finland is now 99_FIN_scripted_triggers.txt in 46bbac03cb and Taiwan is 99_TAI_scripted_triggers.txt in 07cbb2b2cd. Finland also still had its allowed blocks and has now lost them. The remaining six chains already fold their trigger into an existing country file, so those three were the only ones left.

AngriestBird (approved):

PR #3193, opened 2026-08-25
Add the remaining American corporate histories

JasonBreen:
### Changes

- Adds nine American corporate histories: Dell, Oracle, NVIDIA, Micron, Texas Instruments, IBM, Apple, Google, Motorola and Microsoft
- Drives every chain from the yearly pulse, gated on the Corporate History game rule, so Full presents the choice and Outcomes Only writes the historical route silently
- Puts every chain's effects in 99_USA_main_scripted_effects.txt, the ideas in one USA_corporate_history_ideas.txt, and the triggers in the existing USA scripted trigger file

Two shapes are present, which the earlier version of this description flattened into one and got wrong. Eight of the chains, NVIDIA and Micron and Texas Instruments and IBM and Apple and Google and Motorola and Microsoft, run six milestones and close with a capstone that resolves the route into one of four national ideas. Dell and Oracle instead follow the shape Hewlett-Packard already uses on main, carrying their outcomes as country flags and the us_economic_policies dynamic modifier with no terminal ideas, and Oracle runs five milestones rather than six. Thanks to the Copilot reviewer for catching that.

This replaces #3176 (Dell) and #3181 (Oracle). Both were opened per company, and both were left conflicting when Hewlett-Packard merged, because all three touch the same three files. Folding them into one tag PR fixes that permanently: each tag now lands as a single commit series, so the chains cannot conflict with one another. Every event and effect from those two PRs is present here unchanged, so nothing is lost by closing them.

Worth flagging honestly: you approved the format on around sixteen events across Hewlett-Packard, Dell and Oracle. This brings fifty nine, because the other seven chains were already built and staged behind them. The shape is repetitive rather than varied, so the review should be shorter than the line count suggests, but it is a lot more than you have seen so far and I would rather say so than have you find it.

The four software histories are deliberately shaped as groundwork for a later AI race system, so that it can start from what the industry actually chose rather than from nothing. Apple records whether the silicon is owned, Google records frontier versus staged deployment and owned versus external accelerators, Microsoft records datacentre ownership and in-house versus partnered models, and IBM records the open hybrid position. Nothing here depends on that system existing.

Events occupy usa_economic_events 310 to 405, clear of Hewlett-Packard at 300.


common/scripted_effects/00_yearly_effects.txt
```diff
@@ -1802,6 +1874,10 @@ trigger_year_2020_events = {
 	}
 }
 trigger_year_2021_events = {
+		USA_ti_milestone_2021 = yes
```
AngriestBird, line 1967: Not wrapped in USA

common/ideas/USA_apple_ideas.txt
```diff
@@ -0,0 +1,59 @@
...
+			}
+			research_bonus = {
+				CAT_computing_tech = 0.03
+				CAT_internet_tech = 0.02
+			}
+		}
+
+		USA_apple_resilient_silicon_network = {
+			picture = production_bonus
+			allowed = { original_tag = USA }
```
AngriestBird: allowed's aren't needed. We can remove these. I would suggest also unifying the idea files.

common/scripted_triggers/USA_apple_triggers.txt
```diff
@@ -0,0 +1,9 @@
+# The awarded idea is the authoritative outcome; no parallel flag is kept.
```
AngriestBird: Unify the scripted trigger files.

common/scripted_effects/00_yearly_effects.txt
```diff
@@ -1706,6 +1772,8 @@ trigger_year_2018_events = {
 	}
 }
 trigger_year_2019_events = {
+	USA = { USA_microsoft_milestone_2019 = yes }
+		USA_oracle_milestone_2019 = yes
```
AngriestBird, line 1864: not wrapped in USA

AngriestBird (changes requested):

common/scripted_effects/00_yearly_effects.txt
JasonBreen: Fixed in e46107c78e, and there were nine of these rather than two. They came from the Dell and Oracle branches, where the calls were correctly nested inside an existing USA block; folding the tags together replayed them at the year anchor, which kept the indentation and dropped the scope. Every USA milestone call is now consolidated into a single USA block per year, sixty four across twenty five years, and I checked structurally that none sit outside one.

common/scripted_effects/00_yearly_effects.txt
JasonBreen: Same fix as the other one, e46107c78e. This was one of nine unscoped calls, all now inside a single consolidated USA block per year.

common/ideas/USA_apple_ideas.txt
JasonBreen: Done in e46107c78e. The allowed blocks are gone from all thirty two ideas, and the eight per-company files are now one USA_corporate_history_ideas.txt. I kept allowed_civil_war since it looks like the commoner convention upstream, a hundred and six idea files use it against thirty nine for allowed, but say the word and it goes too.

common/scripted_triggers/USA_apple_triggers.txt
JasonBreen: Done in e46107c78e. The eight per-company trigger files are folded into the existing common/scripted_triggers/99_USA_scripted_triggesrs.txt rather than a new file, so it is nine triggers in there now. Worth mentioning that filename has a typo upstream, triggesrs, which I left alone as out of scope for this PR.

common/scripted_effects/00_yearly_effects.txt
```diff
@@ -433,6 +440,7 @@ trigger_year_2002_events = {
 	UZB = { country_event = { id = tajik_lore.9 days = 270 } }
 }
 trigger_year_2003_events = {
+		USA_oracle_milestone_2003 = yes
```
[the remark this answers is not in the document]
JasonBreen: Already fixed at the head. This review ran against d52b957775; e46107c78e consolidates every USA milestone call into a single USA scope block per year, sixty four across twenty five years, and I verified structurally that none sit outside one. Good catch though, and it found more than the six cited: there were nine unscoped calls in total, all from the Dell and Oracle branches where they were correctly nested inside an existing USA block before the tag fold replayed them at the year anchor.

common/scripted_effects/99_USA_main_scripted_effects.txt
```diff
@@ -944,3 +944,1206 @@ USA_hp_milestone_2020 = {
...
+		}
+		custom_effect_tooltip = { localization_key = modifies_dynamic_modifier_tt MODIFIER = us_economic_policies }
+		add_to_variable = { econ_productivity = 0.01 tooltip = productivity_growth_modifier_tt }
+		set_country_flag = USA_dell_ai_selective_margin
+	}
+}
+
+# One entry point per milestone, called from the yearly pulse. Full presents the
+# choice; Outcomes Only writes the historical route silently and grants nothing.
+USA_dell_milestone_2007 = {
```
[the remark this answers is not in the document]
JasonBreen: You are right and the description was wrong, now corrected. Eight of the chains run six milestones and close with a capstone resolving into one of four ideas. Dell and Oracle instead follow the shape Hewlett-Packard already uses on main, carrying outcomes as country flags and the us_economic_policies dynamic modifier with no terminal ideas, and Oracle runs five milestones rather than six. The code is right; my summary flattened two shapes into one.

common/scripted_effects/00_yearly_effects.txt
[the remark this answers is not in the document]
JasonBreen: Correct on all nine, and already fixed at the head. This review ran against d52b957775; e46107c78e consolidates every USA milestone call into a single USA scope block per year, sixty four across twenty five years, verified structurally so none sit outside one.

Your reading of the consequence is the right one and worth recording: the yearly meta-effect runs from on_monthly, so the current scope is whichever country first enters the guarded monthly block, and those calls would have applied Dell, Oracle, Micron and TI state to an arbitrary country including its treasury and flags. They came from the Dell and Oracle branches, where the calls sat correctly inside an existing USA block; folding the tags together replayed them at the year anchor, which kept the indentation and dropped the scope.

AngriestBird (approved):

PR #3210, opened 2026-08-25
Fix two corporate history defects on main: the Foxconn terminal and ten unrewarded paid options

JasonBreen:
### Changes

- Fixes the Foxconn Outcomes Only terminal so it awards the same idea the Full historical route reaches
- Gives ten paid corporate history options a research bonus, so they stop being pure penalties

Two corporate history defects that reached main. Both were found by Codex reviews on the chain PRs, one of them after the chain had already merged.

The Foxconn one first. Outcomes Only records TAI_foxconn_sharp_acquired at the 2016 milestone, and taiwan.305 tests that flag before contract scale, so the Full historical route resolves to the electric vehicle supplier while the 2024 terminal hard-coded the factory city. The two game rule settings produced different permanent national ideas from the same history. Outcomes Only exists to reproduce the historical route silently, so it should award what that route reaches, and the electric vehicle supplier is also the historically right answer. Codex raised this on #3205 and the fix missed that merge by about a minute.

The second is broader. Codex then found on #3206 that paid options which grant nothing read as penalties rather than choices. Sweeping every chain in the series turned up ten more on main:

| Chain | Options | Cost |
| --- | --- | --- |
| Taiwan | 300.a, 301.a, 302.a, 303.a, 304.a | 15, 25, 20, 20, 25 political power |
| Canada | 331.a, 331.b, 334.b | 4.00, 10 political power, 2.00 |
| United States | 383.b, 400.b | 10, 15 political power |

Every one of the five Foxconn milestones was affected, and canada.331 is the only event in the series where both options cost and neither rewarded. Each now grants a research bonus with a category matching the choice: industry for the mainland network and the Zhengzhou campus, electrical for the Sharp recapitalisation, construction for Wisconsin, computing for the patent fight and the antitrust appeal, encryption for the enterprise mobility pivot, and AI research for the defence contract Google walked away from.

Free options that grant nothing are deliberately left alone. Those are do-nothing branches and several gain resources instead, so they are real choices rather than defects.

Every category was checked against common/technology_tags/00_technology.txt before use, after #3199 found four invalid ones. I re-swept all nine touched events across the four historical and bankruptcy states afterwards: no state leaves every option at zero weight and no ahistorical branch keeps weight under historical focus. Style, events and localisation validators are clean.

The same missing-bonus defect was present on the Russian and Ukrainian chains and is already fixed on #3204 and #3206.


JasonBreen:
@AngriestBird one note on ordering, since there are five of these open at once: this is the only one that fixes something already live on `main` rather than adding new content.

Verified at `fcd4de80e1`:

- `TAI_foxconn_milestone_2016` sets `TAI_foxconn_sharp_acquired` under Outcomes Only, `common/scripted_effects/99_TAI_scripted_effects.txt` line 75
- `TAI_foxconn_milestone_2024` then applies `TAI_foxconn_apply_factory_city`, line 97
- but `taiwan.305.c`, the Full route to that same idea, is gated on `NOT = { has_country_flag = TAI_foxconn_sharp_acquired }`

So Outcomes Only awards a permanent national idea that its own recorded history makes unreachable under Full. Same game rule, same events, two different outcomes, and a Taiwan campaign running Outcomes Only past 2024 gets the wrong one today. The fix is the single line at 97.

The other nine files here are the unrewarded-option sweep and are not urgent. That one line is, because the chain merged in #3205 and it cannot be fixed on any other open branch.

_🤖 Posted by [Claude Code](https://claude.com/claude-code)_

AngriestBird (approved):

PR #3221, opened 2026-08-26
Optimize North Korea's SPA mechanic

### LINKED ISSUE #3220: Simplify and optimize North Korea's SPA mechanic
**Context**

The current two-active-mandate model for North Korea's Supreme People's Assembly (SPA) has avoidable recurring checks and several state/UI inconsistencies: focus-lifted penalties can be reintroduced if SPA unlocks afterward, the banner can remain stale after a focus lift, delayed flavor events can outlive the mechanic, and some English descriptions no longer match the implemented modifiers.

**What**

Keep the existing player loop—eight domains, at most two active mandates, 365-day duration, current costs, rewards, and flavor—while making its state transitions reliable, its UI immediately accurate, and its runtime work proportional to actual SPA actions. Simplify duplicated state and redundant paths in the current mandate implementation, and use authoritative focus and idea state for resolved domains. Legacy session code and compatibility with older saves are out of scope.

**Acceptance**

- Completing a domain's penalty-lifting focus before or after SPA unlock never reintroduces that penalty.
- No more than two mandates can be active during normal operation; the displayed count remains accurate.
- The SPA banner refreshes immediately when a mandate starts, expires, or a focus permanently resolves a domain.
- Civil war or collapse removes SPA state once and prevents already-scheduled SPA flavor events from granting later rewards.
- English descriptions match the implemented research and economy modifiers and accurately describe flavor-chain limits.
- Duplicated state, redundant cleanup paths, and recurring checks in the current mandate implementation are simplified without changing mandate costs, duration, rewards, or the eight-domain selection loop.
- Legacy session code remains unchanged.
- No migration, recovery, or compatibility handling is added for saves created before this change.

Blazer135:
Closes #3220

**What**

- Enable the Supreme People's Assembly and all eight standing domain penalties from the 2000 start.
- Simplify mandate state changes, focus resolution, and collapse cleanup.
- Refresh the SPA banner on every current state transition and guard delayed flavor rewards after collapse.
- Align English SPA descriptions with implemented modifiers and event limits.

**Why**

SPA is a foundational North Korea mechanic, and gating it behind an early focus added one-time setup code for only a short delay. The current implementation also performed redundant checks, duplicated permanent focus state, and could leave stale UI or delayed rewards after collapse.

**How**

North Korea's history now seeds the umbrella idea, eight unresolved penalties, and zero active mandates, then schedules the existing onboarding event for day one. The one-shot initializer and Humanitarian Aid unlock wiring are removed. Authoritative focus completion controls permanent resolution, including the Russian-focus bypass path; the existing civil-war on-action owns cleanup. Legacy session code and older-save compatibility remain out of scope.


common/decisions/05_north_korea.txt
```diff
@@ -945,13 +945,7 @@ NKO_arduous_march = {
 			}
 		}
 
-		ai_will_do = {
-			base = 100
-			modifier = {
```
AngriestBird, line 950: Why remove the AI will do here?

common/ideas/north_korea.txt
```diff
@@ -924,7 +920,6 @@ ideas = {
 		NKO_spa_debuff_production = {
 			picture = generic_production_bonus
 			allowed = { original_tag = NKO }
```
AngriestBird: Remove unneeded allowed = { }

AngriestBird (changes requested):

AngriestBird (approved):

PR #3227, opened 2026-08-26
Link TSMC's growth to Taiwanese and American resource trade

### LINKED ISSUE #2151: [SUGGESTION] TSMC MIO Addition
> *Reported by **discord-sync** via [Discord](https://discord.com/channels/210890013334831104/1520882088197226656) on 2026-06-28*
> *Tags: Minor Content*

**Description**
TSMC on level up/select traits should modify an idea for both the USA and Taiwan allowing for more resources to be traded per civilian factory.

This has a gameplay benefit of giving the USA an incentive to keep Taiwan alive (cheaper microchips/composites per civ) and Taiwan the ability to buy the resources needed for said chips without having to make half their island civs to feed the eternal composite hunger.

Taiwan's  largest importing and exporting nation is the USA, with an extreme dependence on running large portions of their economy, it would be nice to reflect this dependence in game.

**Additional context**

JasonBreen:
### Changes
- TSMC's 130nm Production Lines, Advanced Lithography Procedures and 14nm Production Lines traits now raise Taiwan's resources traded per civilian factory to +1, +2 and +3
- The United States receives the same bonus, held only while Taiwan still exists, still controls Taipei, and is not at war with America

Closes #2151


common/military_industrial_organization/organizations/MD_TAI_organizations.txt
```diff
@@ -930,6 +970,26 @@ TAI_tsmc_manufacturer = {
 					add_to_variable = { tai_tsmc_energy_use_var = adder }
 				}
 				custom_effect_tooltip = tai_chip_energy_tt
+				hidden_effect = {
+					if = {
+						limit = { check_variable = { TAI_tsmc_resource_trade_cost_bonus_per_factory < 3 } }
+						set_variable = { TAI_tsmc_resource_trade_cost_bonus_per_factory = 3 }
```
AngriestBird, line 976: How does the value get updated? Do we have any checks whether or not Taiwan can just stop this trade?

AngriestBird (dismissed):

common/military_industrial_organization/organizations/MD_TAI_organizations.txt
JasonBreen: afaik each of the three traits writes the variable in on_complete, guarded by
check_variable = { x < n } so a tier can only go up, never down. 14nm has seven
parents so without that guard picking up a lower trait later would knock you
back. the dynamic modifier reads the variable live, so it tracks without being
re-added. on the second point, no. taiwan has no way to cut it off short of war. the only
gate is the enable block on USA_tsmc_access_modifier: taiwan has to exist, still
control taipei, and not be at war with the usa. taiwan's own bonus isn't gated
at all. working on giving taiwan actual agency there now.

JasonBreen: added two decisions on the taiwan side - restrict and resume semiconductor
exports to the us. restricting sets a flag the USA_tsmc_access_modifier enable
block checks, costs 50 pp and an opinion hit from washington, and cant be
reimposed for 180 days. so taiwan can now actually use chip access as leverage
instead of only losing it by getting annexed.

went with a decision rather than hanging it off trade agreements because the ai
never signs those, so the usa bonus would have been dead in most games.

common/decisions/taiwan.txt
```diff
@@ -177,4 +177,55 @@ TAI_category = {
...
+		cost = 25
+
+		complete_effect = {
+			log = "[GetDateText]: [Root.GetName]: Decision TAI_tsmc_resume_us_exports"
+			clr_country_flag = TAI_tsmc_us_export_ban
+		}
+
+		ai_will_do = {
+			base = 10
+		}
```
SBeausoleil, line 233: Suggestion: add a check that Taiwan is independent and has positive US opinion. I am concerned that otherwise the AI could ban, and then quickly unban US export even though the conditions that led to the ban are still there.

localisation/english/MD_decisions_l_english.yml
```diff
@@ -725,6 +725,10 @@
  TAI_CHI_confidence_building_desc: "Open channels of dialogue with Beijing and pursue cross-strait confidence-building measures, easing tensions and raising mutual trust across the strait."
  TAI_CHI_assert_sovereignty: "Assert Our Sovereignty"
  TAI_CHI_assert_sovereignty_desc: "Reaffirm Taiwan's sovereignty through public statements, deepened ties with foreign partners, and a firm line toward Beijing. Cross-strait trust will suffer for it."
+ TAI_tsmc_restrict_us_exports: "Restrict Semiconductor Exports to the United States"
```
SBeausoleil:
```suggestion
 TAI_tsmc_restrict_us_exports: "Restrict Semiconductor exports to the United States"
```

common/decisions/taiwan.txt
JasonBreen: good catch, that was a real bug. the resume decision had an unconditional
ai_will_do base of 10, so an ai taiwan would ban at opinion < -50 and then lift
it again on the next evaluation with none of the conditions actually changed.

resume now carries weight only once opinion recovers above -25, so theres a 25
point dead band between the two thresholds. restrict still has days_re_enable =
180 on top of that. also widened the independence check from "not a us puppet"
to is_subject = no.

localisation/english/MD_decisions_l_english.yml
JasonBreen: applied, and did the same to the resume decision so the pair matches.

one thing worth flagging - the rest of MD_decisions_l_english.yml is title case
("Campaign for Lower Government Spending", "Discourage Population Growth"), so
these two are now the odd ones out. happy to flip both back if youd rather keep
the file consistent.

SBeausoleil (approved):

PR #3236, opened 2026-08-27
Stop poor nations locking to minimum spending desire

### LINKED ISSUE #3226: [BUG] Some nations have only minimum spending desire
**What happens**
What you saw, and what you expected instead.

**Where**
Unknown, spending logic most likely.

**Versions**
Game version and checksum, MD version (release number, or the dev branch date you pulled).
Game version: 1.19.2
MD version: DEV, pulled on 2026-08-26 around 15:00, but was first noticed on 2026-07-16

**To Reproduce**

1. Start any new game on the dev branch. Use default rules.
2. Let a few months pass (I verify after 4 months, but my first game on the dev version lasted until 2026 and I confirm the issue was still present)
3. Tag to switch countries. You'll notice that several have no to very little spending demand, leading them to reduce to zero their investment across all spending categories.

**Operating System**

- [x] Windows
- [ ] Linux
- [ ] Mac

**Additional context**

Image

The exact countries affected vary, however India and China are pretty much always affected.
I notice consistently:

- Chad
- China
- Congo DR
- Cuba
- Egypt
- India
- Indonesia
- Iran
- Kenya
- Liberia
- Mali
- Niger
- Nigeria
- Romania
- Uzbekistan
- Vietnam

A common thread I found is that they are all countries with a very low per-capita GDP. Gamewise that kinda makes sense, extremely poor populations have low expectations from their governments. However this lead to serious debuffs to very large and developping nations such as China and India which are supposed to develop into superpowers over time. These nations then become permanently handicaped because they are only spending the barest of minimums.

AngriestBird:
## Bottom line

Poor nations were pinned to level 1 in every budget category forever. Two defects stacked: the desire offsets were being erased before anything could read them, and the GDP/C penalty was harsh enough to floor the entire developing world on its own. Closes #3226.

## Why

`calculate_expected_spending` opens by zeroing `expected_*` and rebuilding them from the party table plus the GDP delta. In `on_monthly` it ran 27 lines *after* `recalculate_law_desires`, so every offset that effect applied was wiped before the AI, the protest triggers, or the budget tab ever saw it.

That was harmless until #1439. The `wants_bureau_03`-style flags used to snapshot the post-offset value and survive the reset for the rest of the month; replacing them with direct `check_variable = { expected_adm_spending … }` reads pointed every consumer at the offset-free baseline. Dead as a result: `increase_spending_if_possible`, every focus and decision `change_expected_*_spending` grant, and the whole threat-driven military boost.

It also deadlocked. Once one offset hit 1, `has_*_law_desired` went false so the ratchet stopped adding, while the AI still read baseline and never raised the law.

Separately, a flat -1.8 for everything under $5k GDP/C put the whole developing world at the floor by itself. Weighted party bases average 2.25 to 2.9, so tier 0 landed at 0.5 to 1.4 and clamped to 1 in all six categories. India: welfare 2.59 - 1.8 = 0.79. That one bucket also lumped Somalia in with Ukraine, a 30x spread.

## How

**Single writer.** The offsets and the military threat logic now live inside `calculate_expected_spending`, alongside `modifier@expected_*_modifier` in the same formula. `recalculate_law_desires` no longer writes `expected_*` at all; it only runs the AI decisions (despair reset, ratchet, deficit cuts) off finished numbers. The five near-identical `recalculate_*_desire` wrappers and their duplicated clamps are gone, and the ordering hazard with them. This also fixes `on_puppet` and the tag-switch path, which called the recompute without ever applying offsets.

**GDP tiers.** Nine per-capita bands instead of seven, split at 1 and 2.5 with a smoother ramp below 15k. Everything above 15k keeps its current values exactly. On top of that, an economy-size term keyed to `gdp_total`, clamped so it can only offset a poverty penalty and never adds to wealth. That is what separates a developing giant from a country a hundredth its size while leaving the genuinely destitute at the floor.

**Military.** Desire now weighs `potential_and_current_enemies^num` at two thresholds plus `we_are_being_justified_against_or_wargoal_against_us`, replacing an `any_home_area_neighbor_country` scan that was both narrower and more expensive. Peacetime ceiling goes 4 to 5, which the offset path already reached by design.

Also folded in: `round_variable` was rounding the offset instead of the total it had just been added to, and the ratchet's `add_to_variable` bypassed the clamp that `change_expected_*_spending` applies. Both were inert while the offsets were dead and are live now.

Note this is a balance change as well as a bug fix. Every sub-$15k/c nation's opening budget shifts.

BLUF

### DIFF
--- a/common/on_actions/MD_on_actions.txt
+++ b/common/on_actions/MD_on_actions.txt
@@ -1016,7 +1016,6 @@ on_actions = {
 				every_country = {
 					recalculate_party = yes
 					gov_pop_calculation = yes
-					recalculate_law_desires = yes
 					update_neighbors_effects = yes
 
 					civilian_microchip_consumption = yes
@@ -1044,6 +1043,7 @@ on_actions = {
 					}
 
 					calculate_expected_spending = yes
+					recalculate_law_desires = yes
 					apply_protest_effects = yes
 
 					un_check_if_valid_for_sanctions = yes
--- a/common/scripted_effects/00_desired_law_calculations.txt
+++ b/common/scripted_effects/00_desired_law_calculations.txt
@@ -1,164 +1,9 @@
 # Author(s): Angriest Bird, Mkemper, Crocomoth
 
-# Bureaucracy desire is chiefly determined by the size/strength of the nation. More powerful the nation, more it wants a higher level of bureaucracy.
-# Secondary to size and power of the nation is the GDP of the nation - impoverished nations can't afford an extensive bureaucracy.
-recalculate_bureaucracy_desire = {
-	add_to_variable = { expected_adm_spending = bureau_desire_offset }
-	round_variable = bureau_desire_offset
-	clamp_variable = {
-		var = expected_adm_spending
-		min = 1
-		max = 5
-	}
-}
-
-# Police desire should tend to gravitate toward the middle (police_03) in most nations. This is also affected by politics - repressive governments
-# want more police apparatus than more liberal/libertarian ones. Also influenced by GDP - impoverished nations cannot afford large police forces.
-recalculate_police_desire = {
-	add_to_variable = { expected_police_spending = police_desire_offset }
-	round_variable = police_desire_offset
-	clamp_variable = {
-		var = expected_police_spending
-		min = 1
-		max = 5
-	}
-}
-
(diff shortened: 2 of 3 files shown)


MrP0tter (approved):

SBeausoleil:
Amazing work! @AngriestBird

PR #3239, opened 2026-08-27
Add the E3 corporate history for the United States

JasonBreen:
### Changes
- Adds an E3 corporate history for the United States, thirteen events running from the industry's central marketplace in 2000 through the 2006 summit reform, the 2009 restoration, public ticketing in 2017, the 2020 collapse, the ReedPop revival, the show's retirement in 2023, and iicon in 2025
- Each option carries its own treasury, research and political cost and records its outcome as a country flag, so the preview tracks what actually happens
- Nine national ideas follow the show's changing character and swap as each era ends
- Milestones follow rule_corporate_history, firing the events under Full and recording the historical outcome under Outcomes Only


JasonBreen:
Closing to reopen with the chain shared across the countries whose industries actually shaped E3, rather than as a United States only history.

Issue #3246, opened 2026-08-27
Add engine-backed regression coverage for error.log fixes


AngriestBird:
## Bottom line

Static validation covers runtime OOB references, but the remaining error.log repairs depend on live game state and have no deterministic regression coverage.

## How

Add a scenario that asserts the relevant error.log messages do not recur for owned-state compliance/resistance, duplicate pact and leader-trait grants, railway path adjacency, occupation-law scope, invalid advisor-slot trait mutations, and third-party border-war event delivery.

Keep static validators as the fast gate. Use the scenario for the runtime transitions they cannot prove.

BLUF

AngriestBird:
This was handled as part in #3248

PR #3248, opened 2026-08-27
Fix three script errors from a 2005-2007 error.log

AngriestBird:
## Bottom line

Fixes the three scripted errors in this run. The AI command spam and occupation-law lookup have no script location and stay engine-side (#3246).

## Why

- `GCC_start_the_uprising` cancelled guarantees against a hardcoded GCC list, including the completer. SAU hitting it logged `diplomatic_relation: cant have a relation to yourself`.
- `create_unit` cannot parse umlauts. Sweden spawned `militärdistriktet` from Forsvarsbeslutet and the engine rejected the token.
- `sov_liberals.31` (Yeltsin's death) gated option b on a short ideology list. Putin is emerging Conservative, which is in that list, but any other government (oligarch, autocracy, etc.) got no valid option.

## How

Skip self in the GCC guarantee teardown. Rename the Swedish template to `militardistriktet`. Make option b the non-western fallback.

## Not fixed

- `unlock_trait_command` / `create_equipment_variant_command`: AI posted invalid commands, no script path.
- `GetOccupationLaw with no occupatin data`: engine, no script path.

BLUF


AngriestBird:
Pushed 1423888 — seven more validator findings from the 2015 run, beyond the three in the description.

- Yemen partition wired up: `on_monthly_SAU` now sets `yemen_partition_north` and the `HOU_north_yemen` / `YEM_south_yemen` cosmetic tags once both Yemens exist at peace. The flag and both tags were read-only before, so `yemen_partition_category` and its two drop-name decisions were unreachable.
- Defined `GFX_idea_economic_prosperity`; 14 SAU ideas rendered a blank icon.
- `antarctica_post_treaty_military_preparations`: `icon = army_reform` is 50x40 category art, switched to `army` (32x31).
- `05_iran.txt`: dropped the umlaut from the `Huseyncilar Briqadasi` template and its 12 `create_unit` references, same class as the Swedish one above.
- `SAU_Normalzation_with_zionist_20`: added the `can_staff_an_industrial_complex` guard.
- Renamed the event-picture `GFX_ukr_oun` to `GFX_ukr_oun_event`; it collided with the party icon of the same name and lost the load order, so the event art was dead.

Clean afterwards: cosmetic_tags, ideas, oob_units, gfx_references, variables, on_actions, common_mistakes.

BLUF

MrP0tter (approved):

PR #3253, opened 2026-08-28
Add balanced American and Chinese AI laboratory programmes

JasonBreen:
Late-game content, with both national sets shipped together so the American laboratories have a Chinese counterpart.

### Changes
- Adds eight American laboratory choices: four structural routes each for OpenAI and Anthropic.
- Adds twelve Chinese laboratory choices: two routes each for Baidu Ernie, Alibaba Qwen, Zhipu, ByteDance Doubao, DeepSeek and Moonshot.
- Treats each choice as a 365-day national programme rather than a permanent, stackable spirit.
- Allows only one AI laboratory programme per country at a time; resolving a company or laboratory still permanently closes its alternative routes.
- Gates every route on the repository's 2020 AI technology tier and an availability date matched to the represented laboratory or strategy.
- Requires sufficient treasury before paid routes can be selected and prevents the AI from choosing them while underfunded.
- Gives historical AI one historical route per laboratory and makes it wait until that route becomes available.
- Keeps the content independent of Corporate History because these are present and future policy programmes rather than historical event reconstruction.

### Notes for review
The Chinese decisions remain in the existing industrial investment category, which already opens on the semiconductor Big Fund. The American decisions remain in the existing economic category. No new categories, event pictures or permanent workflow files are added.

The one-year duration prevents six Chinese laboratory spirits or two American laboratory spirits from compounding permanently while preserving every authored route. Decision descriptions state both the duration and the one-at-a-time rule.

Availability is staggered from 2020 through October 28, 2025. `artificial_intelligence_8` is used consistently because it is the 2020 technology tier; the previous AI-9-and-later gates delayed nominally 2020-2023 content into the 2030s and beyond.

### CI note
The three red checks are infrastructure, not this branch. `coding-pipeline.yml` moved to `pull_request_target` in 2c546de9ee, and the pinned `actions/checkout` v7.0.1 refuses to check out fork pull request code under that event unless `allow-unsafe-pr-checkout: true` is set. Both jobs that check out the PR tree therefore fail before any validator runs, and `Pipeline` fails as the aggregate. This cannot be fixed from this branch, because `pull_request_target` loads the workflow from the base branch. The suggested two-line fix and the reasoning are in the comment thread.

The earlier OOB unit-name failure is resolved: #3260 is merged and is included here.


JasonBreen:
maybe not this way

JasonBreen:
i think the oob conflicts is bc of pr 3260

JasonBreen:
Reworked and validated in `dd5d126`:

- converted all 20 laboratory spirits into mutually exclusive 365-day programmes;
- corrected the technology/date gates and historical AI weighting;
- added treasury affordability checks and visible requirement tooltips;
- gave every timed lock flag `value = 1`, so the one-at-a-time gate works in engine;
- removed the Corporate History dependency and `allowed_civil_war` boilerplate.

Style, structural lint, localisation, idea references, technology categories, decision validation, common scripting checks, and variable validation are clean on the final tree. The remaining OOB unit-name failure reports 31 created-variant errors across 16 untouched files; that backlog is handled by #3260. This PR still changes only its original six content files.

JasonBreen:
The three red checks here are the workflow, not this branch.

`coding-pipeline.yml` moved to `pull_request_target` in 2c546de9ee. The pinned `actions/checkout` v7.0.1 refuses to check out fork PR code under `pull_request_target` unless `allow-unsafe-pr-checkout: true` is set, so both jobs that check out the PR tree die in about 3 seconds:

- `Prepare Workspace`, step "Checkout PR content" (`.github/workflows/coding-pipeline.yml:311`)
- `File Paths`, step "Checkout PR tree" (`.github/workflows/coding-pipeline.yml:861`)

Everything downstream skips, and `Pipeline` fails as the aggregate.

This only bites fork PRs, and #3253 is currently the only open one, so the change meant to get fork PRs validating against base-owned tooling is the thing blocking the only fork PR. I cannot fix it from here: `pull_request_target` loads the workflow from the base branch, so nothing pushed to this branch changes the run.

The fix is `allow-unsafe-pr-checkout: true` on those two steps only. The third checkout (`coding-pipeline.yml:960`) uses `github.repository` and needs nothing.

The opt-in looks safe as the job is already written. That guard exists to stop fork code executing in a trusted context, and this workflow never does that: the PR checkouts are data only with `tools/` and `.github/` outside the sparse list, `persist-credentials: false` is set on all of them, there is an explicit step rejecting symlinks and submodules in the PR index, and the validators execute from `../trusted/tools/` checked out from the base repo at `base-sha`.

On the content side this branch validates clean locally. `validate_localisation` and `validate_file_paths` report no issues, and `validate_decisions` returns no findings against any of the added decisions. The last run that actually reached the validators (33205540265) failed only on `Validate OOB Unit Names`, which #3260 fixed, and that is merged in here now.

AngriestBird (approved):

PR #3255, opened 2026-08-28
Tighten the two new common-mistake checks against the review findings

JasonBreen:
### Changes
- Restricts the `is_at_war` check to the boolean trigger form, so a variable or scripted-trigger definition of that name is no longer reported as an unknown engine trigger
- Reads modifiers inside single-line `ai_chance = { base = N modifier = { ... } }` blocks, which the AI fallback check previously skipped entirely
- Treats a positive `add` after a `factor = 0` as restoring the option, since AI weight operations apply in order and the option stays selectable
- Stops an option whose own trigger rules it out under bankruptcy from counting as a usable fallback
- Adds a regression test for each of the four

Follow-up to [#3251](https://github.com/MillenniumDawn/Millennium-Dawn/pull/3251), which merged before these fixes were pushed to the branch.

### DIFF
--- a/tools/linting/tests/check_common_mistakes_test.py
+++ b/tools/linting/tests/check_common_mistakes_test.py
@@ -52,8 +52,6 @@ import sys
 import tempfile
 
 from check_common_mistakes import (
-    classify_file_path,
-    _negates_bankruptcy_mission,
     _RE_IS_X_NATION,
     _ai_zero_modifier_conditions,
     _check_active_decision_defined,
@@ -95,8 +93,10 @@ from check_common_mistakes import (
     _equipment_names,
     _files_need_global_refs,
     _get_block,
+    _negates_bankruptcy_mission,
     _provincial_building_types,
     check_file,
+    classify_file_path,
 )
 
 passed = 0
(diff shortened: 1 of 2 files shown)


tools/linting/tests/check_common_mistakes_test.py
```diff
@@ -42,6 +42,7 @@
   39. add_opinion_modifier / add_relation_modifier naming no modifier
```
AngriestBird, line 42: While we are in here, did we add proper regression tests for the windows file pathing? If so merge otherwise add those.

AngriestBird (dismissed):

tools/linting/tests/check_common_mistakes_test.py
JasonBreen: afaik, no, starting on it

JasonBreen: wait wait, spoke too soon, pulled the classification out of check_file, and added three tests covering backslash for ideas, national focus and events

AngriestBird (approved):

PR #3256, opened 2026-08-28
Add the Xbox corporate history for the United States

JasonBreen:
### Changes
- Adds the Xbox platform as an eleven event chain running from the 2001 launch through Xbox Live, the 360, the red ring of death, Kinect, the Xbox One reversal, the games-first turn, Game Pass, the Series launch, and the cloud era
- Ends on one of five permanent outcomes: a consolidated platform, a hardware innovator, a developer platform, a cloud-native service, or a resilient games publisher
- Records the historical outcome by flag alone when the corporate history rule is set to outcomes only
- Adds the eight Xbox event pictures and their six textures, and the 2017 corporate history dispatch that did not exist yet

Every route through the chain reaches exactly one reachable outcome, and each event keeps an option the AI can still afford during bankruptcy.


JasonBreen:
this one should be the last event pr for now

PR #3284, opened 2026-08-29
Fix Autonomy Overlord Error Spam (#3263)

### LINKED ISSUE #3263: [BUG]Diplo/Autonomy error spam
The following in 50-60 error blocks every daily tick. Mostly the autonomous_states error with various countries
[01:22:40][2001.07.04.01][triggerbase.cpp:559]: common/autonomous_states/MD_chinese_autonomies.txt:295: overlord: invalid event target: OVERLORD
[01:22:40][2001.07.04.01][triggerbase.cpp:559]: common/scripted_triggers/00_diplomacy_triggers.txt:13: overlord: invalid event target: OVERLORD

TemplarGeneral:
### Changes
- Autonomy states no longer spam the error log every day on countries that have no overlord, removing roughly 55 error lines per daily tick
- Every OVERLORD check across the 23 autonomy files and the shared diplomacy trigger now stops early on non-subjects, so no autonomy availability or annexation condition changes

Someone needs to test this ingame

Closes #3263


common/autonomous_states/MD_AFG_autonomies.txt
```diff
@@ -32,6 +32,7 @@ autonomy_state = {
 	ai_overlord_wants_garrison = { always = yes }
 
 	allowed = {
```
AngriestBird, line 34: Did we actually test this? As far as I know this may break the actual autonomies if you pre check if they are a subject.

AngriestBird (changes requested):

common/autonomous_states/MD_AFG_autonomies.txt
TemplarGeneral: I can't find the arguments claude gave me but we checked vanilla and the wiki and we saw a reference there so his argument to me looked valid. So thats why i coded it in and it needs to be tested. If that isn't the case then i honestly don't know. It doesn't looked like the engine renders the overlord = x without the check if the country is a subject atleast.

AngriestBird:
Need to look into this. Taking this task.

Issue #3162, opened 2026-08-23
[TASK] Standardise country AI path game rules: Historical / alt-history / Random Path / No Path


TemplarGeneral:
## Problem

`common/game_rules/00_game_rules.txt` defines 57 country AI path rules (`TAG_ai_behavior`, group `RULE_GROUP_AI_BEHAVIOR`, lines 1695–3641). They share no standard:

- Most name their historical route `DEFAULT` ("Default"), whose description talks about "other gameplay settings" rather than about the country. A player cannot tell which option is the historical one — several rules name it after a party or a president instead.
- A few rules carry both `DEFAULT` **and** `HISTORICAL`, so there are two candidates for the same thing.
- 8 rules have no random option; Switzerland has only random and no historical route at all.
- No rule lets the player turn scripted paths off entirely.
- Descriptions run from one clause to five sentences.
- The rules are listed in TAG order, so the in-game list reads Algeria, Armenia, Bahrain, Belarus… but then Netherlands after Greece, and Yugoslavia in the middle.

## The standard

Every `TAG_ai_behavior` rule gets exactly this option set, in this order:

| # | `name =` | Player-facing text | Meaning |
|---|---|---|---|
| 1 | `HISTORICAL` | *country-specific historical title* | The `default = { }` block. What actually happened. Replaces today's `DEFAULT`. |
| 2..n | *one per alt-history path* | *country-specific title* | One option per scripted alternate route the country's tree supports. |
| n+1 | `RANDOM_PATH` | `Random Path` | Picks one of the options above — historical **or** alt-history — at random on startup. |
| n+2 | `NO_PATH` | `No Path` | No scripted path flag is set. The AI picks focuses freely. |

`DEFAULT` is removed everywhere. Where a rule currently has both `DEFAULT` and `HISTORICAL`, they merge into the single `HISTORICAL` default block.

### Naming rules

- The historical option's text names the real thing that happened, and the desc's first sentence makes it unambiguous that this is history (e.g. Algeria's `"The Bouteflika Years"`). Never `"Default"`.
- Alt-history option text describes the outcome in the player's language, not internal jargon. `NKO_un_rules` / `NKO_nam_rules` / `NKO_junta_rules` and `ITA`'s `LEU` / `NONCOM` / `LEFTP` / `RIGHTP` are the worst offenders.
[...]

PR #3307, opened 2026-08-30
Harden Bahrain AI paths (issue #3162)

TemplarGeneral:
Part of #3162. Converts Bahrain to the standard AI-path rule set and fixes the reasons the rule did nothing.

## The rule

`BHR_ai_behavior` had `DEFAULT` / `DAY_OF_RAGE` / `RANDOM`. `DEFAULT` set no flag, `RANDOM` was a bare `random = { chance = 50 }` that drew only the alt path, and there was no way to turn scripted paths off.

Now `HISTORICAL` (the `default` block) / `DAY_OF_RAGE` / `CROWNED_DEMOCRACY` / `RANDOM_PATH` / `NO_PATH`. `RANDOM_PATH` is a `random_list` summing to 100 that includes the historical bucket; `NO_PATH` deliberately sets no flag. Loc moved off the unprefixed `RULE_OPTION_DAY_OF_RAGE` / `_AI_DESC` onto `RULE_OPTION_BHR_*`, with `RANDOM_PATH` / `NO_PATH` reusing the shared `RULE_OPTION_MD_*` keys.

The third path is derived from the prerequisite graph, not from the option names. The real fork is the four-way mutex at `GCC_free_and_fair_elections`, and the tree supports a crowned democracy (western dissidents into `GCC_westernization_policies` into `GCC_preserve_the_royal_family`) that no option claimed.

## Why the rule was dead

**The flag reached zero focus weights.** Across `gulf.txt` and `gulf_shared.txt` there were **0** `ai_will_do` modifiers referencing any `*_FOCUS_PATH` flag — the only consumers were the strategy plan and two immigration decisions. The only path switch in the tree was the global `is_historical_focus_on`, which cannot tell a default Bahrain from a Day-of-Rage Bahrain.

**The AI lost every relevant fork.** `GCC_empower_shia_democrats` and `GCC_empower_western_based_dissidents` were `base = 0` +1 only when `is_historical_focus_on = no`, while sibling `GCC_empower_sunni_democrats` was a flat `base = 1`. Under historical AI — the default — both alt branches lost 1-0.

**The deep end of the path was unreachable.** `GCC_shia_dominance`, `GCC_favor_khomeinism`, `GCC_favor_the_shiraziyyin`, `GCC_non_aggression_pact` and `GCC_align_with_iran` all require `has_government = communism`; `GCC_establish_a_republic` requires `NOT western_autocrats_are_in_power`. Bahrain starts `ruling_party = democratic` (slot 0 `Western_Autocracy`) with `elections_allowed = no` **and `set_partyall_banned = yes`**. `calculate_banned_parties` zeroes `party_pop_elect_array` for every banned party, so even after `GCC_free_and_fair_elections` enables elections the incumbent is the only electable party and wins 100%. Unbanning exists only in the politics-view scripted GUI, which the AI cannot use.

## Reachability fix

`BHR_legalise_the_opposition`, an `is_ai = yes` decision added to the existing `shia_decisions` category, clears the ban flags its path needs — `party7`/`party9` (Conservative, Mod. Vilayat-e Faqih) on Day of Rage, `party2`/`party3` on Crowned Democracy. That is exactly what a human does through the politics view, so it uses the mechanic the country already has rather than force-flipping the ruling party.

The election then resolves on real popularity: the communism bloc takes **49.4%** and the government flips, opening the rest of the spine. No drift ramp or drift-cancelling modifier is needed — Bahrain already carries roughly +0.28 communism drift (`home_of_the_baharna`, `unrepresentative_government_sunni`, `large_shia_population`, `shia_resistance_04`, `shia_conservative_society`) against the 0.25 `drift_defence_factor` of `rentier_state`.

## Blast radius

`gulf.txt` is **one** `focus_tree` serving QAT, UAE, BHR, KUW and OMA, with no `allow_branch` anywhere — every Gulf tag sees every branch. Every modifier added therefore carries `original_tag = BHR` alongside the path flag, asserted programmatically. The other four GCC tags are provably unaffected.

Because the GCC political focuses are `base = 0` with additive modifiers, a `factor` path flag would multiply zero and do nothing: path selection is an `add`, and the killswitch is a `factor = 0` placed last. Boosts sit ahead of pre-existing `factor = 0` guards so they cannot bypass the bankruptcy/staffing brake on `GCC_the_dubai_model`.

The four-way election fork now resolves to exactly one live branch per path. Three mutex forks that sat at identical `base = 1` with no modifiers on either side (`shia_dominance`/`fair_representation`, `favor_khomeinism`/`favor_the_shiraziyyin`, `deepen`/`shrink_the_safety_net`) got Bahrain-guarded tie-breaks.

## Things that were losing the AI the game

- **`GCC_a_radical_reorientation`** is the jihadist-branch entrance leading to `GCC_start_the_uprising` — the only `start_civil_war` in the file, which also transfers five states to `SAG` and moves the capital. Bahrain passes its `available` (it has the `sunni` idea). Killswitched on all three paths and zeroed in all four plans.
- **`GCC_the_war_within`** is `will_lead_to_war_with` all five GCC states and mass-creates `puppet_wargoal_focus` wargoals; its weight fired for **every** Gulf AI after 2021 even in historical games. Killswitched for Bahrain on all three paths.
- **`GCC_expel_us_interests`** runs `transfer_state = 624` (a Saudi state) at a flat `base = 1` with no guard. Now Day of Rage only.
- **`gulf.17`** had no `ai_chance` on either option while option `.a` fires `create_wargoal = { type = topple_government target = SYR }` — at the default weight of 1 each, a coin flip into a war on Syria for any Gulf AI that took `GCC_intervene_in_syria`. Stand-down now carries `base = 100`.

## Strategy plans

`BHR_DAY_OF_RAGE` was the only plan, had no `focus_factors`, and its list mixed mutually exclusive branches. On `DEFAULT` and half of `RANDOM`, Bahrain ran with **no enabled strategy plan at all** — there is no shared Gulf plan to fall back on.

Now four plans with `focus_factors`, including `BHR_no_scripted_path` enabled when no path flag is set, so `NO_PATH` never leaves Bahrain plan-less. Self-conflicts in the old list fixed: `GCC_keep_to_the_saudi_line` permanently locked out `GCC_strategic_hedging`, `GCC_expel_us_interests` was listed *after* the focus that requires it, `GCC_live_and_let_live` needed an absent prerequisite, `GCC_ideological_partners` required not being a GCC member, and `GCC_jordanian_ties` required `NOT communism` beside communist-gated entries.

## New file

`common/ai_strategy/BHR.txt` — Bahrain had no strategy file at all, so no losing-war brake despite `GCC_the_war_within`. `BHR_cancel_war_gcc` follows the `ALG.txt` pattern on `surrender_progress > 0.15`, not the inverted `strength_ratio ratio > 1` form.

## Validation

- `validate_focus_tree`, `validate_events`, `validate_on_actions`, `validate_localisation`, `validate_simplifications` — clean
- `check_common_mistakes` — passed, 6399 files
- `validate_decisions` — 455 warnings in 69 files, **identical to a stashed baseline** (all pre-existing `decision-icon-slot-mismatch`)
- `validate_variables` — 1 warning, pre-existing and unrelated (`unlocalised-available-flag` in `common/decisions/05_CHI_decisions.txt`)
- Scripted assertions: every path-flag modifier is `original_tag = BHR` guarded; the election fork resolves to one branch per path; no path boost sits after a `factor`; no focus listed in a plan is stranded by the killswitches
- LF endings, and BOM only on the two `.yml` files, verified across all nine files
- `pre-commit` is not installed in this environment; CI will run it

Non-English localisation untouched, per `AGENTS.md`.

## Deferred to the cross-cutting pass

Alphabetical reordering of the rule blocks and the loc file, and the `@TAG <short name>` header sweep — `BHR_AI_BEHAVIOR: "@BHR Bahrain"` already conforms.

## Test plan

1. Start as observer with each of the five rule options; confirm the expected `BHR_*_FOCUS_PATH` global flag, and none under `NO_PATH`.
2. Confirm Bahrain has an enabled strategy plan under **every** option, including `NO_PATH`.
3. Under `DAY_OF_RAGE`: the AI walks the charter into parliament, censorship and free elections, takes `BHR_legalise_the_opposition`, wins the next election into the `communism` group, then reaches `GCC_shia_dominance`, `GCC_establish_a_republic` and `GCC_align_with_iran`.
4. Under `CROWNED_DEMOCRACY`: the AI runs the economic spine into `GCC_the_dubai_model` and `GCC_weaken_the_mutaween`, takes western dissidents into `GCC_westernization_policies` and `GCC_preserve_the_royal_family`, and stays Western-aligned.
5. Under `HISTORICAL`: the AI stops short of free elections and runs the crackdown line (`GCC_rally_sunnis`, `GCC_keep_to_the_saudi_line`, `GCC_strengthen_us_ties`).
6. Confirm no Bahrain AI takes `GCC_a_radical_reorientation` or `GCC_the_war_within` on any path.
7. Spot-check that Qatar, the UAE, Kuwait and Oman pick the same focuses as on `main`.


common/decisions/Gulf.txt
```diff
@@ -841,6 +841,56 @@ gulf_society = {
 
 shia_decisions = {
 
+	# The politics view is the only route out of set_partyall_banned, and the AI cannot use it.
+	BHR_legalise_the_opposition = {
+
+		icon = GFX_decision_generic_political_discourse
+
+		visible = {
+			original_tag = BHR
```
AngriestBird: Moved to allowed

AngriestBird (changes requested):

common/decisions/Gulf.txt
TemplarGeneral: Oversight! gimme a sec

TemplarGeneral: Done in d09025e — `original_tag = BHR` moved out of `visible` into a decision-level `allowed` block.

AngriestBird (approved):

PR #3343, opened 2026-08-31
Add an Ericsson military industrial organization for Sweden

JasonBreen:
### Changes
- Adds `SWE_ericsson_manufacturer`, an eleven-trait MIO in two branches, to Sweden's organisation file.
- Adds the 27 English localisation keys it needs.

### Why Ericsson
Sweden ships eight MIOs and none of them is Ericsson, even though the Ericsson corporate history is already upstream in `events/Sweden.txt`. The icon `GFX_idea_SWE_ericsson` already existed and was otherwise unused.

### Built as a research organisation, not an equipment manufacturer
`equipment_type` is empty and every trait moves organisation modifiers and tech bonuses, following `TAI_tsmc_manufacturer` and `GER_siemens_defence_electronics` rather than the platform-maker pattern.

That is deliberate. Ericsson does not build a platform the MIO system can attach to, and per `.claude/docs/mio-reference.md` an equipment bonus naming a stat the target archetype never declares is inert, which `validate_mios` reports as `mio-bonus-no-base-stat` at ERROR severity. Keeping the org on organisation modifiers avoids writing bonuses that look meaningful and do nothing.

Research categories are `CAT_internet_tech`, `CAT_computing_tech` and `CAT_microchips`. The first two are what the Ericsson chain itself already grants, so the MIO and the events agree on what the company is good at.

### Shape
Eleven traits over 359 lines, against a repo median of 411 across the 485 existing MIOs. Networks and Deployment runs base stations through the mobile core to hardened military bearers; Research and Standards runs the laboratory, in-house baseband silicon and the standards work that is the real source of the company's leverage. Both branches converge on an integrated network house. `task_capacity = 2`, matching the doc's guidance for a niche organisation.

### Note on the rest
This is one of several companies from the merged corporate histories that could take a MIO, but it is currently the only one that can be built: STMicroelectronics, Alcatel and Nokia have no icon sprite anywhere in the repo, and the generic `GFX_idea_company_*` set is named for specific real firms, so borrowing one would be wrong. Those need art before they can follow.

### Checks
`validate_mios`, `validate_style` and `validate_localisation` all clean, `check_common_mistakes` passes, brace depth balanced, and every `token`/`name`/`text` reference in the new block resolves to a localisation key.


common/military_industrial_organization/organizations/MD_SWE_organizations.txt
```diff
@@ -7025,3 +7025,363 @@ SWE_swede_ship_marine_manufacturer = {
...
+			SWE = {
+				add_tech_bonus = {
+					name = SWE_ericsson_radio_base_stations
+					bonus = 0.10
+					uses = 1
+					category = CAT_internet_tech
+				}
+			}
+			expenditure_for_mio_upgrade = yes
+		}
```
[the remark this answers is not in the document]
JasonBreen: Good catch, fixed in c570ce9. All seven on_complete blocks now scope via `FROM`.

I had copied the hardcoded tag from `TAI_tsmc_manufacturer`, which turns out to be the outlier rather than the pattern: across the organisation files 95 on_complete tech bonuses use `FROM` and only five did not, three of those being TSMC itself. The expenditure call is also moved above the bonus block to match the same convention.

common/military_industrial_organization/organizations/MD_SWE_organizations.txt
```diff
@@ -7025,3 +7025,363 @@ SWE_swede_ship_marine_manufacturer = {
...
 	}
 }
+
+SWE_ericsson_manufacturer = {
+	allowed = { original_tag = SWE }
+	icon = GFX_idea_SWE_ericsson
+
+	task_capacity = 2
+
+	equipment_type = {
```
TemplarGeneral: This block can be removed if its empty

TemplarGeneral (changes requested):

common/military_industrial_organization/organizations/MD_SWE_organizations.txt
JasonBreen: Removed. Checked it against the rest of the folder first: 182 of the 486 organizations omit `equipment_type` entirely, so it is optional for a research-only org. The only other empty one is `TAI_tsmc_manufacturer`, which is where I copied the block from, so that one is worth the same treatment separately.

TemplarGeneral (approved):

PR #3355, opened 2026-09-01
Wire up and standardise the Ethiopia AI path game rule (#3162)

TemplarGeneral:
Part of #3162. Brings `ETH_ai_behavior` to the standard and wires the tree, which previously read the path flags in 3 of 93 focuses.

## Rule and localisation

`HISTORICAL` → `MONARCHIST_RESTORATION` → `DERG_RESTORATION` → `RANDOM_PATH`, with `NO_PATH` as the `default = { }` block. Option names lose their `ETH_` prefix, `RANDOM` becomes `RANDOM_PATH` on the shared `RULE_OPTION_MD_*` keys, and the historical option's text is now literally `Historical`. All three `_desc`s rewritten to exactly two sentences, present tense about Ethiopia, `§8…§!` on party names.

## Flag wiring

Flags renamed to the `TAG_<PATH>_FOCUS_PATH` standard, plus a new `ETH_HISTORICAL_FOCUS_PATH` — `HISTORICAL` previously set no flag at all. The `RANDOM_PATH` `random_list` now draws a real historical bucket instead of the empty `34 = { }` placeholder. `NO_PATH` sets nothing.

New `common/scripted_triggers/99_ETH_scripted_triggers.txt` carries the ownership triggers.

## Focus weights

22 focuses across the three political spines (5 republican / 10 monarchist / 7 Derg) get the standard multiplicative boost + killswitch pair via `apply_ai_path_weights.py`, replacing the ad-hoc `factor = 150` / `factor = 10` modifiers on the three spine roots. Applied by tool, not by hand.

Left path-neutral on purpose: `ETH_endorse_the_african_union` (owning it to historical would strand `ETH_the_new_african_union_headquarters`), the `reckless_industrial_expansion` ⟷ `secure_agricultural_production` pair (cross-spine `OR` prerequisites), and the whole economic / diplomatic / military / Eritrea trunk.

## AI hardening

**Monarchist reachability.** `ETH_restore_the_throne` needs `nationalist > 0.35`. Party 23 starts at `0.07` and the tree's only boosts are `legacy_of_abyssinia` (+0.15), `endorse_the_monarchists` (+0.10) and `glorify_the_italo_ethiopian_wars` (+0.10) — 0.42 only if all three land. `glorify_the_italo_ethiopian_wars` and `ETH_italian_ethiopian_friendship` mutually exclude each other via reciprocal `available` blocks, and friendship sits in the diplomacy spine at `base = 1` with no path awareness, so an AI that takes it first caps nationalist at 0.32 and the entire Crown path dies behind an unsatisfiable gate. Fixed with the standard AI-only popularity ramp (`ETH_ai_path_category` + `ETH_rally_the_monarchists`, no loc keys) rather than by killswitching friendship, which would orphan `ETH_befriend_france`.

**Derg military gap.** `ETH_modern_officer_corps` required `ETH_guarantee_freedom_of_religion` (republican) or `ETH_imperial_nostalgia` (monarchist), amputating it from the Derg path. `ETH_restore_the_pdre` added as a third `OR`.

**Eritrea war.** `ETH_reclaim_eritrea` creates an `annex_everything` wargoal at `base = 1`, and `ETH_eritrea_border_conflicts` pins `conquer ERI value 100`. `common/ai_strategy/ETH.txt` ended on a dangling comment for a brake that was never written; it is replaced with an `ETH_avoid_unready_war_with_eri` / `ETH_prepare_war_with_eri` readiness pair modelled on the BOS blocks, gated on `strength_ratio = { tag = ERI ratio > 1.5 }`.

## Verification

- `ai_path_report.py --tag ETH` — rule/wiring **clean**, 22 owned focuses, **0 orphans and 0 stranded in all 10 state rows** (5 options × historical AI on/off), no additive path modifiers, no unreferenced flags, no surviving `has_game_rule` reader.
- `validate_focus_tree.py`, `validate_localisation.py`, `validate_variables.py`, `validate_style.py`, `validate_simplifications.py`, `validate_unused_scripted.py`, `validate_common_mistakes.py` — all clean.
- `validate_decisions.py` — 283 errors / 458 warnings / 69 files, identical to the pre-change baseline; no new `ai-only-decision-localisation`.

`NO_PATH` verified working: every focus keeps `base = 1`, so with no flag set and historical AI off the AI chooses freely; with historical AI on, `ETH_ai_historical_path` still steers it to the republican spine.

## Out of scope, pre-existing

`ETH_eritrea_dissolve_assembly` and `ETH_eritrea_federation` each require in `available` the very country flag their own `completion_reward` sets, and nothing else in the repo sets either — the whole 14-focus Eritrea integration spine is unreachable. Unrelated to AI paths; not touched here.


common/scripted_triggers/99_ETH_scripted_triggers.txt
```diff
@@ -0,0 +1,38 @@
...
+		is_historical_focus_on = yes
+		has_global_flag = ETH_HISTORICAL_FOCUS_PATH
+		has_global_flag = ETH_MONARCHIST_FOCUS_PATH
+	}
+}
+
+ETH_ai_can_attack_eri_trigger = {
+	country_exists = ERI
+	has_wargoal_against = ERI
+	NOT = { is_in_faction_with = ERI }
```
Blazer135, line 36: [P2] Account for Eritrea's defensive bloc before declaring war

This gate compares Ethiopia only with Eritrea. NOT is_in_faction_with = ERI excludes a shared faction, but it still passes when Eritrea belongs to another faction or is a subject, so ETH_prepare_war_with_eri adds +200 declare_war without considering the forces of an overlord or allies. Please add target-open checks comparable to the BOS/DEN gates (at minimum no faction and no overlord; ideally guarantees as well), or include those defensive powers in the readiness test.

Blazer135 (approved):

PR #3358, opened 2026-09-01
Standardise the France AI path game rule (#3162)

TemplarGeneral:
Part of #3162. Brings `FRA_ai_behavior` onto the standard and makes every path it offers actually reachable.

## The two real bugs

**Six paths were unselectable.** `RANDOM_WESTERN` set an umbrella flag that a second `if` then resolved into one of six party flags at random, and `RANDOM_NATIONALIST` rolled National Rally / Kingdom / Empire. Each of those flags already had its own strategy plan and its own 22-28 focus spine, but the player could not pick any of them — `FRA_RN_FOCUS_PATH` in particular was unreachable by direct choice.

**Choosing a regime change usually did not produce that regime.** Both decisive forks carried a bare `ai_will_do = { base = N }` with no path weighting:

| fork | republic | alternative |
| --- | --- | --- |
| `FRA_question_republic_fn` | `FRA_republic_safe_fn` base 5 | `FRA_restore_kingdom_fn` / `FRA_restore_empire_fn` base 2 each |
| `FRA_constitutional_crisis_pcf` | `FRA_republic_safe_pcf` base 5 | `FRA_french_commune_peaceful` base 2 |

So "The Kingdom of France" and "The French Commune" more often than not left France a republic.

## Rule and localisation

Options are now `HISTORICAL` + ten alt-history paths + `RANDOM_PATH` + `NO_PATH`, with `NO_PATH` as the `default = { }` block. `DEFAULT`, `RANDOM_WESTERN`, `RANDOM_NATIONALIST` and `RANDOM` are gone.

Historical · The Socialists Return · The Liberal Centre · The Gaullist Revival · The Green Republic · France Unbowed · The National Rally · The French Commune · The Kingdom of France · The French Empire · Random Path · No Path

`REPUBLICAN` is deliberately **not** an option: its plan's `focus_factors` were byte-identical to Historical's, both open on `FRA_LR`, and the LR plan already needed a `NOT` guard against the historical flag to stop both firing. The plan stays in the file, reached through `has_completed_focus = FRA_LR`.

The historical option's text is now literally `"Historical"`; the Chirac/Sarkozy/Hollande/Macron trajectory moved into its desc, without years. Every desc is exactly two sentences with `§8…§!` on party names. `Random Path` / `No Path` reuse the shared `RULE_OPTION_MD_*` keys.

## Wiring

One `if` per option. Flags renamed to the standard `TAG_<PATH>_FOCUS_PATH`; the umbrella `FRA_WESTERN_FOCUS_PATH` and the now-dead `FRA_AI_LR_FOCUS_PATH` are removed. `common/ai_strategy/FRA.txt` and `events/France.txt` read the new `FRA_ai_atlanticist_path` scripted trigger instead of the umbrella — it was what flattened Socialist, Liberal, Republican and Green France into one identical foreign policy at runtime.

`RANDOM_PATH` draws from an 11-bucket list that includes the historical bucket.

## Focus weights

All 264 political-spine focuses are now owned by their path, not just the 11 spine roots. Previously a killswitched root stranded its whole subtree: orphans went from 62 in the baseline to 263 once the roots were killswitched. They now sit at the 40 pre-existing Francosphere/Levant cases in **every** state, historical AI on or off.

Two further defects fell out of that pass:

- Six party roots carried `is_historical_focus_on` `add` modifiers that exactly cancelled their own base. Under historical AI the Green, France Unbowed and Commune roots were dead **even when the player had explicitly picked them**. The path system expresses this steering correctly, so the legacy pair is removed.
- `FRA_ai_historical_path` defers to an explicit choice: with historical AI on, picking Commune now zeroes the historical spine instead of leaving both alive at equal weight. Every option now behaves identically whether historical AI is on or off; `No Path` still follows history under historical AI, and is fully free without it.

The two Commune mutex ties (`FRA_deal_devil_pcf`/`FRA_left_front`, `FRA_eu_communism_PCF`/`FRA_tie_moscow`) are broken toward the focuses that actually advance the path — one cuts democratic drift, the other adds PCF popularity.

## Reachability

Seven paths could not get their party elected: against a net starting drift of democratic **+0.18** and `drift_defence_factor 0.40`, UDF starts at 0.06, RPF 0.11, EELV 0.03, LFI 0.00, FN 0.12, PCF 0.07, and `common/decisions/France.txt` had zero `is_ai` decisions. Adds an `is_ai` `FRA_ai_path_category` with a rally/takeover pair per target party index. Historical needs none (`france_md.550-553` force it) and Socialist needs none (PS starts at 0.32, the plurality leader). Kingdom and Empire reuse the FN ramp, since both enter through the National Rally holding power first.

No localisation keys are added for the category or the twelve decisions, per the AI-only rule.

## AI hardening

- `FRA_not_done`, `FRA_turn_ecology`, `FRA_party_unity`, `FRA_stole_voter` and `FRA_cut_with_unions` sat at `base = 800` against a tree-wide range of 1-50. Once a No Path AI opened the Socialist spine it was railroaded down that sub-branch, defeating the point of No Path. Normalised to 10.
- `ai_is_threatened` weighting added to the 23 combat-capacity focuses; the tree had none.
- `FRA_RN_plan` zeroed `FRA_question_republic_fn`, the sole prerequisite of `FRA_republic_safe_fn`, dead-ending the National Rally path at `FRA_focus_insecurity`.
- No losing-war brake was added: `MD_avoid_new_wars_when_outmatched` already brakes every country at `enemies_strength_ratio > 0.75`, so a stricter per-tag block would be a strict subset and dead code.

## Tooling fix

`tools/standardization/apply_ai_path_weights.py` discarded the `base` of a single-line `ai_will_do`, silently rewriting `base = 80` to `base = 1` — it hit seven France focuses on the first run. Fixed, with a regression test; it now also refuses an inline block with a nested modifier rather than dropping it. This affects every remaining #3162 country.

## Validation

- `ai_path_report.py --tag FRA` — rule/wiring clean, 0 path-relevant mutex ties, no additive path modifiers, no unreferenced flags, orphans flat at the pre-existing 39-40 across all 24 states.
- `validate_focus_tree.py --path .` — clean repo-wide, unchanged from baseline.
- `validate_decisions.py` — 283 errors / 458 warnings in 69 files, byte-identical to baseline; the twelve new decisions add nothing.
- `check_common_mistakes.py` — passed.
- `pytest tools/tests/standardization/` — green.

## Noted, not fixed (out of scope)

- `common/ai_strategy/FRA.txt:488` `FRA_help_georgia` has `allowed = { original_tag = SOV }` despite the name and its FRA-volunteer trigger, so it never runs for France.
- `france_md.374` grants +0.10 to party index 21 (`Nat_Fascism`), which no French party maps to — royalists and Bonapartists both sit at index 23.


common/decisions/France.txt
```diff
@@ -7989,3 +7989,318 @@ FRA_GREEN_EU_assistance_decision_category = {
...
+
+		cost = 25
+
+		days_re_enable = 30
+
+		visible = { has_global_flag = FRA_FRANCE_UNBOWED_FOCUS_PATH }
+
+		available = {
+			has_civil_war = no
+			has_country_flag = FRA_LFI_CREATED
```
Blazer135: The France Unbowed path cannot start because this decision requires `FRA_LFI_CREATED`, but that flag is only created in the Socialist branch.

Please remove `has_country_flag = FRA_LFI_CREATED` from this decision's `available` block and add `set_country_flag = FRA_LFI_CREATED` to its `complete_effect`. The first rally decision will then create LFI and unlock the `FRA_LFI` focus.

events/France.txt
```diff
@@ -11421,10 +11418,7 @@ country_event = {
 			}
 			modifier = {
 				factor = 0
-				OR = {
-					has_global_flag = FRA_WESTERN_FOCUS_PATH
-					has_global_flag = FRA_HISTORICAL_FOCUS_PATH
-				}
+				FRA_ai_atlanticist_path = yes
```
Blazer135, line 11421: The National Rally path can randomly allow the traditionalists and enter the Kingdom/Empire coup chain. The restoration focuses are then disabled because the selected path is still National Rally, leaving the AI stuck.

Please add this modifier to the "Allow the traditionalist faction to organize" option in event 371:

```txt
modifier = {
	factor = 0
	has_global_flag = FRA_NATIONAL_RALLY_FOCUS_PATH
}
```

This makes National Rally always choose the republican option and prevents it from entering the incompatible restoration branch.

common/decisions/France.txt
TemplarGeneral: Fixed in dab133b065 exactly as suggested — `has_country_flag = FRA_LFI_CREATED` is out of `available`, and `set_country_flag = FRA_LFI_CREATED` is now the first effect after the log line in `complete_effect`. The first rally founds LFI and unlocks the `FRA_LFI` focus.

For the record on the trace: the flag was only ever set by `france_md.520.a`, which is fired by `france_md.162`, which is fired only by the focus `FRA_cut_with_unions` (`factor = 0 FRA_ai_not_socialist_path = yes`) — so France Unbowed could never reach it. I checked the sibling flags too: `FRA_LREM_CREATED` and `FRA_RN_CREATED` only ever appear inside `NOT = { ... }` blockers, so no other path has the same problem.

events/France.txt
TemplarGeneral: Fixed in dab133b065, but I put the guard on `france_md.375` option `.b` rather than on 371 `.a` — flagging the difference in case you disagree.

`FRA_question_republic_fn` is `allow_branch`-gated on `FRA_allowed_extremists_right`, so blocking 371 `.a` would keep National Rally out of the End-the-Republic branch entirely. That branch is not only the restoration route: `FRA_question_republic_fn` is weighted `FRA_ai_nationalist_path = yes` (which includes National Rally) and its crackdown ending `FRA_republic_safe_fn` is weighted `factor = 25 has_global_flag = FRA_NATIONAL_RALLY_FOCUS_PATH` with `factor = 0` for every other path. Gating 371 would make that focus AI-unreachable on all paths.

So instead:

```txt
# france_md.375 option .b - "Let them be, the restoration continues"
modifier = {
	factor = 0
	has_global_flag = FRA_NATIONAL_RALLY_FOCUS_PATH
}
```

National Rally now always takes 375 `.a`, gets `FRA_crackdown_extremists_right`, and completes `FRA_republic_safe_fn` — deterministic the same way Kingdom and Empire already are through 371/375/374. The restoration chain is untouched for Kingdom and Empire, and the soft-lock is gone. Happy to switch to the 371 version if you would rather cut the branch from RN altogether.

Blazer135 (approved):

PR #3363, opened 2026-09-01
Convert ruling_party array to a scalar variable

AngriestBird:
## Bottom line
A country has exactly one ruling party, so the single-element `ruling_party` array is now a plain scalar variable. This removes ~3,000 per-check array scans from code the engine evaluates constantly (political scripted triggers, subparty policy ideas, decision visibility, focus availability).

## How
- `add_to_array` -> `set_variable` (384 history files plus every effect site); `clear_array` -> `clear_variable`, or deleted where an unconditional set follows
- `is_in_array` -> `check_variable` (~2,900 sites). Index 0 (Western_Autocracy) routes through `western_autocrats_are_in_power`, which now carries a `has_variable` guard: an unset variable reads 0, and 9 countries plus fresh civil-war tags never set it, so a bare 0-check would have marked them autocracies
- `ruling_party^0` -> `ruling_party`; `ruling_party^num` size checks -> `has_variable`; `for_each_loop` over the array -> direct `party_pop_array^ruling_party` indexing behind a `has_variable` guard
- `set_party_index_to_ruling_party` keeps its technocrat-under-autocracy branch (picks the biggest coalition partner)
- Fixed a latent Bulgaria bug: three coalition-invite focuses appended a second party to `ruling_party` instead of `gov_coalition_array`; they now target the coalition array, matching their tooltips
- `gov_coalition_array`, vanilla `set_politics` fields, and localisation are untouched (`[?ruling_party|0]` renders a scalar correctly)

Verified: zero remaining array-form references repo-wide, all 24 `*_are_in_power` trigger indices unchanged, clear-deletion sites audited for stale-value paths, pre-commit content validators pass.

BLUF


AngriestBird:
Second commit addresses a deep audit of the conversion:

- Kept the old slot-0 write in `hold_elections` / `recalculate_party_strength` — the original loop's undeclared iterator was the index (always 0), so converting it to `^ruling_party` would have silently changed sham-election math. The latent bug is filed as #3369 for a separate balance call.
- Bulgaria's green/libertarian branches: the old append bug made the invited party count as "in power", which is what unlocked the partner branch. Those gates now use the existing `*_in_power_or_coalition` triggers, so the invite focuses still open them (a party in a regular coalition now also qualifies — reviewer call if that's unwanted).
- Added `has_variable = ruling_party` guards to `is_party_in_government`, the politics-view attack/ban button checks, and the largest-opposition scans: an unset variable reads 0, and three tags (PAT, SIL, KSH) allocate party arrays without ever setting a ruling party.

BLUF

MrP0tter (approved):

PR #3390, opened 2026-09-02
Add the Samsung and ASML corporate histories

JasonBreen:
### Changes

- Adds the **Samsung** (KOR) and **ASML** (HOL) corporate histories, 15 milestones each — the two tags the series has nowhere, fork or upstream
- Adds a **semiconductor market layer** that couples them: ASML books a tool order, Samsung takes delivery, and a quarterly global demand curve moves both. Three clamped globals (demand, tool backlog, access tension), driven once per quarter behind a global flag so the two country hooks cannot double-run it
- Wires 29 `*_milestone_<year>` effects into the yearly pulse, `full` firing the event and `outcomes_only` recording the historical result
- Ideas go into `Korea.txt` and `05_netherlands.txt` next to the existing ones with no `allowed` block, matching the TSMC and Arm chains

### Scheduling was rebuilt, not ported

These were prototyped against a per-country monthly sweep: KOR re-tested a 15-branch date window every month, HOL ran a date-gated catch-up pass, and both carried `_pending` / `_delivery_expected` bookkeeping to stop events firing twice. That is replaced by one milestone effect per year off `00_yearly_effects.txt`, on the `ENG_arm_milestone_*` shape. The per-event flag bookkeeping and both hidden reconstruct events go with it.

### Verification

- **Rule modes** — all three resolve. KOR's capstone previously resolved only under `full`; `outcomes_only` now awards it too
- **Flag reachability**, both directions, whole-repo: zero set-but-never-read, zero read-but-never-set. Stripping the fork's dashboard and reconstruct model stranded 22 setters and left two reads (`HOL_asml_tool_capacity_active`, `HOL_asml_global_field_service_active`) that nothing could ever set — the setters are gone and that condition collapses to the reachable maturity/resilience pair
- **Localisation** — 94 keys whose only consumers were the dashboard decisions and their scripted localisation were dropped
- Every `picture` resolves; all 29 milestones are both defined and called; ideas, events and common-mistakes validators are clean

### One thing to look at

HOL's events gate sequentially — event N requires `HOL_asml_event_<N-1>_resolved` — so a beat missed while collapsed stalls the remainder. The monthly recovery sweep used to paper over this. I kept the chain's own design rather than rewriting it here; KOR has no such coupling, its events gate on their own outcome flags. Happy to flatten HOL's gating if you would rather it match.


common/scripted_effects/00_semiconductor_market_effects.txt
```diff
@@ -0,0 +1,275 @@
...
+		}
+		subtract_from_variable = { global.semiconductor_market_access_tension = 1 }
+	}
+}
+
+semiconductor_market_deliver_tools = {
+	if = {
+		limit = { check_variable = { global.semiconductor_market_tool_backlog > 0 } }
+		subtract_from_variable = { global.semiconductor_market_tool_backlog = 1 }
+
```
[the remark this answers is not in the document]
JasonBreen: Correct, and it was worse than inert: all four owner effects (`HOL_asml_register_samsung_tool_order`, `HOL_asml_complete_samsung_tool_order`, `KOR_samsung_receive_asml_tool_delivery`, `KOR_samsung_receive_secondary_supplier_delivery`) had no caller at all — their only callers were the dashboard decisions this slice drops, which is also 4 of the 6 unused-scripted warnings. Fixed in 36d4d03ae4: accepting the Samsung co-investment now registers the order with ASML, and `semiconductor_market_deliver_tools` completes it on ASML's side and hands Samsung the tools instead of moving the backlog behind their backs, falling back to the secondary supplier when ASML has nothing pending. The two remaining unused-scripted warnings were dead triggers and are deleted.

common/scripted_effects/KOR_samsung_effects.txt
```diff
@@ -0,0 +1,464 @@
...
+	KOR_samsung_initialize_state = yes
+	if = {
+		limit = { has_game_rule = { rule = rule_corporate_history option = full } }
+		country_event = { id = KOR_samsung_events.1 days = 243 }
+	}
+	else_if = {
+		limit = { has_game_rule = { rule = rule_corporate_history option = outcomes_only } }
+		KOR_samsung_record_nand_leadership = yes
+	}
+}
```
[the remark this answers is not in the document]
JasonBreen: Right, that was mine — the milestones were generated with the init hoisted above the branches, so the yearly pulse seeded state even with the rule disabled. Fixed in 36d4d03ae4: `*_initialize_state` now sits inside both the `full` and `outcomes_only` branches across all 29 milestones, matching `ENG_arm_milestone_*`.

AngriestBird (commented):
## Bottom line

Two majors. The 2012 co-investment event ignores HOL's invite decision, and the AAT ASML capstone swap no longer has an idea to remove.

## Findings

**KOR_samsung_milestone_2012** (`common/scripted_effects/KOR_samsung_effects.txt:345`) always queues `KOR_samsung_events.5` on day 217. HOL option a sends a second copy from `events/HOL_asml_events.txt:172`. The Korean event never checks that HOL invited. If HOL finances internally, Korea still gets "ASML has invited Samsung," can accept, and `HOL_asml_record_samsung_coinvestment` stamps `HOL_asml_samsung_coinvestment_accepted` plus a tool order. `fire_only_once` only blocks a duplicate fire. Drop the yearly `full` schedule and let HOL option a be the only sender.

**HOL_asml_dutch_technological_dominance** (`common/national_focus/05_netherlands.txt:2739`) still does AAT `HOL_asml1` -> `HOL_asml3b`. `HOL_asml_strategic_partnerships` now always consumes `HOL_asml1` for `HOL_asml2a` (`:2667`). AAT used to keep `HOL_asml1` on that focus. `swap_ideas` no-ops without the remove idea, so AAT HOL never gets the tree capstone. Point the AAT branch at `HOL_asml2a`.

Copilot's init/delivery notes are already fixed. The HOL collapse stall is the chain's stated design.

BLUF

JasonBreen:
Both fixed in 93891da54f.

**Co-investment sender.** Confirmed — `KOR_samsung_events.5` only ever checked Korea's own flags, so the yearly schedule fired it regardless of what the Netherlands chose, and accepting wrote `HOL_asml_samsung_coinvestment_accepted` and a tool order onto a HOL that had taken the internal-financing branch. Dropped the `full` schedule from `KOR_samsung_milestone_2012`; HOL option a is now the only sender, and the event has exactly one caller. `outcomes_only` still records the historical co-investment on both sides, so that path is unchanged.

**AAT capstone.** Confirmed too, and `HOL_asml_strategic_partnerships` is a hard `prerequisite` of `HOL_asml_dutch_technological_dominance`, so `HOL_asml1` is guaranteed gone by the time the capstone runs — the swap could never match. Both branches now consume `HOL_asml2a`. The chain reads `add_ideas = HOL_asml1` → `HOL_asml2a` → `HOL_asml3a`/`HOL_asml3b`, with each idea removed once and added once.

Re-ran the port checks after the change: flag reachability clean both directions, all 29 milestones still defined and called, validators clean.

JasonBreen:
Swept the chain for more of the two classes you found. One more real bug, fixed in fe7a164d46.

**`KOR_samsung_apply_global_galaxy_ecosystem` was effectively unreachable.** `KOR_samsung_resolve_capstone` evaluated the memory-champion branch third, ahead of the Galaxy one. That branch is the only outcome with no strategy flag and it carries the loosest bars in the set — memory leadership above 8 and supply resilience above 5, against starting values of 7 and 4 with 13 and 18 points of gain spread across 10 and 16 write sites. A country on the Galaxy route that picked up two points of each anywhere in the chain, which is close to unavoidable, took the memory outcome instead. The three flag-gated outcomes are now evaluated first and the unflagged one stays as the fallback for a country that committed to no strategy. Introduced in `85092d8c3e` and never revised, so nothing deliberate is being undone.

Also checked and found sound, listing them so they are not re-investigated:

- **K9 Thunder ideas.** `KOR_samsung_k9_thunder2` and `3` are removed by the transfer effect but never granted, which looks like the AAT bug. It is not: `KOR_hanwha_thunder_tests` is gated `date > 2015.6.29` and calls `KOR_samsung_transfer_techwin_to_hanwha` before its own swap, so a pre-2015 country on the Samsung line is converted first, and the granting focus already branches on ownership. Only the tier-2 and tier-3 transfer branches are unreachable — dead code, no gameplay effect.
- **Market coupling, both directions.** Backlog can only be non-zero if Korea accepted the co-investment, so neither the ASML delivery nor the secondary-supplier fallback can fire for a Korea that never ordered.
- **Flag reachability.** No option or event trigger in either chain gates on a flag nothing sets, and no flag is set that nothing reads.
- **ASML capstone thresholds.** All four gates are satisfiable, and each is satisfiable while the earlier ones fail, so no ASML outcome is shadowed. `HOL_asml_strategic_control` reaches 10 against a `> 7` gate.
- **Idea chains in both trees.** Every idea added once and removed once; terminal ideas never removed.

Not yet swept: the per-option variable arithmetic against each individual historical route, which would need path enumeration rather than range checks.

AngriestBird (approved):

common/scripted_triggers/HOL_asml_triggers.txt
```diff
@@ -0,0 +1,19 @@
+HOL_asml_high_na_innovation_monopoly_qualified = {
```
AngriestBird: Fold into a 99_HOL

common/scripted_effects/KOR_samsung_effects.txt
```diff
@@ -0,0 +1,478 @@
+KOR_samsung_initialize_state = {
```
AngriestBird: Fold into 99_KOR

AngriestBird (changes requested):

common/scripted_triggers/HOL_asml_triggers.txt
JasonBreen: Folded into `99_HOL_scripted_triggers.txt` in 4376f576b0, behind a one-line section marker in the style that file already uses. `HOL_asml_effects.txt` went into `99_HOL_Scripted_Effects.txt` the same way, since the same reasoning applies to it.

common/scripted_effects/KOR_samsung_effects.txt
JasonBreen: Folded into `99_KOR_scripted_effects.txt` in 4376f576b0. `00_semiconductor_market_effects.txt` is the one file I left standalone — it is the shared layer both countries drive rather than either one's content, which is how `00_gpu_development_effects.txt` is homed. Say the word if you would rather it went into 99_KOR too.

JasonBreen:
Followed through on the rest of the folding in fc2941773f, so this chain now adds no files of its own outside localisation.

- `HOL_asml_events.txt` into `events/05_netherlands.txt` and `KOR_samsung_events.txt` into `events/Korea.txt`, matching how the Arm chain sits in `05_united_kingdom.txt`. Each carries its own `add_namespace` across, so every event id is unchanged and all 30 fired ids still resolve to a definition.
- `00_semiconductor_market_effects.txt` into `99_KOR_scripted_effects.txt` as well, reversing what I said in the thread above. It leaves nothing standalone.

Net effect on the file list: the four new `.txt` files this PR used to add are gone, and the content sits in `99_HOL_Scripted_Effects`, `99_HOL_scripted_triggers`, `99_KOR_scripted_effects`, `05_netherlands.txt` and `Korea.txt`.

TemplarGeneral:
I already added an ASML national spirit for the Netherlands with Country Content so i would suggest integrating them and/or tie this content into the ASML focus tree

JasonBreen:
Agreed, and thanks for flagging it — let's integrate rather than run two ASML systems side by side. Taking the swap route.

**What overlaps today.** Your tier ladder is `HOL_asml1` → `HOL_asml2a` → `HOL_asml3a`/`HOL_asml3b`, granted through the focus tree. This PR adds five terminal outcome ideas plus two market-pressure ones. Nothing collides by name, so nothing breaks — but a country that finishes the tree *and* the chain holds both stacks at once, themed the same and sharing `dutch_asml_idea` for art. Your tier 3 gives `research_bonus` on CAT_3d, computing, microchips and Civilian at 0.10 each; the chain's top outcome gives `research_speed_factor 0.05` and `microchip_export_multiplier_modifier 0.10`.

**The change.** `HOL_asml_resolve_terminal` will swap off your tier idea into the chain outcome instead of adding alongside it, so ASML settles on a single idea whichever route the player took.

One detail I want to get right rather than repeat a bug this PR already had to fix: `swap_ideas` silently no-ops when the remove idea is absent, which is exactly how the AAT branch of `HOL_asml_dutch_technological_dominance` was leaving Arms Against Tyranny runs with no capstone at all. A country can reach the chain's terminal without having taken your tree, so the resolve effect will swap when a tier idea is held and add when it is not, rather than assuming one is present.

**The one thing I would rather you decide:** which modifiers survive the merge. The obvious reading is that the chain outcome replaces tier 3 outright, since it is the later and more specific result — but that drops your four research-category bonuses, which are stronger and more targeted than what the outcome carries. I am happy to fold them into the outcome ideas instead so the tree's investment still shows up in the endgame. Say which you prefer and I will write it that way.

The integration already runs one direction, for what it is worth: seven of your focus completion effects now initialise chain state and move its variables, so the tree feeds the chain today. This closes the loop the other way.

JasonBreen:
Implemented in 09d3c27ff8, taking the folding-in route rather than a plain replacement, so the tree's investment is not thrown away at the end of the chain.

**The swap.** Each of the five terminal outcomes now swaps off whichever tier the country is actually holding — `HOL_asml3a`, `HOL_asml3b`, `HOL_asml2a` or `HOL_asml1` — and only falls back to `add_ideas` when none is present. All four rungs are covered rather than just tier 3, because the chain can resolve at any point in the tree, and the plain add is kept because the chain can finish without the tree being taken at all.

**The bonuses.** The five outcomes now carry `CAT_3d`, `CAT_computing_tech`, `CAT_microchips` and `CAT_Civilian` at 0.10 — the set `HOL_asml3b` grants — on top of their own modifiers. So finishing the tree and then the chain leaves you strictly better off than before rather than trading one stack for another.

**The one deliberate loss, which is yours to overrule.** `HOL_asml3a` also carries `CAT_Military = 0.02`, and I left it out of the outcomes. A country that took 3a therefore gives up that single line; 3b is break-even and everything else improves. I kept it out because it is the one bonus specific to the non-AAT branch, and folding it in would hand it to routes that never earned it — including a chain-only run that skips your tree entirely. Say the word and I will add it.

Worth naming the trap, since this PR already hit it once: `swap_ideas` no-ops silently when the removed idea is absent. That is exactly how the AAT branch of `HOL_asml_dutch_technological_dominance` was leaving Arms Against Tyranny runs with no capstone idea — it removed `HOL_asml1` when `HOL_asml_strategic_partnerships` had already consumed it. Fixed earlier in this PR, and the reason the new swap checks each tier rather than assuming one.

JasonBreen:
Fixed — this was reverting your Netherlands work, and it shouldn't have been. The diff against main is now **2603 insertions and zero deletions**.

Restored:
- `microchip_plant` rewards and the `can_staff_an_microchip_plant` gates across the ASML branch
- `specialization_nuclear` and the nuclear facility construction on `HOL_asml_foundation_of_innovation`
- the AAT arms-factory branch and its `can_staff_an_arms_industry` gate on `HOL_asml_strategic_partnerships`
- `available = { microchip_plant > 9 }` and the asml1/asml2a ladder on `HOL_asml_dutch_technological_dominance`
- the 70/30 `random_list` in `HOL_economy.021`, including the `microchip_plant` half I had flattened away
- the `KOR_samsung_thunder_*` focus ids and their k9_thunder ideas
- your English descriptions

The ASML axes are now raised from rule-gated `if` blocks appended to the four existing completion rewards, so the corporate-history layer hooks into your tree instead of replacing its effects. The five new Samsung focuses sit on free grid coordinates rather than taking the Thunder branch's slots.

One thing I'd rather ask than change unilaterally: the K9 Thunder line went to Hanwha when Samsung Techwin was sold in 2015, so the focuses arguably belong under a Hanwha id. I had renamed them and have reverted that — happy to do it properly in a separate PR if you want it, or leave it alone.

TemplarGeneral (dismissed):

AngriestBird (dismissed):

AngriestBird (approved):

PR #3418, opened 2026-09-02
Do not let a short-of-cash AI member stall NATO accession

JasonBreen:
### Changes

- The AI bypasses the 20 political power check in `NATO_auto_ratify_accession`. The player automation path keeps the gate and keeps respecting `automation_pp_reserve`

The charge itself is unchanged: the member is still billed the flat 20, and political power floors at zero, so one below the cost spends down to nothing rather than going negative. That is what the other 2,342 `add_political_power = -N` sites in the mod already rely on, so this does not introduce a new assumption. Billing only the available amount would need a variable-valued charge for no difference in outcome.

### Why

#3379 let AI members ratify through the weekly pass but left the political power gate in front of it. That gate belongs to the player's automation settings — it exists so automation does not spend below the reserve the player set. The AI has no reserve and no budget to protect.

Accession needs **every living member**, so one member sitting below 20 blocks the candidate for everyone. Nothing times out, and the member is re-tested every week against the condition that just failed. The `ratify_country_accession` weights say a member is always meant to ratify: `base = 500`, and not one of the eight modifiers is `factor = 0`.

### Confirmed against a live save

A 2002 single-player save, read directly. Every NATO member holds the `NATO_member` idea, so the idea is not the gate — political power tracks the stall exactly:

| Member | Political power | Ratified |
| --- | ---: | --- |
| Spain | 16.1 | Austria, Sweden — **not Finland** |
| Bulgaria | 18.2 | **none** |
| Germany | 21.5 | **none** |
| Denmark | 8.9 | all three |
| Czechia | 21.8 | all three |
| Luxembourg | 212.9 | all three |
| every other member | 24 – 258 | all three |

Spain is the clearest case: it ratified Austria and Sweden while it had political power to spare, then fell to 16 and stopped, and Finland's accession is still waiting on it. Bulgaria joined on 20 April with 18 and has never ratified anything. Germany hovers just at the line and has never ratified in four months.

Denmark shows the flip side — 8.9 now, but all three ratified, because it happened to be above 20 on the weeks the pass ran. Whether the alliance can expand comes down to where a member's political power sits on one particular tick.

The consequence in that save: **zero accessions completed**. Sweden 18 of 19, Austria 18 of 19, Finland 17 of 19, no `join_NATO` fired in four months, while the weekly pass itself was working — 42 automatic ratifications plus 11 by decision.


common/scripted_effects/01_NATO_effects.txt
```diff
@@ -37,10 +37,17 @@ NATO_auto_ratify_accession = {
 				}
 				if = {
 					limit = {
-						check_variable = {
-							var = automation_available_pp
-							value = 20
-							compare = greater_than_or_equals
+						# The reserve is a player setting, so the AI is not held to it.
+						# Accession needs every living member, so one member sitting
```
AngriestBird: comment isn't needed.

AngriestBird (changes requested):

common/scripted_effects/01_NATO_effects.txt
JasonBreen: Removed in the latest push. The reasoning is in the PR description instead.

AngriestBird (approved):

PR #3423, opened 2026-09-02
Give Iran a working set of AI paths (#3162)

TemplarGeneral:
Part of #3162. Iran's `PER_ai_behavior` was one of the three rules in that issue that are **defined but never read by any loaded script** — a repo-wide grep returned exactly one hit, the definition. No `PER_*_FOCUS_PATH` flag existed, there was no branch in `999_game_rules_on_actions.txt`, and `has_game_rule` appeared zero times in the 608-focus tree. The player picked a path and nothing happened.

Underneath that sat a bigger problem: **an AI Iran could not leave the Islamic Republic at all.** All twelve of the tree's political spines are post-collapse content gated behind the clergy balance-of-power chain, and every gate on that chain was shut to the AI.

## The rule

Nine options plus `RANDOM_PATH` and `NO_PATH` (the `default = { }` block, last). `DEFAULT` and `RANDOM` are gone; `PER_NO_REVOULT_IRAN` ("No Revolutions") was the historical path under another name and merges into `HISTORICAL`. Every alt option now owns a real spine — before this, eight political spines (~180 focuses) belonged to nobody.

| Option / flag | Spines owned | Ruling party | Entry |
| --- | --- | --- | --- |
| `HISTORICAL` | the six Islamic Republic trunks + `PER_VAJA` + the historical economy fork | 8 / 9 | start state |
| `RESTORATION` | `PER_oath_of_barandaz` | Western_Autocracy 0 | `iranian_events.234.a` |
| `IRANIC_FEDERATION` | `PER_party_of_iranians` | Nat_Populism 20 | `iranian_events.234.b` |
| `MOJAHEDIN` | `PER_maryams_plan` | anarchist_communism 5 | **new** `iranian_events.235.a` |
| `SECULAR_REPUBLIC` | `PER_march_towards_freedom`, `PER_god_science_freedom` | Neutral_Libertarian 16 / Neutral_conservatism 14 | `iranian_focus.158.a` / `.c` |
| `GREEN_PATH` | `PER_path_of_hope`, `PER_the_masculine_cat` | Neutral_MB 12 / Neutral_Autocracy 13 | `iranian_focus.158.b`, `iranian_events.226` else-branch |
| `HEIRS` | `PER_ahmadinejads_opportunity`, `PER_qalibafs_movement`, `PER_mashaeis_approach` | Autocracy 6 / oligarchism 15 / Nat_Fascism 21 | `iranian_events.233.a/.b/.c` |
| `WORKERS_REPUBLIC` | `PER_liberty_equality_workers_rule` | Neutral_Communism 19 | `iranian_events.235.b` |
| `MILITARY_JUNTA` | `PER_restore_order` | Nat_Autocracy 22 | `iranian_events.233.d1` |

379 of 608 focuses are owned; 229 stay path-neutral (the economy convergence both economy roots share, the army/ministry/agriculture/energy trunks, and the three non-IR foreign-policy trunks whose `available` already excludes an Islamic Republic government). Convergence focuses reached from two spines are owned by an alias trigger (`PER_ai_reform_path`, `PER_ai_nationalist_path`) rather than one side's flag.

## Making the paths reachable

Four gates stood between a path-flagged AI and any alt spine. All four are now keyed to the path triggers, so a human game and a `NO_PATH` AI game are untouched:

1. **The BOP decisions** pushed the AI the wrong way — four clergy-positive decisions at `ai_will_do base 10` against `PER_relax_religious_laws` at **base 0**. The four now zero out under `PER_ai_not_historical_path`; the secularizing one gets `add = 20`.
2. **New AI-only category** `PER_ai_path_category` (`is_ai = yes`, no localisation keys) with `PER_ai_undermine_the_clergy` (-0.10 BOP, 60-day cycle — the BOP has to travel -1.15 from its `initial_value = 0.25`) and `PER_ai_ferment_unrest`, which restarts the protest wave. `iranian_events.300` fires once and its only protest branch was `ai_chance base = 0`.
3. **The 220-225 volatility walk** carried no `ai_chance` at all, so `PER_revolution_volatility > 2` — the gate on the branch fork — was a five-step coin flip. Each escalate option is now favoured 20:1 under a non-historical path. `iranian_events.228.a`, which had no `ai_chance` and lost 1-against-10 to two siblings, gets one too.
4. **The fork events** `.232` / `.233` / `.234` / `.235` and `iranian_focus.125` / `.158` / `.1261` were all flat `ai_chance = { base = 1 }`. Each option now boosts on its owning flag and zeroes on the rival ones; `.232.d1` (do nothing) is zeroed under any path.

Three spines were unreachable outright and are fixed at the source:

- **`PER_maryams_plan`** — nothing anywhere set `rul_party_temp = 5`. `iranian_events.235` gains a Mojahedin-e Khalq option; its loc key `iranian_events.235.a` was already in the file, written and never wired.
- **`PER_qalibafs_movement`** and **`PER_mashaeis_approach`** — `iranian_events.233.b` / `.c` set the party but never installed the leader the spines' `available` requires. Fixed through the roster (`set_country_flag = set_oligarchism` / `set_Nat_Fascism` then `set_leader = yes`), not an inlined `create_country_leader`.
- A pre-2009 collapse takes the Montazeri branch of `iranian_events.226.a`, which never fires `iranian_focus.158` and so stranded `SECULAR_REPUBLIC`. That branch now fires it for an AI on that path.

## AI hardening

- `ai_is_threatened` was used **zero** times across all 608 focuses. 55 combat-capacity focuses (the Artesh and Ministry of War trunks, the IRGC service focuses) now carry `factor = 2`.
- `common/ai_strategy/PER.txt` had no losing-war brake. Added `PER_cancel_war_neighbours` on the `ALG` model — `surrender_progress > 0.15` against IRQ, ISR, SAU, AZE, AFG, TUR, PAK, UAE. No per-tag `avoid_starting_wars`: `MD_avoid_new_wars_when_outmatched` already covers it.
- 30 intra-path mutex forks previously resolved by file order. Each now has a deterministic preferred side chosen to match its path's identity — Pezeshkpour over Forouhar for the Pan-Iranists, Yazdi over Bazargan for the Freedom Movement, Taqvaee and secularism for the Worker-Communists, Tehran over Tabriz for VAJA's headquarters.
- Fixed the log id in `iranian_events.235.b`, which cited `.235.a`.

## Deliberately not done

- **No government walker.** The report verdicts an undated successor roster (22 sub-ideology blocks, one dated entry in the whole file), and Iran already owns its election flow behind `generic_election_killswitch` with its own `PER_election_events` / `PER_auto_update_gui_weekly` machinery. A walker over an undated roster installs the wrong person.
- **No strategy plans.** Iran has none today and needs none — every path is expressed in the tree's own weights, and `NO_PATH` runs on sane bases.
- `iranian_events.226.a` passes `disable_elections = 1` in both branches and nothing for PER ever passes `disable_elections = 0`, so every post-collapse Iran is a permanent no-election state. That guarantees a path survives its own election, but it is pre-existing behaviour that affects humans too — left alone here.
- `PER_a_new_dawn`, `PER_decentralized_state` and `PER_one_nation` still report as orphans under `HISTORICAL` and `NO_PATH`. They sit behind `PER_striking_the_great_satan`, which is deliberately dormant at `base = 0` with a conditional `add = 10`. Same three orphans the report showed before this branch.

## Verification

- `ai_path_report.py --tag PER` — rule and wiring clean, **0 orphans and 0 stranded focuses in all eight alt states**, 0 path-relevant mutex ties (was 30), no additive path modifiers, no unreferenced flags, clean `government` section.
- `validate_focus_tree.py --path .` — no issues.
- `validate_common_mistakes.py` — no issues.
- `validate_localisation.py` — no issues.
- `validate_decisions.py` — warning and error counts identical to a baseline taken before the first edit.
- `validate_events.py` — no findings in `Iran.txt`.


common/decisions/Iran.txt
```diff
@@ -15344,3 +15349,52 @@ PER_us_war = {
...
+
+		available = {
+			has_civil_war = no
+			NOT = { has_country_flag = PER_protests_ongoing }
+			has_power_balance = { id = PER_clergy_balance_of_power_category }
+		}
+
+		complete_effect = {
+			log = "[GetDateText]: [Root.GetName]: Decision PER_ai_ferment_unrest"
+			country_event = { id = iranian_events.301 days = 1 }
```
KianGhk1530: Remove this event, otherwise you'll have a revolution in Iran every 180 days.

common/decisions/Iran.txt
```diff
@@ -15344,3 +15349,52 @@ PER_us_war = {
 		}
 	}
 }
+
+PER_ai_path_category = {
```
KianGhk1530: No need for a separate category, integrate the 2 decisions into the existing BoP and make them only visible to the AI

common/decisions/Iran.txt
```diff
@@ -15344,3 +15349,52 @@ PER_us_war = {
...
+
+	PER_ai_undermine_the_clergy = {
+
+		icon = generic_decision
+
+		cost = 50
+
+		days_re_enable = 60
+
+		available = {
```
KianGhk1530: Also, for both decisions, disable them/make them no longer visible once the IR collapses because then you're just having the AI spam decisions that cost it PP for no gain.

events/Iran.txt
```diff
@@ -17938,6 +18010,10 @@ country_event = {
 				country_event = { id = iranian_flavor.64 days = 140 random_days = 30 }
 			}
 		}
+		ai_chance = {
+			base = 1
+			modifier = { factor = 100 PER_ai_not_historical_path = yes }
```
Blazer135, line 18023: [P1] Only option .228.a schedules event 232, but this weight of 100 competes with unchanged base-10 weights on .228.b and .228.c. The six non-historical paths whose government is selected in event 232 therefore have a combined 16.7% chance to skip their selector entirely. Give .b and .c factor 0 while PER_ai_not_historical_path is true, or otherwise make the route into event 232 deterministic.

events/Iran.txt
```diff
@@ -19405,6 +19489,7 @@ country_event = {
 		log = "[GetDateText]: [This.GetName]:iranian_events.232.d1 executed"
 		ai_chance = {
 			base = 1
+			modifier = { factor = 0 PER_ai_not_historical_path = yes }
```
Blazer135: [P1] This disables .232.d1 for every non-historical path, but SECULAR_REPUBLIC and GREEN_PATH match none of the families handled by .232.a, .232.b, or .232.c, so all four option weights become zero for those reform paths. Keep .d1 selectable for PER_ai_reform_path, or restrict this zero-weight modifier to the six paths routed through the other options.

Blazer135 (changes requested):

events/Iran.txt
TemplarGeneral: Fixed in 4ff897e. Both `.228.b` and `.228.c` now carry `modifier = { factor = 0 PER_ai_not_historical_path = yes }`, so `.228.a` (100) is the only live option and the route into `.232` is deterministic for every non-historical path. Historical and `NO_PATH` weights are unchanged.

Same class of miss one event upstream: `iranian_events.225.b` was the only de-escalate option in the 221-225 walk without an `ai_chance` block, so it defaulted to 1 against `.225.a`'s 20. Zeroed it to match its four siblings.

events/Iran.txt
TemplarGeneral: Fixed in 4ff897e. `.d1` now reads `modifier = { factor = 0 PER_ai_not_historical_path = yes PER_ai_not_reform_path = yes }`, so the zero skips SECULAR_REPUBLIC and GREEN_PATH.

`.d1` is the right landing for those two: they take their government from `iranian_focus.158` (fired out of `iranian_events.226`, including the new `is_ai` + SECULAR_REPUBLIC branch), so "The government will never resign!" holds the government focus.158 already installed rather than handing it to a rival family.

Resulting table for `.232`: HEIRS / MILITARY_JUNTA to `.a`, RESTORATION / IRANIC_FEDERATION to `.b`, MOJAHEDIN / WORKERS_REPUBLIC to `.c`, SECULAR_REPUBLIC / GREEN_PATH / HISTORICAL to `.d1`. `NO_PATH` with historical off stays a flat 1/1/1/1.

events/Iran.txt
```diff
@@ -19440,6 +19531,8 @@ country_event = {
 		}
 		ai_chance = {
 			base = 1
+			modifier = { factor = 50 has_global_flag = PER_HEIRS_FOCUS_PATH }
```
KianGhk1530: Not necessarily related to this specific line, but in the event where the game brings you to pick a leader, the PER_HEIRS_FOCUS_PATH should be split into four, Mashaei, Ghalibaf, Ahmadinejad, and the Artesh. Otherwise it's still random and not really hardened AI paths that would give the player complete choice. All 4 paths have different outlooks & content of how the interact externally.

Likewise if there's something like this where it bunches up based on the decision, the monarchists & pan iranists should be separated, etc.

PR #3452, opened 2026-09-02
Give Hong Kong a country AI path game rule (#3162)

TemplarGeneral:
Part of #3162.

## Problem

Hong Kong had **no `HKG_ai_behavior` rule at all**, so its whole political fork was unreachable for the AI.

The four branch roots — `HKG_loyal_puppet`, `HKG_great_hong_kong_nation`, `HKG_little_bird_under_british_wing`, `HKG_city_state_idea` — are mutually exclusive and carry no `available` block; they are gated purely by `ai_will_do` modifiers on `has_government`. Hong Kong starts `ruling_party = communism`, so the three alternate roots sat at `factor = 0` from turn one, and nothing in the country grew the parties that unlock them (nationalist popularity starts at 0.1%). The AI could only ever walk the loyal-puppet branch.

The four `HKG_*_plan` strategy plans already described the four branches, but they were gated on `has_government` rather than path flags, and every one carried `abort = { is_subject = no }` — Hong Kong starts as a Chinese SAR and all three alternate roots run `overlord = { end_puppet = HKG }`, so each alternate plan aborted the instant its own root focus completed.

## Changes

**Game rule and localisation.** `HKG_ai_behavior` with `HISTORICAL`, `INDEPENDENCE` ("A Nation of Our Own"), `BRITISH` ("Back Under the Crown"), `CITY_STATE` ("The Free City"), `RANDOM_PATH`, and `NO_PATH` as the `default = { }` block. `HKG_AI_BEHAVIOR: "@HKG Hong Kong"` added to the replace file; `RANDOM_PATH` / `NO_PATH` reuse the shared `RULE_OPTION_MD_*` keys.

**Flag wiring.** `HKG = { ... }` block in `999_game_rules_on_actions.txt` setting `HKG_<PATH>_FOCUS_PATH` global flags, with the historical bucket included in the `RANDOM_PATH` `random_list` (40/20/20/20). `NO_PATH` sets nothing.

**Scripted triggers and focus ownership.** Nine AI-internal triggers in `99_HKG_scripted_triggers.txt`; 36 of 107 focuses re-owned across seven groups via `apply_ai_path_weights.py`. The `free_hong_kong` alias carries the democratic-reform spine for the British and city-state paths; a `not_historical` alias carries the European-business economic side.

Two deliberate neutrality calls:

- **The carrier chain stays path-neutral.** Only `HKG_independence_plan` lists it, but `HKG_contract_carrier_branch_done` requires all eight carrier focuses for the China naval-contract capstone. Owning it to the independence path would kill that system under the historical path.
- **`HKG_gong_zyu_ou_integration` stays un-owned at `ai_will_do = { base = 0 }`.** Its reward has China `annex_country = { target = HKG }` — Hong Kong deletes itself. It was also removed from the pro-China plan's focus list, where it could never fire.

`HKG_west_kowloon_cultural_district` is mutually exclusive with `HKG_suppress_cantonese_local_culture` and is killswitched under the historical path; `HKG_world_class_destination` takes `disney OR west_kowloon`, so nothing is stranded.

**AI-only popularity ramps.** New `HKG_ai_path_category` (`is_ai = yes`, no localisation) with rally/takeover pairs for Nat_Populism (index 20), conservatism (index 1) and oligarchism (index 15). All three pass `temp_outlook_increase`: reading `change_relative_party_popularity`, the no-outlook branch redistributes only inside the target's own ideology group, and each of these parties either is its whole group (oligarchism 29.7 = neutrality 29.7), starts at zero (Nat_Populism), or is capped below 50% by its group ceiling (conservatism 23.8 inside democratic 33.3). Elections are not disabled — the rally decision re-fires below 55% popularity, so the party wins its own re-election rather than being installed by fiat, which matters on two paths whose own focuses legislate universal suffrage.

**Strategy plans.** `enable` now reads `has_global_flag = HKG_<PATH>_FOCUS_PATH` OR (`HKG_ai_no_focus_path = yes` AND the plan's original government condition), so `NO_PATH` behaviour is preserved exactly and the AI always has a live plan, while a chosen path is authoritative even while its ramp is still running. `abort = { is_subject = no }` dropped from the three plans whose root focus ends the puppet; kept on `HKG_pro_china_plan`, where it is meaningful. The pro-China, independence and city-state lists were missing housing focuses that are hard prerequisites of focuses they do list — added.

**AI strategy.** `common/ai_strategy/HKG.txt` held only naval-contract strategies. `HKG_irredentist_claim_pearl_river` adds a claim on state 532 and a named threat against China, so a target-keyed `declare_war id = "CHI" value = -4000` brake was added while Hong Kong is weaker than China. Not a targetless `avoid_starting_wars` — the mod-wide `MD_avoid_new_wars_when_outmatched` already covers that and a stricter per-tag copy would be dead code.

**Dead focus fixed.** `HKG_global_wealth_management` required `HKG_belt_and_road_financing` **and** `HKG_green_finance_centre`, which descend from the mutually exclusive `HKG_chinese_entrepreneurs` / `HKG_european_business` pair. It was unreachable for every player and every AI path while sitting in three plan focus lists. Its two economic prerequisites are now an OR group.

**Historical government.** `ai_path_report.py` verdicts *undated successor roster* — every sub-ideology roster has zero dated entries — so **no walker was written**. The ramp decisions deliver the party; a walker over an undated roster would install the wrong person.

## Verification

`ai_path_report.py --tag HKG`: rule and wiring clean, 4 path flags, 9 scripted triggers, **0 orphans and 0 stranded focuses in all ten option × historical-AI states**, no additive path modifiers, no unreferenced flags, and mutex ties down from 7 to 0.

`validate_focus_tree.py --path .` reports nothing for Hong Kong. `validate_decisions.py` is unchanged against a pre-change baseline (283 errors / 459 warnings / 69 files) — the new `is_ai = yes` category adds no `ai-only-decision-localisation` findings. `precommit_validate.py` green across all eight validators.

## Report-only findings (judged, not fixed)

- **`Basic_Law_HKG` is a bonus, not a burden.** Stability +0.1, trade opinion +0.10, political power +0.05, factory efficiency +0.05, consumer goods −0.03 — all favourable. The two focuses that remove it (`HKG_hong_kong_state_constitution`, `HKG_revise_basic_law`) replace it with their own constitution as a path outcome, so the report's "every cure is dead" line is a path cost, not a stranded malus.
- **`HKG_idea_railway_development_strategy_2000`** (`political_power_factor = -0.05`) is a real malus with no cure in any focus or decision, in any rule state, before or after this change. Killswitching takes nothing away from it.
- **`CHI_HKG_integration_modifier`** has backing variables that are all `CHI_*` and unset on Hong Kong at start; it applies as a 0-effect modifier driven entirely by China's own focuses.
- **`HKG_housing_policy_modifier`** is never removed by design — the eight `HKG_housing_*` variables behind it are the real burden, and their cures sit in the path-neutral housing chain, live in every rule state.
- **`HKG_protest_decision_categories`** is `visible = { always = yes }` with `visible_when_empty = yes` and contains no decisions — an empty always-on tab, pre-existing.
- **No `focus_factors`** added to the plans: ownership is carried by the tree's `ai_will_do`, and duplicating it into the plans risks the two disagreeing.


common/game_rules/00_game_rules.txt
```diff
@@ -2764,6 +2764,30 @@ GRE_ai_behavior = {
...
+		name = BRITISH
+		text = "RULE_OPTION_HKG_BRITISH"
+		desc = "RULE_OPTION_HKG_BRITISH_DESC"
+	}
+	option = {
+		name = CITY_STATE
+		text = "RULE_OPTION_HKG_CITY_STATE"
+		desc = "RULE_OPTION_HKG_CITY_STATE_DESC"
+	}
+}
```
Blazer135, line 2800: [P1] The latest merge closes HKG_ai_behavior here without the RANDOM_PATH option or the default NO_PATH option that existed in the original PR commit. The HKG random branch in 999_game_rules_on_actions.txt is therefore unreachable, and the rule has no explicit default. Restore both option blocks before this closing brace.

common/ai_strategy_plans/HKG_strategy_plans.txt
```diff
@@ -312,16 +333,22 @@ HKG_city_state_plan = {
...
 
-	abort = { is_subject = no }
-
 	ai_national_focuses = {
 		HKG_public_housing_problem
 		HKG_sun_housing_policy
+		HKG_end_tenants_purchase_scheme
 		HKG_revitalise_private_market
 		HKG_relax_mortgage_rules
+		HKG_suspend_home_ownership_scheme
```
Blazer135, line 353: [P1] This CITY_STATE sequence reaches HKG_green_form_subsidised_home_ownership_scheme without HKG_subsidised_sale_flats_project, although Green Form has separate prerequisite blocks requiring that focus alongside land reclamation and My Home. The forced plan cannot complete Green Form or its descendants. Add HKG_subsidised_sale_flats_project to this sequence; HKG_british_plan has the same omission and needs the same fix.

Blazer135 (changes requested):

Blazer135 (approved):

PR #3455, opened 2026-09-02
Fix four dynamic flag names that can never match

JasonBreen:
MD builds per-partner flags by appending a scope token: `trade_agreement@TAG`, `mutual_investment_treaty_@TAG`. Four names are written in a form the parser cannot resolve to what reads them, so the checks are permanently false.

## Missing `@`

`common/scripted_diplomatic_actions/00_scripted_diplomatic_actions.txt` sets the flag with the `@`; these three checks omit it.

| Written | Should be | Sites |
|---|---|---|
| `trade_agreementCHI` | `trade_agreement@CHI` | 1 |
| `trade_agreementUSA` | `trade_agreement@USA` | 4 |
| `mutual_investment_treaty_USA` | `mutual_investment_treaty_@USA` | 4 |

**`NKO_expand_sezs` could never be completed.** Its `available` wrapped `ROOT = { has_country_flag = trade_agreementCHI }` in a `custom_trigger_tooltip`, so the focus stayed locked behind a tooltip telling the player to propose a trade agreement to China — exactly what the corrected flag records.

**The Great Recession reverberation scaling was dead.** `usa_economic_events.36` fires at countries under heavy American influence (plus CHI/JAP/KOR/TAI) and picks a 240-, 180- or shorter-day `reverberations_of_the_great_financial_crisis` from whether the recipient holds a trade agreement and/or investment treaty with the US. With both flags unsettable, every recipient always fell to the `else`.

## Token not last

`set_country_flag = libya_casablanca_accords_@ROOT_left` is the **only** flag in the mod with text after its `@` token — the other 207 all end on the scope reference. The four decisions reading `libya_casablanca_accords_<TAG>_left` never match it, so a participant who signs the Casablanca Accords and later walks away leaves no way to re-approach them.

Both sides move to `libya_casablanca_accords_left_@ROOT`. Worth being explicit about the reasoning: I can't run the parser to prove the suffixed form fails, and the evidence is MD's own 207-to-1 convention rather than a vanilla precedent (vanilla ships no dynamic flags at all). The rename is safe either way — if the old form did resolve, this is an equivalent rename; if it did not, it repairs it.

## Provenance

Not deliberate removals — `git log -S` puts each malformed name in at the commit that introduced its block (#278 for the GFC refactor, #2210 for the North Korea sprint), never in a working form.

## Method note

Found by sweeping every `has_country_flag` read against every setter. A line-based grep misses the multi-line `set_country_flag = { flag = X days = N }` block form and reports ~170 false positives; matching that form brings the whole mod down to 11 genuinely unset flags, of which these are four. `LBA_casablanca_acccords_temp_alliance` (three c's) is left alone — the typo is consistent between its setter and its reader in `common/ai_strategy/LBA.txt`, so it works.

`validate_common_mistakes.py --staged` is clean, and `validate_events.py` cites none of the changed files, on this branch or on the baseline.


events/United States.txt
```diff
@@ -3292,8 +3292,8 @@ country_event = {
 		modify_treasury_effect = yes
 		if = {
 			limit = {
-				has_country_flag = trade_agreementUSA
```
AngriestBird, line 3295: Don't these evaluate the same?

AngriestBird (changes requested):

events/United States.txt
JasonBreen: No, they're different strings. The `@` is stored as part of the flag name, not a separator the parser strips off.

`common/scripted_diplomatic_actions/00_scripted_diplomatic_actions.txt:94` writes `trade_agreement@ROOT` (and `@PREV` for the other side), so each partner ends up holding a flag literally named `trade_agreement@<tag>`. That the `@` survives into the stored name is visible in shipped code that uses literal tags on both ends: `common/national_focus/05_saudi_arabia.txt:15894` sets `trade_agreement@SAU`, and `common/ai_strategy/AFG.txt:125` reads `trade_agreement@PAK`.

So `has_country_flag = trade_agreementUSA` looks for a name nothing ever writes and is permanently false. Same for `mutual_investment_treaty_USA`, where the setter form is `mutual_investment_treaty_@TAG` (`common/continuous_focus/generic.txt:196`).

If it helps as a sanity check: every one of the 207 other dynamic flags in the mod is written and read with the `@` in place.

AngriestBird (approved):

PR #3469, opened 2026-09-03
Remove no-op zero-valued modifiers (cwtools CW235)

AngriestBird:
## Bottom line
Drops 108 lines across 24 files: additive modifiers written with value 0 that cwtools flags as CW235. Pure deletions, zero behavior change.

## Why
These modifiers are additive, so a 0 does nothing, and autonomy levels, chief traits, and ideas swap their modifier blocks rather than stack them, so omitting the key equals writing 0 (vanilla omits rather than zeroing). The two single-entry `modifier = {}` blocks in subparty_policies.txt are removed whole so no empty block remains. `mercenary_leader`'s `nationalist_drift = 0.` may have been a truncated intended value, but as written it was a no-op.

BLUF


AngriestBird:
All no-opt dead code cleanup. Closing manually by myself. Reviewed + 1 agent review.

PR #3477, opened 2026-09-03
Give Israel a country AI path game rule (#3162)

TemplarGeneral:
Part of #3162.

## Problem

Israel had **no `ISR_ai_behavior` rule at all** — 667 focuses, zero path flags, zero scripted path triggers, zero `focus_factor` entries in `common/ai_strategy/ISR.txt`. The AI picked focuses by raw `ai_will_do` base and never walked a coherent branch.

Two structural facts drove the design:

- **The tree does not fork by prerequisite.** All 15 party spines hang off one shared OR-mesh under `ISR_speak_to_government`; the three trunks (`ISR_democratic_first`, `ISR_democratic_jewish_balance`, `ISR_jewish_first`) are separated only by their `available` ruling-party gates.
- **Israel starts as socialism/Labour (index 3, 0.233) with no drift ideas** and a 44-month election cycle, so an unassisted AI re-elects Labour forever and the other two trunks are permanently dead. Historical Israel is Likud (conservatism, index 1), so *even the historical path* needed a route to power.

## Options

Exactly `HISTORICAL` + four alt paths + `RANDOM_PATH` + `NO_PATH` (the `default = { }` block, last).

| Option | Displayed | What it is |
| --- | --- | --- |
| `HISTORICAL` | Historical | Likud after the talks collapse; second intifada, barrier, disengagement, US alignment, Abraham Accords, Iran pressure, Gaza blockade |
| `PEACE_PROCESS` | The Peace Camp | Labour with Meretz and the Arab parties; two-state, PA rearmed as a partner, Gaza port and airport |
| `SECULAR_CENTRE` | The Secular Centre | Yesh Atid / Blue and White / New Hope; religion prised out of the state, courts shielded, Europe and the OECD |
| `GREATER_ISRAEL` | Greater Israel | Otzma Yehudit and Religious Zionism; Oslo torn up, settlements annexed, Temple Mount status quo rewritten, pre-emptive strike north |
| `LOOK_EAST` | Turn to the East | A Likud government pivots from Washington: Haifa port to Chinese operators, arms to Beijing, Moscow over Kyiv |

## What changed

- **327 focuses weighted** across nine ownership groups, applied only through `tools/standardization/apply_ai_path_weights.py`. Ownership was derived from each focus's `available` party gate (with polarity — a dozen focuses gate on `western_socialism_are_in_power_or_coalition = **no**`, and a dozen more on a *foreign* country's government inside a `TAG = { }` scope) and then closed over the prerequisite graph so no spine child is left behind when its root is killswitched.
- **18 scripted path triggers** appended to the existing `99_ISR_scripted_triggers.txt` — no second ISR trigger file. Four alias groups carry spines and axes that several options share: `ISR_ai_likud_path` (Historical + Look East), `ISR_ai_hardline_path`, `ISR_ai_two_state_path`, `ISR_ai_western_alignment_path`, `ISR_ai_restraint_path`. `ISR_ai_historical_path` guards its `is_historical_focus_on` arm with `ISR_ai_alt_path`, the France/Georgia shape — unguarded, every alias built on it stays true under an explicit alternate rule and suppresses its own killswitch.
- **`ISR_ai_path_category` + three AI-only ramp pairs**, the category and the Likud pair gated on `ISR_ai_likud_path` rather than raw flags so the ramp is live under `NO_PATH` + historical AI (which sets no flag) and hidden under an explicit alternate rule — conservatism (1) shared by Historical and Look East, liberalism (2), Nat_Fascism (21). Nat_Fascism is not initialised in Israel's history file, so its ramp passes `temp_outlook_increase`; the other two sit in the incumbent's own group and correctly do not. No localisation keys, bare `check_variable` in `available`.
- **99 `ai_is_threatened` weightings** on the military branch under `ISR_speak_to_staff`, which had none at all.

## Fixed along the way

- `israel_labour.3` / `.4` / `.5` set `party_index` and `party_popularity_increase` in option `.b` and **never called `change_relative_party_popularity`** — three dead options. All three now call it. Both options of each event also gained `ai_chance`: without it the AI coin-flipped whether it ever got the `ISR_yeshatid` / `ISR_resilance` / `ISR_newhope` flags that the Secular Centre spines gate on.
- `ISR_denounce_polisario_iran` carried a bare `is_historical_focus_on = yes factor = 0` that stranded `ISR_restore_security_links` and `ISR_usa_mor_isr_deal` under historical AI. That is backwards — Israel did normalise with Rabat — so the killswitch is gone.
- 16 further legacy `is_historical_focus_on` killswitch pairs were replaced by proper path modifiers on focuses that now have an owner, including one with an `add = 5` sitting *after* a `factor = 0`.

## Judgement calls worth a second opinion

- **`ISR_binational_state` is now `base = 0`.** Its `available` requires `NOT = { has_idea = the_military }` and nothing anywhere in the mod removes that idea for Israel, so the branch root is permanently unreachable — yet it sat at a live `base = 3` over a reward that annexes PAL and HAM, runs `change_tag_from` twice, `add_stability = -0.3` and rolls a 50% nationalist civil war. Zeroing the root is deliberate and is the sole reason the report still lists 16 orphans (its subtree) under `PEACE_PROCESS` and `NO_PATH` + historical off. Filed separately as #3478.
- **No strategy plan was added.** Only 2 of 57 converted countries have one, and duplicating the tree's weights into `focus_factors` is the "plan disagrees with the tree" bug class. `NO_PATH` + historical off leaves all 667 focuses live at their authored bases, which the report confirms.
- **No per-tag war brake was added.** `MD_avoid_new_wars_when_outmatched` (`enemies_strength_ratio > 0.75`) and `MD_avoid_new_wars_while_losing` (`surrender_progress > 0.15`) already cover every country targetlessly; any ISR-specific brake would be a strict subset and dead code.
- **No historical-government walker.** The report verdicts an undated successor roster — the `date <` markers in `ISR_political_leaders.txt` are 2017-bookmark skip guards, not tenure dates — so the ramp decisions deliver the party and a walker would install the wrong person.
- The other five `start_civil_war` focuses (`ISR_puppet_jordan`, `ISR_royal_road_iraq`/`_syria`, `ISR_operation_silk_road`, `ISR_premptive_strike_lebanon`) start the civil war in the **target** country, not Israel. Left alone.
- `ISR_farright_america` and `ISR_kaplan_trials` remain stranded under Greater Israel: the first requires `has_completed_focus = ISR_balance_favor_china` while being `mutually_exclusive` with `ISR_farright_china`, the second requires a Likud-gated focus while itself requiring `nationalist_fascist_are_in_power`. Both are pre-existing content contradictions, reported not papered over.
- The Jewish Home spine (20 focuses, `western_autocrats_are_in_power`) and the Yisrael Beiteinu spine (11 focuses, `neutrality_oligarchism_in_power_or_coalition`) are left path-neutral: no option's party can satisfy their `available` gate, so owning them would be noise. Both hang off the three-way trunk OR-mesh, so neutrality does not strand them.

## Verification

- `ai_path_report.py --tag ISR` — rule and wiring **clean**, **0 orphans in every rule state** bar the deliberate binational kill, **0 path-relevant mutex ties** (12 before), **0 mutex pairs boosted by rival paths**, **0 focuses alive only via global history under an explicit rule**, no additive path modifiers, no unreferenced flags, `mechanics` clean, `government` clean. The last two are new checks added in this PR (see below); the first review pass was clean only because the tool did not yet look.
- `validate_focus_tree.py --path .` — **no issues**, unchanged from the pre-work baseline.
- `validate_decisions.py` — 290 errors / 459 warnings in 69 files, **identical to the branch baseline measured with the changes stashed**; no new `ai-only-decision-localisation`.
- `python -m pytest` — green, including 12 new regression tests for the report checks.
- `validate_common_mistakes.py`, `validate_variables.py` — no issues. `validate_events.py` — 0 errors, no Israel findings.
- Encoding verified per file: `.txt` UTF-8 no BOM, loc `.yml` UTF-8 with BOM, all LF.

## Review round 2

Blazer's three findings are fixed in 997672fd40 and answered inline.

1. **Global history leaked into every alias.** `ISR_ai_historical_path` now guards its
   `is_historical_focus_on` arm with the new `ISR_ai_alt_path`. Under `PEACE_PROCESS` + historical on,
   `ISR_blockade_gaza` goes 75 -> 0 and `ISR_propose_hudna` 25 -> 25, so the mutually exclusive pair
   resolves to the path the player picked.
2. **Likud ramp unreachable under `NO_PATH` + historical AI.** Category and both Likud decisions gate
   on `ISR_ai_likud_path` instead of the raw flags.
3. **31 focuses wrongly owned by `SECULAR_CENTRE`.** The Jewish Home and Yisrael Beiteinu spines are
   path-neutral, applied through `apply_ai_path_weights.py`. The description above now matches the
   diff — it previously claimed path-neutrality the diff did not deliver.

### Tooling and docs

`ai_path_report.py` grows the three checks that should have caught these, with regression tests:

- a focus alive under an explicit rule **only** because historical AI is on;
- a `mutually_exclusive` pair boosted by **rival** paths in one state (a pair inside one path's own
  spine stays a flavour choice);
- a decision or category gate that is dead under `NO_PATH` + historical AI, or visible during an
  alternate path.

The ownership section also lists the ruling parties each group's focuses require, which is what makes
a finding-3 style mis-ownership visible at a glance. `.claude/skills/country-ai-path/references/write.md`
gains the rule that the unguarded historical arm is safe only inside `ai_will_do`.

### Pre-existing hits the new checks report — separate PRs, not fixed here

| tag | finding |
| --- | --- |
| IRQ | 88 focuses alive under an alt rule via the unguarded `IRQ_ai_national_accord_path` — identical bug |
| EGY | `EGY_ai_rally_the_generals`, `EGY_ai_generals_take_the_country` gate on the bare historical trigger |
| IND | `IND_rally_the_awakening` likewise |
| GER, GRE | `*_ai_path_category` and their ramp decisions gate on the raw `*_HISTORICAL_FOCUS_PATH` flag |


common/decisions/categories/israel_categories.txt
```diff
@@ -134,3 +134,22 @@ israel_lebanon_occup_category = {
...
+		is_ai = yes
+		original_tag = ISR
+	}
+
+	icon = GFX_decisions_category_political
+	priority = 100
+
+	visible = {
+		OR = {
+			has_global_flag = ISR_HISTORICAL_FOCUS_PATH
```
Blazer135: [P1] The default NO_PATH setting sets no ISR_HISTORICAL_FOCUS_PATH flag, yet this category and the Likud ramp decisions are visible only with raw path flags. With historical focuses enabled, ISR_ai_historical_path suppresses non-historical political weights, but the popularity ramp never runs, so Israel can remain Labour and never enter the weighted Likud branch. Gate the category and Likud decisions on a normalized historical-path trigger, with explicit alternate rules overriding global history.

common/scripted_triggers/99_ISR_scripted_triggers.txt
```diff
@@ -346,3 +346,126 @@ ISR_electronics_outcome_settled = {
...
+ISR_ai_historical_path = {
+	OR = {
+		is_historical_focus_on = yes
+		has_global_flag = ISR_HISTORICAL_FOCUS_PATH
+	}
+}
+
+ISR_ai_likud_path = {
+	OR = {
+		ISR_ai_historical_path = yes
```
Blazer135, line 373: [P1] is_historical_focus_on remains enough to make ISR_ai_historical_path true even when an explicit alternate rule flag is set. On PEACE_PROCESS this activates both peace and hardline weights: ISR_blockade_gaza gets 75 versus ISR_propose_hudna at 25 despite being mutually exclusive. Define an alternate-path trigger and make the global-history arm conditional on no explicit alternate flag, while retaining the explicit historical flag.

common/national_focus/05_israel.txt
```diff
@@ -5559,6 +5909,8 @@ focus_tree = {
 				add = 5
 				has_active_mission = ISR_isr_jewish_home_coalition_ticker_mission
 			}
+			modifier = { factor = 25 has_global_flag = ISR_SECULAR_CENTRE_FOCUS_PATH }
```
Blazer135: [P2] This and every focus in the Jewish Home spine are weighted exclusively for SECULAR_CENTRE and zeroed for other explicit paths, even though the spine requires western_autocrats_are_in_power_or_coalition and the rule description presents this party spine as path-neutral. The same ownership error starts on the Yisrael Beiteinu spine at line 5352. Remove these Secular Centre owner/kill modifiers, or provide explicit path options that intentionally own these spines.

Blazer135 (changes requested):

common/scripted_triggers/99_ISR_scripted_triggers.txt
TemplarGeneral: Fixed in 997672fd40. Added `ISR_ai_alt_path` (the four non-historical flags) and guarded the global-history arm with it, matching `FRA_ai_historical_path` / `GEO_ai_georgian_dream_path`:

```
ISR_ai_historical_path = {
	OR = {
		has_global_flag = ISR_HISTORICAL_FOCUS_PATH
		AND = {
			is_historical_focus_on = yes
			NOT = { ISR_ai_alt_path = yes }
		}
	}
}
```

The explicit historical flag is retained, and every `ISR_ai_not_*_path` derives from these so nothing else in the file changed.

Your example now resolves the right way — evaluated with the report tools own evaluator:

| option | hist | ISR_blockade_gaza | ISR_propose_hudna |
| --- | --- | ---: | ---: |
| HISTORICAL | on/off | 75 | 0 |
| PEACE_PROCESS | on | **0** | **25** |
| PEACE_PROCESS | off | 0 | 75 |
| GREATER_ISRAEL | on/off | 75 | 0 |
| NO_PATH | on | 75 | 0 |
| NO_PATH | off | 3 | 3 |

Root cause worth naming: the unguarded arm `write.md` mandates is safe only inside `ai_will_do`, where the mandated `factor = 0 <not trigger>` comes last and that not trigger is a pure alt-flag OR. ISR ORed the historical trigger into four aliases whose not triggers open with `NOT = { <alias> }`, so the alias being true suppressed its own killswitch. Ive added that caveat to `.claude/skills/country-ai-path/references/write.md` §4, and `ai_path_report.py` now reports the shape. It flags IRQ as carrying the identical bug on 88 focuses via `IRQ_ai_national_accord_path` — separate PR.

common/decisions/categories/israel_categories.txt
TemplarGeneral: Fixed in 997672fd40.

`ISR_ai_path_category` and both Likud decisions now gate on the normalized trigger instead of raw flags:

```
	visible = {
		OR = {
			ISR_ai_likud_path = yes
			has_global_flag = ISR_SECULAR_CENTRE_FOCUS_PATH
			has_global_flag = ISR_GREATER_ISRAEL_FOCUS_PATH
		}
	}
```

and `visible = { ISR_ai_likud_path = yes }` on `ISR_rally_the_likud_base` / `ISR_likud_forms_the_government`. `ISR_ai_likud_path` is `ISR_ai_historical_path` OR the LOOK_EAST flag, and `ISR_ai_historical_path` is now guarded against `ISR_ai_alt_path` (your other P1), so explicit alternate rules override global history exactly as you asked - under PEACE_PROCESS or SECULAR_CENTRE the Likud ramp is hidden, under NO_PATH with historical AI on it is visible and Israel can actually reach Likud.

The Secular Centre and Kahanist ramp pairs keep their raw flag gates: those options always set a flag, so the trigger buys nothing there.

I did not add the `TAG_ai_historical_path = yes` + `TAG_ai_not_historical_path = no` pair from `write.md` §8 - `ISR_ai_not_likud_path` opens with `NOT = { ISR_ai_likud_path = yes }`, so the second half is tautological once the arm is guarded.

`ai_path_report.py` now checks decision and category gates in both directions - dead under NO_PATH + historical AI, or visible during an alt path. It flags `GER_ai_path_category` / `GRE_ai_path_category` for the raw-flag version of this bug and `EGY_ai_rally_the_generals` / `IND_rally_the_awakening` for the unguarded-trigger version. Separate PR.

common/national_focus/05_israel.txt
TemplarGeneral: Fixed in 997672fd40 - both spines are path-neutral now.

Removed the `SECULAR_CENTRE` owner/kill pair from 31 focuses, done through `apply_ai_path_weights.py` with a `-` map so the ticker-mission `add`s and bankruptcy guards stayed put (62 path modifiers removed, nothing else touched):

- **Jewish Home**, 20 focuses (`western_autocrats_are_in_power(_or_coalition)`): `ISR_jewish_home_path`, `ISR_get_five`, `ISR_dont_apologize`, `ISR_kaminitz_law`, `ISR_associations_law`, `ISR_conservative_judges`, `ISR_nation_state`, `ISR_bennet_rule`, `ISR_promote_trickle_down`, `ISR_machette_bureaucracy`, `ISR_limit_the_labor`, `ISR_a_new_singapore`, `ISR_tax_reductions`, `ISR_crush_the_shelters`, `ISR_shaked_rule`, `ISR_trust_busting`, `ISR_israels_thatcher`, `ISR_over_legislation`, `ISR_promote_national_service`, `ISR_public_works`
- **Yisrael Beiteinu**, 11 focuses (`neutrality_oligarchism_in_power_or_coalition`), starting at the line you pointed at: `ISR_russian_party_path` through `ISR_national_secular_government`

You were right that the rule description and the diff disagreed - the description was written from the plan, not from the diff. That is on me.

`ISR_democratic_jewish_balance` stays `SECULAR_CENTRE`-owned: it is that path's trunk. Neutralising the spines does not strand them, since both hang off the three-way OR-mesh (`ISR_jewish_first` / `ISR_democratic_jewish_balance` / `ISR_democratic_first`) and at least one trunk is live in every rule state; their own `available` party gates keep the AI out under Likud or Labour. `SECULAR_CENTRE` ownership drops 78 -> 47 focuses and the report shows 0 orphans and 0 stranded for that option.

The report's ownership section now also prints the ruling parties each group's focuses require, which is what would have made this visible at a glance.

Blazer135 (approved):

PR #3479, opened 2026-09-03
Build slots 72

JohnsonvilleF221:
### Changes

Opening this so there's something concrete to look at.

Building slots for high-population states: categories state_09 to state_18 now ramp
from 30 to 85 base slots instead of 25 to 50, and MAX_SHARED_SLOTS goes 56 -> 100 so
railway and infrastructure modifiers still have headroom above the base.

Categories below state_09 (under 10.5m population) are unchanged. 146 of 1250 states
affected (at game start); total base slots across them 4999 -> 8035, about +22% of the 13764 world
total. Slots are a cap, not buildings.

Performance seems fine on my PC: i7-12700k, 32gb ram. I have play tested a version of this for over 2
months and it only slightly lowered performance on the last patch vs stock MD.

At the raised cap the state view grid overflows its panel, so:
- grid refitted to 12 columns at 29x25
- slot entry sprites scaled to the cell (building_status_overlay was drawing 46x46
  unscaled and overlapping neighbouring cells)
- GFX_damage_bar_slim added alongside GFX_damage_bar, not replacing it
- Renewable Energy Hotspot readout shrunk and moved to clear the bottom row

Commits are split so the ramp, the cap and the UI can be taken separately.

Changelog updated under Balance and Graphics.


Blazer135:
@JohnsonvilleF221 so after talking with the team, we think the 100 Slots is to much and probably will not benefit all the nations except the Asian nations with the high pop.

We agreed on the number to be 72 for now, while we look at more solutions for after 2.0 release 

please add the Fixes and Tag me, i will re-review it again 

Also Can you Run a Test and see what is the difference in preformance

JohnsonvilleF221:
@Blazer135, I have changed the slot cap to 72.

I reworked the ramp to keep the feel of the stock MD rather than replacing it like my submod does. Every category is stock × 1.2333, rounded. This works out to state_01 = 4 through state_18 = 62, while state_00 rounds back to 1, so it remains unchanged.

As for performance, I ran a 30-minute test and logged the results. The test was conducted as Nauru at 5 speed from the year 2000 on both a fully stock version and with the mod. When comparing a fixed time window such as from 2003.01.01 to 2004.12.13, both versions tested at exactly 58.2 days per minute.

I also compared the industry ledger at 2004.12 across 19 different countries in both runs: 1949 vs 1926 total factories (1.19%), and 1422 vs 1414 civilian (0.56%). Individual nations moved by up to ±19 from normal AI divergence, which is far larger than the gap between the two runs.

I believe there is no performance impact due to the fact slots are an integer cap on the state, not an allocation, an empty slot is iterated by nothing. Tick cost tracks buildings actually built, and how many get built is set by construction capacity, not slot availability.

Caveat: There was one run per version over a five-year period. This shows that the cap's effect is below the run-to-run noise within that window, rather than demonstrating that the effect is zero in a 30-year game. If you would like me to test a longer run, say two hours, I can conduct that test tonight.

Blazer135 (approved):

Blazer135:
@JohnsonvilleF221 good job, Approved and Merged this

PR #3483, opened 2026-09-03
Give Japan a five-path country AI path game rule (#3162)

TemplarGeneral:
Closes part of #3162 (Japan).

Japan had a two-option rule (`JAP_HISTORICAL` / `JAP_RETURN_OF_THE_EMPIRE`) whose flags nothing in the focus tree read. `ai_path_report.py --tag JAP` reported **0 owned / 365 path-neutral**, **63 orphaned focuses** under historical AI, and three coherent spines that no option claimed:

- the **left / social-democratic spine** (`JAP_left`, `JAP_information_disclosure`, `JAP_management_documents` plus ~63 descendants), zeroed at the roots by a killswitch and by both strategy plans while its descendants kept positive weight;
- the **pacifist / Article 9 tail** (`JAP_leading_disarmament` to `JAP_principle_of_no_armed_forces` to `JAP_abrogating_us_japan_security_treaty`), zeroed by both plans;
- the **remilitarisation spine** (32 focuses behind `power_balance_value > 0.70`), which the Article Nine decision ladder caps the historical AI out of at 0.55, so only the Empire path could ever revise Article 9.

## The option set

| `name =` | text | owns |
|---|---|---|
| `HISTORICAL` | Historical | LDP, US alliance, Article Nine stands |
| `NORMAL_NATION` | A Normal Nation | Article Nine rewritten, SDF becomes a standing military, still democratic and US-aligned |
| `SOCIAL_DEMOCRACY` | The Second Postwar | disclosure, labour law, redistribution, open immigration |
| `PACIFIST` | The Peace Constitution | arms-export ban, no armed forces, treaty abrogation |
| `RETURN_OF_THE_EMPIRE` | The Return of the Empire | imperial restoration and the conquest tail |
| `RANDOM_PATH` | shared key | `random_list` over all five |
| `NO_PATH` | shared key | `default = { }`, last, sets nothing |

Each path's branch root is reachable by something its AI can satisfy. HISTORICAL and NORMAL_NATION start from the LDP already in power (NORMAL_NATION's gate is the Article Nine balance, opened below); SOCIAL_DEMOCRACY and PACIFIST get AI-only ramps; RETURN_OF_THE_EMPIRE keeps its existing monarchist ramp.

## Changes

**Rule and loc.** Option names unprefixed, `RANDOM` becomes `RANDOM_PATH` on the shared `RULE_OPTION_MD_*` keys, `NO_PATH` kept as the `default = { }` block last. `RULE_OPTION_JAP_HISTORICAL` now reads literally `"Historical"` (it was `"The Postwar Consensus"`, which moved into the desc). Every `_desc` is two sentences with the party names highlighted. The `@JAP Japan` header key already existed and is unchanged. The file is **not** reordered, since that is a separate cross-cutting item.

**Flags.** `JAP_HISTORICAL_AI_FOCUS` to `JAP_HISTORICAL_FOCUS_PATH`, `JAP_RETURN_OF_THE_EMPIRE_AI_FOCUS` to `JAP_RETURN_OF_THE_EMPIRE_FOCUS_PATH`, plus `JAP_NORMAL_NATION_FOCUS_PATH`, `JAP_SOCIAL_DEMOCRACY_FOCUS_PATH` and `JAP_PACIFIST_FOCUS_PATH`. Roughly 300 references across decisions, events, the tree, `ai_strategy` and the plans.

**Tree ownership.** 15 scripted path triggers and 211 focuses re-owned across eight groups (`historical`, `conservative`, `rearmament`, `us_alliance`, `autonomy`, `left`, `pacifist`, `empire`), written by `apply_ai_path_weights.py`. Ownership is assigned by single-group reachability from each spine's seeds: any focus reachable from two paths' seeds stays path-neutral, which is what keeps the shared electronics chain and the middle side of every three-way mutex alive in every state. 154 focuses stay path-neutral, including the economy, army and equipment trunk.

**Party ramps** in `JAP_ai_path_category` (`is_ai = yes`, no loc): the DPJ (index 2, inside the incumbent's own group, so `party_popularity_increase` only) and the SDP (index 18, a different group from the incumbent, so both temp vars, mirroring the existing monarchist ramp). Neither disables elections; the takeover decision re-enables every 180 days instead.

**Article Nine ladder.** The 20 rearmament decisions now push for `JAP_ai_rearmament_path` (NORMAL_NATION plus EMPIRE) rather than the empire flag alone, the 0.55 cap no longer applies to a rearmament path, and PACIFIST zeroes them outright. The capped historical ladder now also covers SOCIAL_DEMOCRACY and `NO_PATH` with historical AI on. The reactor-restart nudges follow `JAP_ai_conservative_path` and are zeroed for the left paths, matching `JAP_nuclear_phaseout`'s ownership in the tree.

**Plans.** Three thin path plans plus `JAP_unscripted_plan` for the `NO_PATH` with historical-AI-off state, which previously enabled **no plan at all**. Both existing plans keep their `ai_national_focuses` lists; their `focus_factors` collapse to `JAP_dissolving_sdf = 0`, since the tree now owns the killswitching and the old lists had begun to contradict it.

**AI hardening.** `JAP_no_war_if_at_war` switches off permanently once `JAP_imperial_war_cabinet` completes, after which the empire chain keeps pushing `antagonize` and `prepare_for_war` with no check on whether Japan is losing. The eight `JAP_empire_*` pre-war blocks now gate on `NOT = { surrender_progress > 0.15 }`, which lifts that pressure for every target the chain can reach rather than for a hand-maintained list; the mod-wide `MD_avoid_new_wars_while_losing` already supplies the targetless declaration penalty on the same threshold, so no per-tag brake is added. The `JAP_empire_blitz_*` blocks are untouched — they gate on `has_war_with` and steer an existing war rather than start a new one. `JAP_historical_no_offensive_wars` enables on the `JAP_ai_historical_path` / `JAP_ai_not_historical_path` pair rather than the raw `JAP_HISTORICAL_FOCUS_PATH` flag, so the default `NO_PATH` plus historical AI state keeps its offensive-war suppression; the seven `Japan_economy` option `ai_chance` nudges that read the same raw flag take the pair too. `JAP_electronics_superpower` gets a real weighting modifier for the `JAP_model_competition` malus it cures. `Japan.101` keys its `ai_chance` on `JAP_ai_historical_path` rather than `is_historical_focus_on`.

**Historical government walker.** `JAP_ai.4` (hidden, `is_ai = yes`, no loc) asserts both the ruling party and the roster index at thirteen dates, scheduled from `00_yearly_effects.txt`. Japan sets no `term_limit`, so nothing rotated its Prime Minister. The walker covers the 2009-2012 DPJ interregnum, which is why explicit index assertion matters: conservatism index 6 is Sadakazu Tanigaki, LDP president but never Prime Minister, so the walker steps 5 (Aso) to 7 (Abe's return) while the ruling party sits on liberalism. The five flavour events that installed Prime Ministers with inline `create_country_leader` (`japan_classic.16`, `JAP_historical.20010426`, `.20060926`, `.20070926`, `.20080924`) now assert the roster index and call `set_ruling_leader` / `set_leader` instead. The roster entries are byte-identical to the data they inlined, so nothing player-visible changes, and the pointer stops desyncing.

**Tooling.** `ai_path_report.py` could not read Japan's roster shape: `_parse_roster_branch` only looked at the top level of a sub-ideology branch and only at `if`, so Japan's date-wrapped, `else_if`-chained roster parsed as zero entries and every walker branch was reported OUT OF RANGE. It now recurses and accepts `else_if`. `parse_year_schedule` also reads `MD_event_on_startup_events` as the mod start year, so events dispatched at startup are no longer reported as unscheduled. One regression test added; `python -m pytest` is green (4135 passed, 22 skipped).

## Verification

- `ai_path_report.py --tag JAP`: rule and wiring **clean**; **0 orphans and 0 stranded focuses in all twelve states**; 0 mutex ties, down from 3 intra-path ties broken with boost overrides; no additive path modifiers; no unreferenced flags; every one of the 15 triggers has callers.
- `government` section: every scheduled date resolves to the person history had (Koizumi, Abe, Fukuda, Aso, Hatoyama, Kan, Noda, Abe, Suga, Kishida, Ishiba, Takaichi). No `change_leader_temp`, no unbounded party change. Two notes remain by design. The tool's blanket "walker over an undated roster" advice stands because Japan's roster carries no per-entry end-of-tenure dates, but it **is** a historically ordered sequence, which is exactly what the explicit index assertion relies on. The final `else` fallback branch is unscheduled, which the bounded-chain shape requires.
- `mechanics` section: no AI-untakeable cure. Every remaining flag on the section attaches to the two cost-shaped keys **seeded negative** in the history file, `JAP_transparency_and_audit_reform_corruption_cost_factor` and `JAP_regulatory_reform_bureaucracy_cost_multiplier_modifier` — bonus ramps, not burdens, so their "every cure is dead" states and the three flat-base cure lines (`JAP_independent_audits`, `JAP_eliminating_amakudari`, `JAP_disclosure_of_political_funding`) are by design and no AI cure route was added for them. The `nothing relieves` line lists four entries and all four are judged **not** burdens: `JAP_ykk_group` is +2.5% civilian factory productivity plus research bonuses, and `JAP_political_modifier`, `JAP_economy_modifier` and `JAP_jsw_steel_modifier` are added in history with their backing variables unset, so they apply nothing until a focus fills them.
- Validators run against `3cb87fe`: `validate_focus_tree.py --path .` clean; `validate_decisions.py` 290 errors / 459 warnings, identical to the pre-change baseline; `validate_common_mistakes.py`, `validate_localisation.py`, `validate_variables.py`, `validate_simplifications.py` and `check_common_mistakes.py` clean; `validate_events.py` zero warnings for `05_japan.txt`. The review-fix commit on top only swaps trigger references inside `enable` and `ai_chance` blocks and deletes one strategy block; CI's validation report covers it.

Stale checklist annotations corrected while here: nothing outside `00_game_rules.txt` and `999_game_rules_on_actions.txt` reads `JAP_ai_behavior` (`events/05_japan.txt` does not), and no `set_country_flag` path flag remained.

### DIFF
--- a/common/ai_strategy/JAP.txt
+++ b/common/ai_strategy/JAP.txt
@@ -88,7 +88,7 @@ JAP_no_war_if_at_war = {
 JAP_historical_no_offensive_wars = {
 	allowed = { original_tag = JAP }
 	enable = {
-		has_global_flag = JAP_HISTORICAL_AI_FOCUS
+		has_global_flag = JAP_HISTORICAL_FOCUS_PATH
 		has_war = no
 	}
 	abort_when_not_enabled = yes
@@ -178,7 +178,7 @@ JAP_empire_naval_breakdown = {
 JAP_empire_naval_arms_race_vs_CHI = {
 	allowed = { original_tag = JAP }
 	enable = {
-		has_global_flag = JAP_RETURN_OF_THE_EMPIRE_AI_FOCUS
+		has_global_flag = JAP_RETURN_OF_THE_EMPIRE_FOCUS_PATH
 		country_exists = CHI
 		naval_strength_ratio = { tag = CHI ratio < 1.2 }
 	}
@@ -191,7 +191,7 @@ JAP_empire_naval_arms_race_vs_CHI = {
 JAP_empire_naval_arms_race_vs_USA = {
 	allowed = { original_tag = JAP }
 	enable = {
-		has_global_flag = JAP_RETURN_OF_THE_EMPIRE_AI_FOCUS
+		has_global_flag = JAP_RETURN_OF_THE_EMPIRE_FOCUS_PATH
 		country_exists = USA
 		naval_strength_ratio = { tag = CHI ratio > 1.2 }
 		naval_strength_ratio = { tag = USA ratio < 1.5 }
@@ -205,7 +205,7 @@ JAP_empire_naval_arms_race_vs_USA = {
 JAP_empire_industrial_base = {
 	allowed = { original_tag = JAP }
 	enable = {
-		has_global_flag = JAP_RETURN_OF_THE_EMPIRE_AI_FOCUS
+		has_global_flag = JAP_RETURN_OF_THE_EMPIRE_FOCUS_PATH
 		has_completed_focus = JAP_formalize_national_defense
 	}
 	abort_when_not_enabled = yes
@@ -462,4 +462,30 @@ JAP_empire_blitz_china = {
 
 	ai_strategy = { type = invade id = "CHI" value = 400 }
 	ai_strategy = { type = conquer id = "CHI" value = 400 }
+}
+
+JAP_hold_off_new_wars_while_losing = {
+	allowed = { original_tag = JAP }
+	enable = {
+		has_war = yes
+		surrender_progress > 0.15
+	}
+	abort_when_not_enabled = yes
+
+	ai_strategy = { type = declare_war id = "CHI" value = -4000 }
+	ai_strategy = { type = declare_war id = "NKO" value = -4000 }
+	ai_strategy = { type = declare_war id = "KOR" value = -4000 }
+	ai_strategy = { type = declare_war id = "TAI" value = -4000 }
+	ai_strategy = { type = declare_war id = "AST" value = -4000 }
+	ai_strategy = { type = declare_war id = "NZL" value = -4000 }
+	ai_strategy = { type = declare_war id = "PHI" value = -4000 }
+	ai_strategy = { type = declare_war id = "MAY" value = -4000 }
+	ai_strategy = { type = declare_war id = "IND" value = -4000 }
+	ai_strategy = { type = declare_war id = "VIE" value = -4000 }
+	ai_strategy = { type = declare_war id = "CBD" value = -4000 }
+	ai_strategy = { type = declare_war id = "LAO" value = -4000 }
+	ai_strategy = { type = declare_war id = "BRM" value = -4000 }
+	ai_strategy = { type = declare_war id = "SOV" value = -4000 }
+	ai_strategy = { type = declare_war id = "RAJ" value = -4000 }
+	ai_strategy = { type = declare_war id = "SIN" value = -4000 }
 }
--- a/common/ai_strategy/JAP.txt
+++ b/common/ai_strategy/JAP.txt
@@ -88,7 +88,8 @@ JAP_no_war_if_at_war = {
 JAP_historical_no_offensive_wars = {
 	allowed = { original_tag = JAP }
 	enable = {
-		has_global_flag = JAP_HISTORICAL_FOCUS_PATH
+		JAP_ai_historical_path = yes
+		JAP_ai_not_historical_path = no
 		has_war = no
 	}
 	abort_when_not_enabled = yes
@@ -219,6 +220,7 @@ JAP_empire_indochina = {
 	allowed = { original_tag = JAP }
 	enable = {
 		has_completed_focus = JAP_reclaim_indochina
+		NOT = { surrender_progress > 0.15 }
 	}
 	abort_when_not_enabled = yes
 
@@ -237,6 +239,7 @@ JAP_empire_myanmar = {
 	allowed = { original_tag = JAP }
 	enable = {
 		has_completed_focus = JAP_reclaim_myanmar
+		NOT = { surrender_progress > 0.15 }
 	}
 	abort_when_not_enabled = yes
 
@@ -253,6 +256,7 @@ JAP_empire_pacific_islands = {
 	allowed = { original_tag = JAP }
 	enable = {
 		has_completed_focus = JAP_reclaim_the_pacific
+		NOT = { surrender_progress > 0.15 }
 	}
 	abort_when_not_enabled = yes
 
@@ -275,6 +279,7 @@ JAP_empire_south_china_sea = {
 	allowed = { original_tag = JAP }
 	enable = {
 		has_completed_focus = JAP_reclaim_south_chinese_sea
+		NOT = { surrender_progress > 0.15 }
 	}
 	abort_when_not_enabled = yes
 
@@ -295,6 +300,7 @@ JAP_empire_korea = {
 			has_completed_focus = JAP_reclaim_north_korea
 			has_completed_focus = JAP_reclaim_south_korea
 		}
+		NOT = { surrender_progress > 0.15 }
 	}
 	abort_when_not_enabled = yes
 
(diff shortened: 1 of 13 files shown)


common/ai_strategy/JAP.txt
```diff
@@ -88,7 +88,7 @@ JAP_no_war_if_at_war = {
 JAP_historical_no_offensive_wars = {
 	allowed = { original_tag = JAP }
 	enable = {
-		has_global_flag = JAP_HISTORICAL_AI_FOCUS
+		has_global_flag = JAP_HISTORICAL_FOCUS_PATH
```
Blazer135: [P1] Apply the historical war brake to implicit history

This still keys the no-war strategy only to the explicit `HISTORICAL` flag. With the default `NO_PATH` rule and global Historical AI enabled, the new plan logic treats Japan as historical (`JAP_ai_historical_path = yes` with no alternate path), but this flag is never set, so the historical plan runs without its offensive-war suppression. Please use the same historical/no-alternate trigger pair as `JAP_HISTORICAL_plan` so the default historical state gets the guard.

common/ai_strategy/JAP.txt
```diff
@@ -463,3 +463,29 @@ JAP_empire_blitz_china = {
...
+	ai_strategy = { type = declare_war id = "PHI" value = -4000 }
+	ai_strategy = { type = declare_war id = "MAY" value = -4000 }
+	ai_strategy = { type = declare_war id = "IND" value = -4000 }
+	ai_strategy = { type = declare_war id = "VIE" value = -4000 }
+	ai_strategy = { type = declare_war id = "CBD" value = -4000 }
+	ai_strategy = { type = declare_war id = "LAO" value = -4000 }
+	ai_strategy = { type = declare_war id = "BRM" value = -4000 }
+	ai_strategy = { type = declare_war id = "SOV" value = -4000 }
+	ai_strategy = { type = declare_war id = "RAJ" value = -4000 }
+	ai_strategy = { type = declare_war id = "SIN" value = -4000 }
```
Blazer135: [P1] Cover every post-cabinet war target

This brake omits targets against which the empire path can acquire wargoals after `JAP_imperial_war_cabinet` disables `JAP_no_war_if_at_war`. For example, the Myanmar chain can target KAC/WAA/SHN/KAR, while the Pacific chain can target PAU/MIC/NAU/KIR/MAR/TUL/SOL/PAP/VAN/FIJ/SAM/TON and USA/FRA/ENG. At more than 15% surrender, those unlisted targets remain eligible for a new declaration, defeating this safeguard. Please cover the complete target set or gate declarations generically.

Blazer135 (changes requested):
Two blocking AI-war safeguards need correction: the default historical state loses its offensive-war suppression, and the new losing-war brake omits reachable targets.

common/ai_strategy/JAP.txt
TemplarGeneral: Fixed in 9118637. `enable` now reads `JAP_ai_historical_path = yes` + `JAP_ai_not_historical_path = no`, the same pair as `JAP_HISTORICAL_plan`, so the default `NO_PATH` + historical AI state gets the guard.

common/ai_strategy/JAP.txt
TemplarGeneral: Right, and the block should not have existed: `MD_avoid_new_wars_while_losing` in `common/ai_strategy/MD_war_declaration_ai.txt` already has the byte-identical gate (`has_war = yes` + `surrender_progress > 0.15`), no `allowed` block, and a targetless `avoid_starting_wars = -200`. Any target list I write is a strict subset of it.

Removed it in 9118637 and gated the pressure instead: the eight `JAP_empire_*` pre-war blocks now carry `NOT = { surrender_progress > 0.15 }`, so `antagonize` / `prepare_for_war` lift for every target the empire chain can reach — KAC/WAA/SHN/KAR and the whole Pacific list included — with the mod-wide brake supplying the declaration penalty. The `JAP_empire_blitz_*` blocks are untouched since they gate on `has_war_with` and steer an existing war rather than start a new one.

Blazer135 (approved):

PR #3487, opened 2026-09-03
Give Libya a country AI path game rule (#3162)

TemplarGeneral:
Closes part of #3162 (Libya).

The checklist called `LBA_ai_behavior` "never read by any loaded script". It is, but by the wrong name: the dispatcher at `999_game_rules_on_actions.txt:4722` queried `rule = LBA_AI_BEHAVIOR` while the rule id is `LBA_ai_behavior`. HOI4 identifiers are case-sensitive, so no branch ever matched, no flag was ever set, and all seven `enable` blocks in `LBA_strategy_plans.txt` were permanently false. Libya's entire scripted-path layer — seven hand-written `ai_national_focuses` lists covering 151 focuses — had never run once.

`ai_path_report.py --tag LBA` reported **0 owned / 366 path-neutral**, **100 orphaned focuses** under historical AI, and no `focus_factors` on any plan. The only steering that existed was 40 blanket `factor = 0 is_historical_focus_on = yes` killswitches and two 25%-chance country flags (`LBA_AI_go_salafist_kingdom`, `LBA_AI_go_salafist_caliphate`), so which Libya an AI game produced was decided by dice rather than by the player's rule.

## The option set

| `name =` | text | owns |
|---|---|---|
| `HISTORICAL` | Historical | crush the LIFG, renounce the weapons programme, buy out of sanctions, then the post-Gaddafi LNA |
| `ISLAMIC_CALIPHATE` | Islamic Caliphate | tolerate the LIFG, Gaddafi assassinated, Salafist state, Sahel jihad |
| `GRAND_MUFTI` | Rule of the Grand Mufti | religious law above the state, the Mufti takes power as a Salafist kingdom |
| `LIBYAN_REPUBLIC` | Libyan Republic | Green Book revised, parties legalised, presidential republic |
| `KINGDOM` | The Senussi Kingdom | royals recalled and compensated, monarchy restored |
| `ANTI_COLONIAL_UNION` | Anti-Colonial Union | Gaddafi stays, arms insurgencies across Africa and the Middle East |
| `UNITED_ARAB_REPUBLIC` | United Arab Republic | Gaddafi stays, Ba'athist union from Mauritania to Yemen |
| `UNITED_STATES_OF_AFRICA` | United States of Africa | Gaddafi stays, continental union |
| `RANDOM_PATH` | shared key | `random_list` over all eight, historical bucket included |
| `NO_PATH` | shared key | `default = { }`, last, sets nothing |

The three union options stay separate even though their plan lists overlap almost completely, because each ends in a different formable. They share the `LBA_gaddafis_dream` → `LBA_regional_politics` body through one alias and differ only on their mutex arm.

Libya's tree is diamond-shaped: the branch identities are 1–8 focus caps sitting on large shared bodies, so ownership is carried by six aliases (`islamist`, `gaddafi_union`, `gaddafi_regime`, `secular`, `reformist`, `rogue`) rather than by duplicating spines per path.

## Changes

**Rule and loc.** `DEFAULT` removed, `HISTORICAL` added first, `RANDOM` becomes `RANDOM_PATH` on the shared `RULE_OPTION_MD_*` keys, `NO_PATH` kept as the `default = { }` block last. `RULE_OPTION_LBA_HISTORICAL` reads literally `"Historical"`. Every `_desc` is two sentences, present tense, no hard dates, with party and movement names highlighted. The `@LBA Libya` header key already existed and is unchanged. The file is **not** reordered — that is a separate cross-cutting item.

**Wiring.** The case bug fixed and the block rewritten to eight `LBA_<PATH>_FOCUS_PATH` global flags plus a `RANDOM_PATH` `random_list` that includes the historical bucket. The old flags were bare option names (`LBA_GADDAFI_UAR`), now `_FOCUS_PATH`-suffixed. `NO_PATH` gets no branch.

**Tree ownership.** 23 scripted path triggers, up from the two Gaddafi-identity ones the file had, and **159 focuses re-owned** across fifteen groups by `apply_ai_path_weights.py`. 207 focuses stay path-neutral: the economy, rail, GMMR, army, air and navy trunk, plus the six post-Gaddafi ideology wings that no path claims. `LBA_ai_historical_path` uses the alt-guarded form because it is read outside `ai_will_do` — plan `enable`, decision `visible` and an event `ai_chance`.

**The 40 legacy killswitches.** 19 sat on focuses a path now owns and were replaced by that path's not-trigger; leaving them would have zeroed an explicit alt pick whenever historical AI was on. The other 21 sat on trunk content — the rail spine, the Great Man-Made River, the NOC, WTO accession — where a blanket historical killswitch stranded whole buildable sub-trees; those are stripped, since the path system now does the steering the killswitch used to approximate.

**Determinism.** The two 25%-chance country flags now follow the rule: the path flag sets them outright, an explicit rival path never sets them, and the dice survive only for a genuinely unscripted AI. The same shape covers `LBA_AI_good_boy` / `LBA_AI_bad_boy` in the sanctions line (18 focuses), which previously rolled 10/90 and only when historical AI was off — historical and the reformist paths now take the disarmament line, the caliphate and the unions take the rogue line.

**Reachability.** `LBA_grand_mufti_rule` gates on `fascism > 0.30`. Libya starts at 0.09 and the only fascism-side gains in the tree are `LBA_peace_with_lifg` (+0.10, index 11, caliphate-only) and `LBA_appoint_grand_mufti` (+0.10, index 10) — a ceiling of 0.19 for this path, so it was unreachable. `LBA_ai_path_category` (`is_ai = yes`, no loc, bare `check_variable`) adds one ramp on index 10, capped at 0.35 and disabled once the takeover focus is done; the takeover itself stays with the focus rather than a decision, so the path is not bypassed. The LIFG fork was also reassigned — the Islamist paths take `LBA_ignore_lifg`, the secular paths take `LBA_crush_lifg` — because sending the Grand Mufti through `crush_lifg` cost 10% fascism and capped the path at 0.10.

**Plans.** The seven enables repointed to the new flags, an eighth `LBA_libya_historical` plan added on the `LBA_ai_historical_path` / `LBA_ai_not_historical_path` pair with a 31-focus prerequisite-closed list, and `focus_factors` added to all eight zeroing the danger rewards each plan must not take.

**Events.** `LibyaCivilWar.15` is the pivot of the whole country — the mission `libya_gaddafi_family_anti_government_protests` fires it on `has_stability < 0.15` — and **neither option carried an `ai_chance`**, so it was a coin flip. Option `.a` keeps LBA as the Gaddafi loyalists and spawns the NTC separately; `.b` turns LBA itself into the NTC and is the only branch that leaves the tag able to take `LBA_end_of_gaddafi`. Historical now takes `.b` and the Gaddafi-survives paths take `.a`, with the sibling zeroed in each case rather than merely outweighed. `LibyaCivilWarContent.24` gets `ai_chance` too, zeroing the tribal-independence option for any path-driven AI. `libya_gaddafi_family_deploy_military` (`base = 5`), which triples the rebel side and calls in NATO, is zeroed for path AIs.

**AI strategy.** No per-tag brake added: `MD_avoid_new_wars_when_outmatched` and `MD_avoid_new_wars_while_losing` carry no `allowed` block, so they already cover Libya, and Libya's 31 `conquer` blocks are already gated on focus, claim and a no-active-neighbour-war check. The real gap was a pre-war readiness gate — `LBA` appeared **zero times** in `MD_war_declaration_ai.txt` — so one rivalry set was added there using that file's existing 6-block template: `LBA_prepare_aouzou_CHA` and its `delay_demo` / `ready_demo` / `ready_nat` / `ready_mid` / `failsafe` siblings, for the Aouzou Strip rivalry the tree pushes hardest. `common/ai_strategy/LBA.txt` is unchanged. 27 combat-capacity focuses in the `LBA_libyan_military` trunk gained `factor = 2 ai_is_threatened = yes`, placed before the path modifiers so the killswitches still win.

**Burdens.** `LBA_general_peoples_congress` had both its cure focuses on the Green Book trunk, which every non-republic path killswitches, and `cancel = { has_government = fascism }` does not fire for the historical LNA junta — `nationalist` is a separate engine ideology from `fascism` in MD. `LBA_end_of_gaddafi`, deliberately left path-neutral so any path can continue after Gaddafi falls, now clears it and `LBA_empowered_revolutionary_committees`.

## Verification

- `ai_path_report.py --tag LBA`: rule and wiring **clean**; `mechanics` **clean**; ownership 159 / 207; 0 path-relevant mutex ties, down from two that resolved by file order; no additive path modifiers; no unreferenced flags; every trigger has callers.
- State matrix: **0 orphans in 12 of the 18 states**, including both `NO_PATH` rows, against 100+ before. `NO_PATH` with historical AI off leaves all 366 focuses live and zero zeroed, so an unconfigured Libya runs unscripted as intended. The six regime-family states keep 1–2 residual orphans, `LBA_constitutional_monarch` and `LBA_call_back_exiles`. Both are pre-existing tree structure, not fallout from this change: each requires prerequisites from **two** separate `prerequisite` blocks that live on mutually-committed spines (`LBA_grassroots_democracy` **and** `LBA_compensation_to_royals`; `LBA_protect_toubou_and_tuaregs` **and** `LBA_approach_nato`), so no AI committed to a single path can satisfy both. They remain reachable for a player and under `NO_PATH`.
- `government` section unchanged: `LBA_political_leaders.txt` carries **zero** date markers and succession is `<SubIdeology>_leader` index-driven, so the verdict is an undated successor roster and **no walker was written**. Libya also starts `elections_allowed = no` with no drift idea, so nothing moves party popularity on its own in either direction.
- `nothing relieves` lists four entries, none of which is a burden needing an AI cure. `LBA_great_man_made_river_phase2` is cleared by `zz_libyan_special_projects.txt`. `gaddafi_family_instability_decay` is seeded **negative** in history (`-2`) and is recomputed as −0.25 per living, non-exiled, non-ruling family member, so it is a decay rate, not a malus; `libya_gaddafi_instability_dynamic_modifier` reads from it and trends to zero for an AI, which takes none of the destabilising family decisions. `LBA_stream_of_refugees` is removed nowhere in the mod — pre-existing, and out of scope here.
- The two remaining `zeroed by every plan` entries are deliberate. `LBA_constitutional_monarch` is mutex with both the republic and the kingdom targets and needs both spines; `LBA_establish_workers_republic` carries `set_politics`. Letting either through would derail the path the player picked.
- Validators run against the branch: `validate_focus_tree.py --path .` clean; `validate_decisions.py` 290 errors / 459 warnings in 69 files, byte-identical to the pre-change baseline, so the new `is_ai = yes` category added no `ai-only-decision-localisation`; `validate_common_mistakes.py` and `validate_localisation.py` clean; `validate_events.py` 12148 warnings, identical to `main`, so no event warning was added.


common/national_focus/05_libya.txt
```diff
@@ -13085,9 +13457,21 @@ focus_tree = {
 
 		bypass = { is_not_ruled_by_a_gaddafi = yes }
 
-		completion_reward = { log = "[GetDateText]: [Root.GetName]: Focus LBA_end_of_gaddafi" }
+		completion_reward = {
```
Blazer135: [P2] Run the post-Gaddafi cleanup on bypass

`available` and `bypass` use the same `is_not_ruled_by_a_gaddafi` condition. As soon as this focus becomes available it is auto-bypassed, so `completion_reward` never executes and these Gaddafi-era ideas remain active after the regime change. Move the cleanup to `bypass_effect` or to the transition event.

localisation/english/MD_game_rules_l_english.yml
```diff
@@ -1255,20 +1255,22 @@
...
+ RULE_OPTION_LBA_ISLAMIC_CALIPHATE: "Islamic Caliphate"
+ RULE_OPTION_LBA_ISLAMIC_CALIPHATE_DESC: "Gaddafi tolerates the §8Libyan Islamic Fighting Group§!, which assassinates him and establishes a Salafist state. Libya becomes a base for jihad across the Sahel, destabilising Chad, Niger, Nigeria, Algeria, Mali and Mauritania."
+ RULE_OPTION_LBA_GRAND_MUFTI: "Rule of the Grand Mufti"
+ RULE_OPTION_LBA_GRAND_MUFTI_DESC: "A §8Grand Mufti§! is appointed and religious law placed above the state. The Mufti eventually takes power, establishing a Salafist kingdom and imposing strict Islamic governance across Libya."
+ RULE_OPTION_LBA_LIBYAN_REPUBLIC: "Libyan Republic"
+ RULE_OPTION_LBA_LIBYAN_REPUBLIC_DESC: "Libya revises the Green Book, empowers the People's Congresses and legalises political parties. The state transforms into a presidential republic with competitive politics replacing Gaddafi's direct democracy."
+ RULE_OPTION_LBA_KINGDOM: "The Senussi Kingdom"
+ RULE_OPTION_LBA_KINGDOM_DESC: "The royal family returns home and is compensated, and the §8Senussi§! monarchy is restored. Libya abandons republicanism for a constitutional kingdom under the restored royal house."
```
Blazer135: [P2] Match the Kingdom description to its forced branch

The Kingdom plan explicitly prioritizes `LBA_the_real_monarch` and sets `LBA_constitutional_monarch = 0`; `LibyaFocus.25` then disables elections. This description promises a constitutional kingdom while the selected rule forces the opposite branch. Either route the plan through `LBA_constitutional_monarch` or describe the outcome as an absolute Senussi monarchy.

common/national_focus/05_libya.txt
TemplarGeneral: Fixed in 0d8ccc7: available and bypass were identical, so the focus always auto-bypassed and completion_reward's idea cleanup never ran. Moved the LBA_general_peoples_congress/LBA_empowered_revolutionary_committees removal into bypass_effect.

localisation/english/MD_game_rules_l_english.yml
TemplarGeneral: Fixed in 0d8ccc7: reworded the desc to match the scripted outcome. The plan forces LBA_the_real_monarch (LBA_constitutional_monarch is zeroed in focus_factors), and the event it fires (LibyaFocus.25) sets disable_elections = 1 on every option, so the Kingdom path installs an absolute Senussi monarch, not a constitutional one.

Blazer135 (approved):

PR #3496, opened 2026-09-03
Give Myanmar a country AI path game rule (#3162)

TemplarGeneral:
## Summary

Converts Myanmar's `BRM_ai_behavior` to the issue #3162 standard, and fixes the reason the rule
never actually worked.

**Before:** `DEFAULT` + an option literally named `RANDOM`, no `HISTORICAL`, a desc opening "Only
applicable to the 2000 start date", and a `RANDOM` list whose historical bucket set no flag at all.
More importantly `ai_path_report.py` saw **0 of 107 focuses owned** and both path flags "never read
in the tree" — every fork in `05_myanmar.txt` routed off the vanilla `is_historical_focus_on` toggle
instead, so the rule only nudged strategy-plan priority and never locked the AI onto the chosen path.
With historical AI off, the three political forks were independent 1v1 coin flips, so a run could
come out civilian-but-persecuting-and-federalising, or junta-but-tolerant.

**After:** `HISTORICAL`, `BEACON_OF_DEMOCRACY`, `AN_IRON_FIST`, `RANDOM_PATH`, `NO_PATH` (default).
Global-flag wiring including a new `BRM_HISTORICAL_FOCUS_PATH`, a `RANDOM_PATH` list that includes
the historical bucket (40/30/30), 65 of 107 focuses owned across five groups via the new
`common/scripted_triggers/99_BRM_scripted_triggers.txt`, a `BRM_no_scripted_path` fallback plan so
`NO_PATH` is never planless, and 24 `ai_is_threatened` weightings on the combat-capacity focuses (the
tree previously had none).

## Design

The three paths are derived from the tree's prerequisite graph and `available` gates, not from the
old option names. All 11 `mutually_exclusive` pairs were enumerated; four are political, the rest are
doctrine and great-power alignment.

| Option | Locks in |
|---|---|
| `HISTORICAL` | `pave_the_way_for_civilian_rule` → `accept_the_role_of_the_tatmadaw` + `the_path_of_intolerance`; stops short of both the mass expulsion and a federal state |
| `BEACON_OF_DEMOCRACY` | `pave` → `undermine_the_influence_of_the_tatmadaw` + `the_path_of_tolerance` + the autonomy/federation chain |
| `AN_IRON_FIST` | `reject_the_way_for_civilian_rule` → `establish_direct_military_rule` + the expulsion and insurgency-war chains |

The historical reading is confirmed by the tree's own weights rather than by the labels: under
historical AI the author had `pave` at ×1.5, `accept` at base 1 and `intolerance` at base 1 against 0
for each rival, and the old `AN_IRON_FIST` desc says the junta represses "instead of responding
positively to pressure to reform as it did historically."

`HISTORICAL` and `BEACON_OF_DEMOCRACY` share the `pave_the_way` spine (both are civilian rule), so
that spine is owned by a `BRM_ai_civilian_path` alias and killed only by the Iron Fist flag. They
diverge at the second fork — whether the civilian government tolerates the Tatmadaw's reserved role
(historical) or breaks it (Beacon) — and at the Rohingya fork, where Historical keeps the
intolerant *stance* but not the mass expulsion.

The three-way great-power alignment fork (US / ASEAN / China military partnership) is deliberately
**not** claimed by any option: all three sides sit at base 0 and resolve off live `influence_array`
via the existing `BRM_PRO_US` / `BRM_PRO_ASEAN` / `BRM_PRO_CHINA` plans. That is a second,
independent axis.

## The rule is authoritative over the alignment plans

Three pre-existing constructs could silently veto the path the player selected. All are removed or
given an escape hatch, because a rule option that the AI can ignore is not a rule option:

- `BRM_PRO_US` zeroed `the_path_of_intolerance`, `support_buddhist_nationalists` and
  `expel_the_rohingya` via `focus_factors`. Removed.
- `BRM_PRO_CHINA` zeroed `turn_on_the_united_wa_state_army`. Removed — that plan is active from turn
  one, since Myanmar starts at 87 Chinese influence.
- `crush_the_rebel_groups` and `turn_on_the_united_wa_state_army` are gated on
  `CHI = { influence_less_20 = yes }`, which Iron Fist has no way to satisfy (its own chain *deepens*
  Chinese ties). Both now read
  `available = { OR = { CHI = { influence_less_20 = yes }  BRM_ai_iron_fist_path = yes } }`.

Consequently `BRM_PRO_CHINA` and `BRM_PRO_US` now carry no `focus_factors`, which the report lists as
an advisory. That is intended: an alignment plan should express preference through its focus list,
not by zeroing focuses the game rule owns.

## Other fixes found while implementing

- **`base = 0` would have defeated the whole rule.** All 22 Iron Fist focuses and four others sat at
  `base = 0` and were lifted off it only by `add = 1 is_historical_focus_on = no`. A multiplicative
  path boost on base 0 is still 0, so picking An Iron Fist with historical AI on would have left
  `reject` at 0 against `pave` at 1.5 — the AI goes civilian regardless of the player's choice. The
  26 affected focuses were rebased to `base = 1` and their 27 legacy `is_historical_focus_on`
  modifiers removed (`apply_ai_path_weights.py` does not strip that shape — it only removes an
  `is_historical_focus_on` block that also carries `factor = 0`).
- **The Beacon plan fought its own rule.** `BRM_BEACON_OF_DEMOCRACY` also enabled on
  `influence_array^0 = USA`, and aborted on `NOT = { influence_array^0 = USA }` — so picking Beacon
  of Democracy gave you a plan that switched itself off in any game where the US did not lead
  influence. Now gated on the path trigger alone, abort dropped.
- **`focus_factors` added** to the Beacon and Iron Fist plans (neither had any), zeroing the rival
  fork roots so plan and tree agree.
- **`BRM_bolster_the_air_force` was `base = 0`** gated on historical-off while its sibling
  `BRM_bolster_the_navy` is `base = 5`, so a historical AI Myanmar never built an air force and
  `improve_our_air_doctrine` / `construct_air_bases` were permanently stranded. Set to `base = 5` to
  match the navy branch.

## Starting burdens, checked per rule state

`ai_path_report`'s burden scan only recognises ideas named `TAG_*` or `*_TAG`, so it reports one
burden for Myanmar and misses every real one. Cross-checking the history file's `add_ideas` block by
hand against cure focuses and their reachability per rule state:

| Burden | HISTORICAL | BEACON | IRON_FIST | NO_PATH |
|---|---|---|---|---|
| `Rohingya_Crisis` | AI repatriation decision | `reform_the_1982_citizenship_law` | `expel_the_rohingya` | decision / both focuses |
| `Western_Sanctions` | `cede_power_to_the_nld` | ✓ | `establish_direct_military_rule` | ✓ |
| `Ethnic_Insurgencies` | **uncured, deliberate** | `proclaim_an_end_to_the_insurgency` | same | uncured with historical AI on |
| `A_Political_Military` | **uncured, deliberate** | `end_the_role_of_the_tatmadaw_in_politics` | uncured, deliberate | uncured with historical AI on |

The two exemptions are stated rather than silent: Myanmar's ethnic insurgencies never ended, and the
Tatmadaw kept its constitutional political role — ending it is precisely what the Beacon path exists
to do. Carrying those two ideas is the historical outcome, not a stranded malus.

Because Historical no longer reaches `expel_the_rohingya`, `Rohingya_Crisis` would otherwise have had
no cure on that path, so the AI gets its own route: `BRM_ai_path_category`
(`allowed = { is_ai = yes  original_tag = BRM }`, no localisation keys) with
`BRM_ai_rohingya_repatriation_accord`, visible only when `NOT = { BRM_ai_alt_path = yes }`.

## Residual report findings, and why they stand

- **Orphans** — `become_a_major_non_nato_ally`, `join_the_shanghai_pact` and
  `push_for_mutual_defence_agreements_between_asean_states` are children of the influence-driven
  alignment fork, whose parents are base 0 plus conditions the static analyser cannot evaluate. Under
  `HISTORICAL` and `NO_PATH` + historical AI, `proclaim_an_end_to_the_insurgency` also shows, which
  is the `Ethnic_Insurgencies` exemption above.
- **One mutex tie** — `criticize_chinas_meddling` / `seek_closer_ties_to_china`, both Iron-Fist-owned
  at 25. It is a flavour choice inside one path's own spine, resolved at runtime by `BRM_PRO_CHINA`'s
  focus list. An earlier revision of this PR down-weighted one side purely to clear the warning; that
  is reverted, since it steered Iron Fist toward keeping Chinese influence high and so helped lock
  the insurgency gate above.

## Not changed, and why

- **`un_aid_subsidies_modifier`** is a shared UN-aid dynamic modifier added and removed by
  `remove_un_aid_subsidies_modifier_if_unused` in `01_international_systems_effects.txt`, not a
  Myanmar burden.
- **No historical government walker.** The report verdicts Myanmar's successor roster as undated
  (7 sub-ideologies, 0 dated entries), so a walker would install the wrong person.
- **No `common/ai_strategy/BRM.txt`.** The Iron Fist plan's antagonize stance and the Thai-border
  chain are already braked by the mod-wide `MD_avoid_new_wars_when_outmatched`
  (`enemies_strength_ratio > 0.75`); a per-tag brake would be a strict subset of it.
- **`BRM_launch_talks_on_a_new_constitution`** is the only "danger reward" flagged: it is a
  federation-vs-unitary cosmetic tag swap, not a disband or hand-away effect.

## Verification

- `ai_path_report.py --tag BRM`: rule/wiring clean, 65 owned, 0 additive path modifiers, 0
  unreferenced flags. Remaining findings are the four listed above, each justified rather than
  engineered away.
- Reachability walked over the *live* focus set per rule state (killswitches applied, then
  prerequisites resolved), confirming each path reaches its own spine and its own burden cures — in
  particular Iron Fist now reaches `crush_the_rebel_groups` → `internal_colonialism` →
  `proclaim_an_end_to_the_insurgency`.
- `validate_focus_tree.py --path .`: 0 warnings in `05_myanmar.txt` (the 15 reported are pre-existing
  Saudi Arabia ones).
- `validate_decisions.py`: 290 errors / 458 warnings, identical to the main baseline — the new
  AI-only category and decision add no findings, including no `ai-only-decision-localisation`.
- `validate_localisation.py`: no issues. `check_common_mistakes.py` on the touched files: passed.
- Every changed line in `05_myanmar.txt` is inside an `ai_will_do` or `available` block; `can_staff_`
  and `bankruptcy_incoming_collapse` guard counts unchanged, focus count unchanged (107).


common/ai_strategy_plans/BRM_strategy_plans.txt
```diff
@@ -5,14 +5,13 @@ BRM_BEACON_OF_DEMOCRACY = {
 
 	enable = {
 		original_tag = BRM
-		OR = {
-			has_global_flag = BRM_BEACON_OF_DEMOCRACY_FOCUS_PATH
-			check_variable = { influence_array^0 = USA }
-		}
+		BRM_ai_beacon_path = yes
```
Blazer135, line 8: **[P1] Keep `NO_PATH` from becoming planless.** With historical AI disabled, `NO_PATH` sets no branch flag, so neither of the two path plans can enable; the remaining alignment plans are conditional and may also all be inactive. The country-AI-path standard requires an unconditionally eligible strategy plan for this state. Add a safe fallback or allow weighted plan selection without setting a rule-path flag.

common/national_focus/05_myanmar.txt
```diff
@@ -507,7 +533,11 @@ focus_tree = {
 			recalculate_party = yes
 		}
 
-		ai_will_do = { base = 1 }
+		ai_will_do = {
+			base = 1
+			modifier = { factor = 25 BRM_ai_rival_of_beacon_path = yes }
```
Blazer135, line 538: **[P1] Prevent the Pro-US plan from overriding the historical fork.** The existing `BRM_PRO_US` plan sets `BRM_the_path_of_intolerance = 0`. If that plan activates after the NLD takes power, this new `factor = 25` is still multiplied by zero and historical Myanmar is forced into the tolerance branch, contrary to the selected rule and its description. Remove or gate the conflicting Pro-US `focus_factors` now that the game-rule paths own this fork.

common/national_focus/05_myanmar.txt
```diff
@@ -560,7 +588,11 @@ focus_tree = {
 			add_war_support = 0.05
 		}
 
-		ai_will_do = { base = 1 }
+		ai_will_do = {
+			base = 1
+			modifier = { factor = 25 BRM_ai_rival_of_beacon_path = yes }
```
Blazer135: **[P1] Do not escalate the historical path to expulsion.** `BRM_ai_rival_of_beacon_path` is true for both Historical and Iron Fist, so Historical boosts this focus and then `BRM_expel_the_rohingya` by 25. That removes 800,000 people and the Rohingya Crisis despite the historical rule promising that the situation remains unchanged. These two follow-up focuses should be owned by Iron Fist only and killed for the historical path.

common/national_focus/05_myanmar.txt
```diff
@@ -2401,7 +2510,11 @@ focus_tree = {
 			}
 		}
 
-		ai_will_do = { base = 1 }
+		ai_will_do = {
+			base = 1
+			modifier = { factor = 25 BRM_ai_iron_fist_path = yes }
```
Blazer135, line 2523: **[P1] Make the Iron Fist insurgency chain reachable.** Myanmar starts with 87 Chinese influence, which activates `BRM_PRO_CHINA`; that plan sets this focus to zero, and the focus itself also requires Chinese influence below 20. The Iron Fist route supplies no way to reduce that influence, so `BRM_internal_colonialism` and `BRM_proclaim_an_end_to_the_insurgency` are stranded even though this option promises to crush the insurgencies. The selected path must override both the availability gate and the conflicting Pro-China factor.

Blazer135 (commented):
Four inline findings. CI status intentionally excluded from this review.

common/ai_strategy_plans/BRM_strategy_plans.txt
TemplarGeneral: Fixed in b0bb1bf. Added `BRM_no_scripted_path` on the DEN/JAP scripted-trigger form — `enable = { original_tag = BRM  BRM_ai_no_scripted_path = yes }`, where `BRM_ai_no_scripted_path = { NOT = { BRM_ai_any_path = yes } }` is a pure flag complement, so it holds in both NO_PATH states rather than only when global historical AI happens to be on (the hole the ITA/LBA "historical plan doubles as fallback" trick leaves). `abort = { is_subject = yes }`, `weight = { factor = 1.0 }`, a path-neutral focus trunk, and `focus_factors` zeroing only the self-harming extremes (`expel_the_rohingya`, `provoke_clashes_on_the_thai_border`, `strike_our_age_old_enemy`), matching `DEN_no_scripted_path_plan`.

You were right that I had read this requirement and talked myself out of it.

common/national_focus/05_myanmar.txt
TemplarGeneral: Fixed in b0bb1bf — removed the three conflicting `focus_factors` from `BRM_PRO_US` (`the_path_of_intolerance`, `support_buddhist_nationalists`, `expel_the_rohingya`), and the same class of conflict in `BRM_PRO_CHINA`.

Worth flagging what I did wrong here, because it was not an oversight: I found this exact conflict during the audit and "solved" it by narrowing the tolerance killswitch so a US-aligned historical Myanmar falls through to tolerance, then wrote in the PR body that this was what the PRO_US author intended. That inverts the authority — an incidental alignment plan must never veto the path the player selected. The narrowing is reverted too: `the_path_of_tolerance` is back on the standard `BRM_ai_not_beacon_path` killswitch now that no plan can zero its rival.

common/national_focus/05_myanmar.txt
TemplarGeneral: Fixed in b0bb1bf. `BRM_support_buddhist_nationalists` and `BRM_expel_the_rohingya` moved from the shared `BRM_ai_rival_of_beacon_path` group to `iron_fist`, so both are zeroed for Historical and Beacon. `BRM_the_path_of_intolerance` stays on `rival_of_beacon` — the stance is historical, the mass expulsion is the Iron Fist escalation.

One consequence needed handling before this was safe: `Rohingya_Crisis` is a starting idea (`stability_factor -0.03`, `trade_opinion_factor -0.25`) whose only two cures were `expel_the_rohingya` and `reform_the_1982_citizenship_law` (Beacon-only). Killing the expulsion for Historical therefore stranded that malus permanently. So the AI gets its own route: a new `BRM_ai_path_category` (`is_ai = yes`, no loc keys) with `BRM_ai_rohingya_repatriation_accord`, visible only when `NOT = { BRM_ai_alt_path = yes }` — i.e. Historical and NO_PATH, never Beacon or Iron Fist, which already have focus cures.

The root cause of the mismatch you spotted was mine: I briefed the loc agent that on this path "the persecution and expulsion of the Rohingya proceeds", it returned "the Rohingya situation continues unchanged", and I shipped that without reading it against the mechanics. The desc is rewritten to match what the path now actually does.

common/national_focus/05_myanmar.txt
TemplarGeneral: Fixed in b0bb1bf, both halves as you specified.

- `BRM_turn_on_the_united_wa_state_army = 0` removed from `BRM_PRO_CHINA` (that plan is active from turn one — Myanmar starts at 87 Chinese influence, so `influence_array^0 = CHI` is true at game start).
- `crush_the_rebel_groups` and `turn_on_the_united_wa_state_army` get a path escape hatch on the availability gate, every other condition kept: `available = { OR = { CHI = { influence_less_20 = yes }  BRM_ai_iron_fist_path = yes } }`.

A reachability walk over the live focus set per rule state now shows `crush_the_rebel_groups` → `internal_colonialism` → `proclaim_an_end_to_the_insurgency` reachable under Iron Fist, which also restores that path's only cure for the `Ethnic_Insurgencies` starting idea.

This one came from a bad habit rather than a missed detail: I had moved that whole chain into `iron_fist` in the first place purely to clear one "mutex pairs boosted on both sides" line from `ai_path_report`, without ever checking Iron Fist could reach it. I also had `criticize_chinas_meddling` down-weighted to 10 purely to silence a tie warning, which steered Iron Fist toward keeping Chinese influence high — the very thing locking this gate. That is reverted to the standard 25; the remaining `criticize_chinas_meddling / seek_closer_ties_to_china` tie is a flavour choice inside one path's own spine, broken at runtime by PRO_CHINA's focus list, and I would rather leave it reported than engineer it away again.

Blazer135 (approved):

PR #3524, opened 2026-09-04
Faction Goals Reworking

MrP0tter:
Reworked Faction Goals to remove all collections, and try to prevent the performance spikes we've been having


common/scripted_effects/00_faction_goal_effects.txt
```diff
@@ -0,0 +1,665 @@
+faction_goal_update_daily_condition_caches = {
```
AngriestBird, line 1: These should be folded into one 00_factions_scripted_effects.txt

common/on_actions/00_on_actions.txt
```diff
@@ -1846,6 +1847,10 @@ on_actions = {
 				}
 			}
 
+			every_country = {
+				faction_goal_update_daily_caches = yes
```
AngriestBird: Why the extra every_country? Can it not be folded into the existing?

events/MD_faction_goal_cache_events.txt
```diff
@@ -0,0 +1,15 @@
...
+
+country_event = {
+	id = faction_goal_cache.1
+	hidden = yes
+	is_triggered_only = yes
+
+	immediate = {
+		if = {
+			limit = { is_faction_leader = yes }
+			log = "[GetDateText]: [This.GetName]: faction_goal_cache.1 executed"
```
AngriestBird, line 12: Unneeded event

common/scripted_effects/05_MERCOSUR_effects.txt
```diff
@@ -55,6 +55,7 @@ MER_recalculate_unasul_member_state_stability_factor = {
 }
 
 MER_remove_unasul_faction_membership_silent = {
+	set_temp_variable = { faction_goal_refresh_transferred_leader = 0 }
```
AngriestBird, line 58: This seems unneeded defensive code. We should already know since we saved the event_target. 

event_targets are ephemeral and are like temp variables. If we are using this in any events we need to ensure it's globally set. I think we can just remove the casting since we are shifting through it.

common/on_actions/MD_on_actions.txt
```diff
@@ -39,6 +39,10 @@ on_actions = {
 			# Per-week decay; must stay out of ingame_update_setup, which re-runs on demand (focuses, events, economy refresh button)
 			update_tension_premium = yes
 			ingame_update_setup = yes
+			if = {
+				limit = { is_faction_leader = yes }
+				country_event = { id = faction_goal_cache.1 days = 1 }
```
AngriestBird, line 45: Unneeded delay?

common/factions/goals/faction_goals_medium_term.txt
```diff
@@ -1315,13 +1148,7 @@ faction_goal_control_mandeb = {
...
 	}
 	visible = {
-		any_collection_element = {
-			collection = collection:faction_controlled_states_collection
-			distance_to = {
-				target = 1020
-				value < 2000
-			}
-		}
+		scope:faction = { check_variable = { faction_goal_mandeb_visible_cache = 1 } }
```
AngriestBird, line 1155: I'd wager that we honestly just state that if any has a capital_scope in the region then do these. Otherwise ignore them. Would reduce the caching effort. Alternatively we can add these to select nations. I don't really forsee many factions going out of their way to ensure the Mandeb strait is controlled. 

We can argue that the Suez and Panama are strategic goals rather then something we should seize to control but something the faction itself should want to do regardless of their distance if they have any capitals in the north or southern Americas or if you are a major global power.

common/factions/goals/faction_eap_goals.txt
```diff
@@ -6,47 +6,26 @@ faction_goal_transform_eap_forces = {
...
-					every_collection = {
-						named_collection = faction_members_collection
-						add = num_battalions_with_type@armor_Bat
-						add = num_battalions_with_type@L_arm_Bat
-						add = num_battalions_with_type@A_L_arm_Bat
-					}
-				}
+			check_variable = {
+				var = faction_goal_tank_battalions_cache
+				value = 4500
```
AngriestBird: Can we reduce this to 750? 4500 is pretty high for MD tank bat values.

common/scripted_effects/00_faction_goal_condition_cache_effects.txt
```diff
@@ -0,0 +1,665 @@
...
+
+	if = {
+		limit = {
+			scope:faction = {
+				OR = {
+					check_variable = { var = faction_goal_communist_member_count_cache value = 1 compare = less_than_or_equals }
+					check_variable = { var = faction_goal_fascist_member_count_cache value = 1 compare = less_than_or_equals }
+					check_variable = { var = faction_goal_anti_communist_member_count_cache value = 1 compare = less_than_or_equals }
+					AND = {
+						ROOT = { has_faction_goal = faction_goal_defeat_of_communism }
```
AngriestBird: Simplification should be done since we can just fold this into a single ROOT with all of the follow on conditions.

AngriestBird (changes requested):
This is a strong step in the right direction. We should explore a couple more of the other improvements that I noted. Agent comments coming soon TM

AngriestBird:
## Bottom line

This cache rewrite is not ready to merge. Several translated comparisons make faction goals complete immediately without meeting their objectives.

- `common/scripted_effects/00_faction_goal_condition_cache_effects.txt:603` **Blocker:** Every military-facility count is checked with `>= 0`, so `faction_goal_ensure_military_r_d` completes without any land, naval, or air facility. The visibility checks at line 595 also allow one facility instead of requiring zero. Use `>= 1` for completion and `<= 0` for visibility.
- `common/scripted_effects/00_faction_goal_condition_cache_effects.txt:285` **Blocker:** Great Offensive completion checks whether captured enemy-core states are `>= 0`, which is always true during war. Visibility at line 275 also permits one captured state instead of zero. Use `>= 1` for completion and `<= 0` for visibility.
- `common/scripted_effects/00_faction_goal_condition_cache_effects.txt:68` **Blocker:** Liberate Territory is always visible because every count is `>= 0`, and line 78 marks it complete with zero or one uncontrolled core. A faction with no occupied territory can select it for immediate rewards. Use `>= 1` for visibility and `<= 0` for completion.
- `common/scripted_effects/00_faction_goal_condition_cache_effects.txt:349` **Blocker:** The communist, fascist, and anti-communist purge goals use `>= 0` for availability and `<= 1` for completion through line 369. They can be selected with no hostile states and complete immediately, or complete while one hostile state remains. The member checks at lines 109-134 also allow one prohibited member. Restore `>= 1` availability and `<= 0` completion/member checks.
- `common/scripted_effects/00_faction_goal_condition_cache_effects.txt:58` **Major:** Other strict comparisons shifted by one. Aggressive Defense and Great Fortress allow one uncontrolled core; Liberate Territory AI thresholds trigger at 4/10 instead of 5/11; Secure Europe/East Asia complete at 16 instead of 17; Crush Democracy completes at 10 instead of 11; and Long Live the Revolution completes at 19 instead of 20. Preserve strict integer comparisons mechanically: `< 1` becomes `<= 0`, and `> N` becomes `>= N+1`.
- `common/factions/goals/faction_goals_long_term.txt:177` **Major:** Strategic Bombing is available at exactly three bombed states while also complete at three. The old condition accepted fewer than three.
- `common/on_actions/MD_on_actions.txt:44` **Major:** Membership-sensitive caches refresh weekly and one day late, while the normal join/leave hooks in `common/on_actions/00_on_actions.txt:306` and line 430 do not invalidate them. An active resource, building, or facility goal can consume stale completion after a contributing member leaves, or remain unavailable for roughly eight days after a member joins. Refresh affected faction caches on both transitions.
- `localisation/english/MD_frontend_l_english.yml:169` **Minor:** This PR also changes the displayed build from DEV to BETA and replaces the release date. Revert or split this unless the PR intentionally cuts the beta.

The GitHub checks are green, but they do not cover these semantic regressions.

BLUF

common/factions/goals/faction_sco_goals.txt
```diff
@@ -4,30 +4,21 @@ faction_goal_sco_connected_capitals = {
...
-							target_state = 716
-						}
-						has_railway_connection = {
-							start_state = 547
-							target_state = 720
-						}
-					}
-				}
-			}
+		has_railway_connection = {
```
AngriestBird, line 7: This is not state scoped. I am not sure this defaults to the scope:faction here should here this with this. Did we confirm that this is not tag scoped and faction scoped? This reads to me that we are checking if ROOT is the only one who has these railway connections.

AngriestBird (changes requested):

events/MD_faction_goal_cache_events.txt
MrP0tter: Needed event

common/on_actions/MD_on_actions.txt
MrP0tter: Needed delay to make sure the calculations will have the proper data. Will update this with a comment as well

common/on_actions/00_on_actions.txt
MrP0tter: resolved

common/scripted_effects/00_faction_goal_effects.txt
MrP0tter: resolved

common/factions/goals/faction_eap_goals.txt
MrP0tter: resolved

AngriestBird (approved):

Issue #2770, opened 2026-08-05
perf(gui): decouple remaining national scripted GUIs from global.md_gui_dirty


hasan4938:
### Summary
Currently, several national scripted GUIs use `dirty = global.md_gui_dirty`. Because `global.md_gui_dirty` updates daily, this forces the Clausewitz UI engine to re-evaluate and re-render national GUIs for all nations every day, creating unnecessary daily tick performance overhead.

### Proposed Improvement
1. Establish nation-scoped dirty variables (`dirty = global.update_TAG_ui`) for remaining national GUIs.
2. Wire nation-scoped dirty update triggers (`add_to_variable = { global.update_TAG_ui = 1 }`) in national focus trees, decisions, and events when national UI states change.
3. Remove `dirty = global.md_gui_dirty` from national scripted GUIs to eliminate daily UI re-evaluation overhead worldwide.

AngriestBird:
This is a good optimization but you need to unify the new variables into the mechanics or anything that could interact with the mechanics.

TemplarGeneral:
## Inventory — 31 entries still on `global.md_gui_dirty`

Audit of every country scripted GUI bound to the global. One branch/PR per country (`2770-gui-dirty-<TAG>`), cut from `main`, landing in any order; the cleanup PR goes last.

### Semantics used for the verdicts

> `dirty` defines when the scripted GUI should be updated; by default GUIs are updated every tick. The GUI will only update when the variable's value changes. **This only applies to updating the properties of the GUI, not to the GUI itself's visibility.**

So:

- Deleting a `dirty` line does **not** freeze a panel — it restores per-tick updating. Deletion is only correct where the update is a genuine no-op: no `triggers`/`properties`/`dynamic_lists`, and the `interface/*.gui` window carries no `[?var]` or scripted-loc text. Everywhere else, deleting is a perf regression.
- The entry-level `visible = { }` block is not dirty-gated, so a GUI whose only dynamic need is its own `visible` needs no dirty var at all.
- Element-level `_visible` / `_click_enabled` triggers **are** gated, and so is variable-driven window text.
- `tick_audit.py` flags a stateful `decision_category` GUI with no dirty binding (`gui-missing-dirty`, HIGH) — never strip `dirty` from an entry that has script blocks.

**Verdicts:** `DEL` = drop the `dirty` line (nothing gated) · `VAR` = nation-scoped dirty var + bumps at every write site · `DEAD` = remove the whole entry.

[...]

PR #3537, opened 2026-09-04
perf(gui): decouple ARM scripted GUIs from global.md_gui_dirty

TemplarGeneral:
Part of #2770 — the **ARM** item on the audit checklist.

## What

`global.md_gui_dirty` is bumped from the `on_weekly` block in `common/on_actions/MD_on_actions.txt:17`, so every panel bound to it re-evaluates for every nation on the same tick. This unbinds the five ARM entries.

**`artsakh_screen_gui` → own counter.** It has an `effects` block and its window text is scripted loc (`NKR_artsakh_desc_loc` → `[select_thingie_artsakh]` / `[discussion_phase_arts]`), so it needs a real dirty binding. New `update_arm_artsakh_dirty_variable` in `99_ARM_scripted_effects.txt`, same shape as `update_afg_toll_dirty_variable` (10000 rollover). Bumped from:

- the four click handlers, cancel/toggle-off included
- the three negotiation stage transitions — `karabakh.12.a`, `karabakh.15.a`, `karabakh.18.a`

Those are the only write sites for either state family; the decisions in `karabakh.txt` only read them.

**No `is_ai = no` guard**, deliberately, on the effect or the event call sites. `karabakh_decisions_category` is allowed to both ARM and AZE, and each stage transition writes the state on both countries from the acting country's scope — a guard on that scope would skip the bump exactly when the AI acts and the human is the other party, which is the case that needs the refresh. The counter fires at most a handful of times per campaign.

**`armeconomic_gui`, `armpolice_gui`, `armarmy_gui`, `nkr_gui` → `dirty` line dropped.** No `effects` / `triggers` / `properties` / `dynamic_lists`, and their windows in `MD_SUB_subjects_guis.gui` are static `iconType`s whose only dynamic content is `pdx_tooltip`, which renders on hover and was never gated by `dirty`. Per-tick updating of these is a no-op, and `tick_audit.py`'s `gui-missing-dirty` only fires on stateful entries.

The file also had mixed leading-space indentation in the `artsakh_screen_gui` block; normalised to tabs.

## Review follow-up

Both of AngriestBird's comments are addressed, and the second one is applied to the negotiation stages as well since they are the same shape:

- `gu_stepanakert_select` / `gu_karvachar_select` / `gu_hadrut_select` → **`NKR_artsakh_ui_selected`** = 1 / 2 / 3, cancel writes `0`.
- `gu_first_stage` / `gu_second_stage` / `gu_third_stage` → **`NKR_artsakh_stage`** = 1 / 2 / 3.

Each click handler and each stage transition now writes one value instead of a set/clr fan-out, and the `NOT = { OR = { … } }` "nothing selected" / "not past this stage" guards in `karabakh.txt` and `99_ARM_scripted_localisation.txt` collapse to a single equality check. All readers are `visible` blocks and scripted-loc triggers, so no `custom_trigger_tooltip` wrappers are needed; the stage writes in `events/Karabakh.txt` keep their existing `custom_effect_tooltip`. `NKR_` domain prefix per AGENTS.md — the state is shared by ARM and AZE, matching `NKR_attack_stepanakert` / `NKR_artsakh_desc_loc`. Net −30 lines across the five files.

The explanatory comment above `update_arm_artsakh_dirty_variable` is gone.

Both points are now in the **Conventions for each PR** section of the #2770 checklist, so the remaining 30 country PRs pick them up.

## Known behaviour change

`NKR_artsakh_desc_loc` also renders `[artsakh_strong_or_weak]`, which reads a live `strength_ratio` against the rival — continuous state with no discrete write site. It now refreshes on interaction (clicks, stage transitions) rather than weekly. It is one flavour line inside a longer descriptive block, the panel is only reachable by ARM/AZE while the conflict is unregulated, and the player's own clicks always refresh it. Adding a periodic bump to keep it live would reintroduce the exact cost this issue removes.

## Validation

- `validate_variables.py` — no issues found
- `tick_audit.py --spot-check` on the changed files — critical=0, high=0, medium=0, low=0
- `check_common_mistakes.py` — no findings on the changed files
- `grep md_gui_dirty common/scripted_guis/99_ARM_scripted_guis.txt` → empty; the six `gu_*` flags have zero references repo-wide

## Test plan

As ARM (or AZE), open the Nagorno-Karabakh decision tab:

1. Click each of the three region buttons — the selection line in the right-hand description changes on the same tick, and the matching `NKR_attack_*` decision appears while the other two do not.
2. Click cancel — the line returns to the "nothing selected" text and all three attack decisions plus the drone recon disappear, on the same tick.
3. Run a negotiation to the first stage, then the second and third — the discussion-phase line updates for both ARM and AZE, including when the other party is AI, and exactly one `NKR_armenia_*_posistion` decision is visible at each stage.
4. Open the Armenian economy / police / army / Artsakh government tabs — icons and hover tooltips render as before.

No changelog entry (dirty-variable plumbing, no player-facing content).


common/scripted_effects/99_ARM_scripted_effects.txt
```diff
@@ -2358,3 +2358,14 @@ ARM_unionstate_integration_effect = {
 		}
 	}
 }
+
+#Own counter rather than global.md_gui_dirty, which every other country's panel also bumps
```
AngriestBird: Comment not needed

common/scripted_guis/99_ARM_scripted_guis.txt
```diff
@@ -1,50 +1,50 @@
...
-		 window_name = "artsakh_screen_gui"
-		dirty = global.md_gui_dirty
+		window_name = "artsakh_screen_gui"
+		dirty = global.ARM_artsakh_dirty_update_var
 
-		 effects = {
+		effects = {
 			region_stepanakert_click = {
-				 set_country_flag = gu_stepanakert_select
+				set_country_flag = gu_stepanakert_select
```
AngriestBird: Less of  a you thing but we can simplify this to a set_variable  = { artsakh_ui_selected = 1 } for this frame and then repeat that for each giving them a unique identifier for this to reduce the code.

AngriestBird (changes requested):

common/scripted_effects/99_ARM_scripted_effects.txt
TemplarGeneral: Removed.

common/scripted_guis/99_ARM_scripted_guis.txt
TemplarGeneral: Done, and applied to the negotiation stage flags too since they are the same shape.

`gu_stepanakert_select` / `gu_karvachar_select` / `gu_hadrut_select` -> `NKR_artsakh_ui_selected = 1 / 2 / 3`, cancel writes `0`. `gu_first/second/third_stage` -> `NKR_artsakh_stage = 1 / 2 / 3`. Every handler and every stage transition now writes one value instead of a set/clr fan-out, and the `NOT = { OR = { ... } }` "nothing selected" / "not past this stage" guards in `karabakh.txt` and the scripted loc collapse to a single equality check. Readers are all `visible` blocks and scripted-loc triggers, so no tooltip wrappers are needed.

Net -60 lines. Domain prefix `NKR_` per AGENTS.md (shared by ARM and AZE, matching `NKR_attack_stepanakert`). Also added both points to the conventions section of the issue #2770 checklist so the remaining 30 country PRs pick them up.

AngriestBird (approved):

PR #3545, opened 2026-09-04
perf(gui): decouple CZE_tech_scripted_gui from global.md_gui_dirty

TemplarGeneral:
## Summary
- Part of the #2770 GUI dirty-decoupling checklist (tag: CZE).
- `CZE_tech_scripted_gui` has no `triggers`/`properties`/`effects`/`dynamic_lists` to gate — only a static entry-level `visible` block, which is never dirty-gated anyway — so `dirty = global.md_gui_dirty` was forcing a needless daily re-evaluation of this panel for every nation. Dropped per the checklist's **DEL** verdict.

## Test plan
- [ ] `validate_scripted_gui.py` on the changed file: no new `DIRTY_VAR_UNDEFINED` / `DIRTY_SCOPE_MISMATCH`
- [ ] In game as CZE: open the tech tree armour folder panel and confirm `CZE_tech_scripted_gui` still renders normally (static label, no regressions)


AngriestBird:
I would attach this to the other CZE GUI dirty variable. The AI is making some assumptions but every element is refreshed so if it's even just doing things like showing data that's not needed to run every single frame.

AngriestBird (approved):

PR #3550, opened 2026-09-04
perf(gui): decouple the ENG devolution button from global.md_gui_dirty

TemplarGeneral:
Part of #2770 — the **ENG** item on the audit checklist.

## What

`global.md_gui_dirty` is bumped from the `on_weekly` block in `common/on_actions/MD_on_actions.txt:17`, so every panel bound to it re-evaluates for every nation on the same tick. Both ENG entries are moved onto a dedicated counter instead.

**`global.ENG_devolution_dirty_update_var`**, bumped by a new `update_eng_devolution_dirty_variable` in `common/scripted_effects/99_ENG_scripted_effects..txt` (standard `< 10000` / else `= 1` shape), is now the `dirty` variable for both `ENG_devolution_gui_window` and `open_devolution_main_button`.

Bump sites:

- every player click handler in both entries — the three map buttons, their three right-clicks, back, close, and the map-mode open button
- a new `on_daily_ENG`, gated on `is_ai = no` plus `has_country_flag = open_devolution_screen`

The daily bump is what keeps the panel honest: it renders `[?ENG_scottish_agitation]`, `[?ENG_scottish_nationalists]` and the Welsh / Northern Irish equivalents through the `*_VAR` loc keys, and those variables are written from 12 sites in `05_ENG_decisions.txt`, 32 in `05_united_kingdom.txt`, 17 in `events/05_united_kingdom.txt` and `on_monthly_ENG`. `ENG_var_cap` is a chokepoint for the first two groups but not the events, so a guarded daily bump is the reliable option rather than instrumenting 60-odd writers. Same shape as the HEZ panel in #3553.

`dirty` does not gate `visible`, so the map-mode button still hides on the tick `ENG_devolution_purged` is set.

## Review follow-up

The **Conventions for each PR** item from #2770 (originally AngriestBird's request on #3537) applies here — the devolution panel's three menu flags are mutually exclusive, are referenced nowhere in the repo outside this file, and are read only in element `_visible` triggers:

`ENG_scotland_menu_opened` / `ENG_n_ireland_menu_opened` / `ENG_wales_menu_opened` → **`ENG_devolution_ui_selected`** = 1 / 2 / 3, with `0` = nothing selected.

Each of the four click handlers now writes one value instead of a set plus two clears, and the `NOT = { OR = { … } }` "nothing selected" guard in the open-button handler collapses to a single equality check. All readers stay in `triggers`, so no `custom_trigger_tooltip` wrappers are needed. `ENG_` tag prefix per AGENTS.md — the panel is ENG-only, unlike the shared `NKR_` state in #3537. The variable is not seeded: unset already reads `0`.

`ENG_stop_cliking_me_headass_SCO` / `_NIR` / `_WAS` are left alone — independent 91-day cooldown flags, not exclusive selection state.

## Drive-by fix

Three trigger keys were declared twice in the same `triggers` block, the second declaration silently overriding the first:

| Key                                   | Weak duplicate       | Dependency-gated copy                       |
| ------------------------------------- | -------------------- | ------------------------------------------- |
| `ENG_scots_mostly_dependent_visible`  | region selected only | region + `ENG_scotland_is_mostly_dependent` |
| `ENG_north_ireland_dependent_visible` | region selected only | region + `ENG_north_ireland_is_dependent`   |
| `ENG_wales_dependent_visible`         | region selected only | region + `ENG_wales_is_dependent`           |

The elements are named `*_dependent` and their loc reads "Dependent" / "Mostly Dependent", and the three flags are set at game start in `history/countries/ENG - United Kingdom.txt:1308-1310`, so the dependency-gated copies are the intended ones. The three weak duplicates are removed; the surviving gates are unchanged in meaning.

## Validation

- `validate_common_mistakes.py` — no issues found
- pre-commit on the changed files — passed
- `grep md_gui_dirty common/scripted_guis/99_ENG_scripted_guis.txt` → empty; the three `*_menu_opened` flags have zero references repo-wide

## Test plan

As ENG, with the devolution button on the map-modes bar:

1. Click the button — the panel opens on Scotland, and the agitation / parliament / dependency rows plus the info icons and back button render on the same tick.
2. Click each of the three map buttons — the region rows swap immediately, the Irish tension icon shows only for Northern Ireland, and the agitation icon only for Scotland and Wales.
3. Confirm each region's "Dependent" / "Mostly Dependent" line appears only while that region's dependency flag is set, not merely because the region is selected.
4. Leave the panel open and complete an agitation-changing decision or focus — the number moves within a day.
5. Click back — all three region panels and the info icons hide; reopen the panel and it lands on Scotland again.
6. Right-click a region whose nation does not exist — the appease decision activates and the button greys out for 91 days, as before.
7. Complete the focus that sets `ENG_devolution_purged` — the button disappears.

No changelog entry (dirty-variable plumbing, no player-facing content).


common/scripted_guis/99_ENG_scripted_guis.txt
AngriestBird, line 2: Entirely missing a dirty variable

AngriestBird (changes requested):

common/scripted_guis/99_ENG_scripted_guis.txt
TemplarGeneral: Bound in ca8d39e. The panel and the map-mode button now share `global.ENG_devolution_dirty_update_var`, bumped from every player click handler and once a day from a new `on_daily_ENG` while the panel is open, which keeps the agitation and parliament numbers live without the per-tick evaluation.

TemplarGeneral: Updated it

AngriestBird (approved):

PR #3553, opened 2026-09-04
perf(gui): decouple the International Hezbollah button from global.md_gui_dirty

TemplarGeneral:
Part of #2770 — the **HEZ** item on the audit checklist.

## What

`global.md_gui_dirty` is bumped from the `on_weekly` block in `common/on_actions/MD_on_actions.txt:17`, so every panel bound to it re-evaluates for every nation on the same tick. This unbinds the one HEZ entry that was on it.

**`open_international_hezbollah_gui` → `dirty = global.HEZ_international_dirty_update_var`.** A **VAR** item, not a **DEL**: the entry carries a real `triggers` block — `open_international_hezbollah_gui_click_enabled`, gated on `has_completed_focus = HEZ_International_Hezbollah` — so the `dirty` line gates genuine work and dropping it would be a per-tick regression.

New `update_hez_international_dirty_variable` in `common/scripted_effects/99_HEZ_scripted_effects.txt`, same shape as `update_afg_toll_dirty_variable` (10000 rollover). The gate flips exactly once per campaign, so there is exactly one bump: a `hidden_effect` in that focus's `completion_reward`. `complete_national_focus = HEZ_International_Hezbollah` has zero references repo-wide, so that is the only path. No `is_ai = no` guard — the panel is HEZ-only and the AI driver in `common/on_actions/99_HEZ_on_actions.txt:16-25` never touches the gate.

The big `HEZ_international_hezbollah` panel entry shares the same variable, `dirty = global.HEZ_international_dirty_update_var`. Its `properties` block drives `mission_progressbar_container = { frame = hez_days_remaining_loc }`, a daily countdown, so the bump schedule carries it: every player click handler in the entry bumps the variable (the nine region buttons, the five country flags, the four mission tiles), and a new `on_daily_HEZ` in `common/on_actions/99_HEZ_on_actions.txt` bumps it once a day while `international_hezbollah_tab_open` is set, guarded on `is_ai = no`. Gating that daily bump on the tab flag rather than `shared_mission_in_progress` also covers the day the 60-day timed flag expires, when the panel has to fall back to the empty picture and re-enable the tiles. The bumps sit in the click handlers rather than inside `HEZ_do_*`, since those are also reached from `HEZ_ai_dispatch_mission` — guard the call site, not the shared effect.

## Review follow-up — flag collapses

The **Conventions for each PR** item on #2770 (originally AngriestBird's request on #3537) applies twice here. Both families are read only in `visible` / element `_visible` triggers and scripted loc, so no `custom_trigger_tooltip` wrappers are needed, and neither variable is seeded — unset already reads `0`.

### Region selection → `HEZ_selected_region`

`HEZ_selected_africa` / `HEZ_selected_south_america` / `HEZ_selected_north_america` / `HEZ_selected_west_asia` → **`HEZ_selected_region`** = 1 / 2 / 3 / 4, with `0` = nothing selected. 168 references, confined to `99_HEZ_scripted_guis.txt` and `99_HEZ_scripted_localisation.txt`; no decisions, events, on_actions or `.gui` file touches them. They were already hand-enforced exclusive — every click handler cleared the other three before setting its own.

- The nine region click handlers drop from 20 lines each to one `set_variable`. Their `if = { limit = { OR = { NOT own flag; any other flag } } … }` guard was pure no-op protection with no side effects inside the branch, so the unconditional set is equivalent. The nine commented-out `#else = { clr_country_flag = … }` lines go with them.
- The leading `NOT = { has_country_flag = HEZ_selected_north_america }` in `country_flag_3/4/5_selected_button_visible` is now implied by the region equality in every surviving `OR` branch, so it is removed rather than restated.

`HEZ_` tag prefix per AGENTS.md — the panel is HEZ-only, unlike the shared `NKR_` state in #3537.

### Mission selection → `HEZ_ongoing_mission`

`has_selected_recruit_militia` / `has_selected_small_arms` / `has_selected_influence_nation` / `has_selected_attack_zionists` → **`HEZ_ongoing_mission`** = 1 / 2 / 3 / 4.

These are timed flags, which normally rules a variable out — but the timer never belonged to them. `hez_days_remaining_loc` reads `[?shared_mission_in_progress:days_left]`, and both families are set in the same effect with the same `days = 60`, so the four type flags were only ever a second copy of *which* mission. `shared_mission_in_progress` is the authoritative "a mission is running" and **stays exactly as it is** — the countdown, the four `*_icon_click_enabled` locks, the `HEZ_ai_dispatch_mission` gate, the `random_list` `factor = 0` modifiers and the AI `on_action` are all untouched.

Every scripted-loc reader is gated on that lock flag, so a stale variable value is never observable and no reset path is needed. Each dispatcher branch keeps the flag check alongside the `check_variable`, so the display does not depend on `defined_text` branch ordering, and the two 12-line `NOT = { OR = { …four flags… } }` fallbacks collapse to one negated flag check each.

## Drive-by fix

The four `*_icon_click` handlers in the scripted GUI were byte-identical copies of `HEZ_ai_do_recruit_militia` / `_sell_weapons` / `_influence` / `_attack_israel` in `99_HEZ_scripted_effects.txt`. The effects are renamed **`HEZ_do_*`** (they are no longer AI-only) and the click handlers now call them. Their only other callers were the four `random_list` buckets inside `HEZ_ai_dispatch_mission`. This halves the write sites for `HEZ_ongoing_mission` from eight to four and removes the way the player and AI paths could drift apart again.

## Save compatibility

A save with a mission already in flight keeps `shared_mission_in_progress` but has `HEZ_ongoing_mission = 0`, so the mission picture and name render blank until that mission ends (≤60 days). The countdown, the lock and the AI are unaffected.

## Validation

- `validate_scripted_localisation.py`, `validate_set_variables.py`, `check_common_mistakes.py` — no issues found
- `tick_audit.py --spot-check` on the changed files — no `gui-missing-dirty`, no new findings (3 pre-existing `every_country` MEDIUMs in `05_Hezbollah.txt` are untouched)
- pre-commit on the four changed files — passed
- `grep md_gui_dirty common/scripted_guis/99_HEZ_scripted_guis.txt` → empty; the eight old flags and `HEZ_ai_do_*` have zero references repo-wide

## Test plan

As HEZ from the 2000 start:

1. Before `HEZ_International_Hezbollah`, the politics-tab button is greyed with its tooltip. Complete the focus — it becomes clickable on the same tick, without waiting for a weekly bump.
2. Open the panel. Click each region button that its unlock focus has enabled — the continent icon swaps to its `_selected` sprite, the five country flags repopulate, and on North America flags 3-5 and the Attack Israel tile hide.
3. Click a country flag, then a mission tile. Confirm the correct mission picture and name appear, the "N Days left" bar counts down, and all four tiles grey out with the "mission in progress" tooltip while it runs.
4. After 60 days the mission event fires, the panel returns to the empty picture and "no ongoing mission", and a new mission can be started.
5. Let an AI HEZ run (communist, focus complete) and confirm `HEZ_ai_dispatch_mission` still launches missions on the same cadence.

No changelog entry (dirty-variable plumbing plus a flag-to-variable refactor, no player-facing content change).


common/scripted_guis/99_HEZ_scripted_guis.txt
AngriestBird, line 2: Doesn't have a dirty variable

AngriestBird (changes requested):

common/scripted_guis/99_HEZ_scripted_guis.txt
TemplarGeneral: Bound in d3e80e6 — the panel now carries `dirty = global.HEZ_international_dirty_update_var`, the same variable as the politics-tab button. Every player click handler in the entry bumps it, and a new `on_daily_HEZ` bumps it once a day while the tab is open (`is_ai = no` guarded), which keeps the days-left countdown and the mission-expiry fallback live without the per-tick evaluation.

AngriestBird: This does not need to bumped once a day. It should move to only if and when the incremental changes are done.

AngriestBird (approved):

PR #3555, opened 2026-09-04
perf(gui): decouple the Israeli scripted GUIs from global.md_gui_dirty

TemplarGeneral:
Part of #2770 — the **ISR** item on the audit checklist.

`global.md_gui_dirty` is bumped weekly from `common/on_actions/MD_on_actions.txt:17`, so every panel bound to it re-evaluates for **every nation** on that tick. This unbinds both remaining ISR entries.

## `ISR_attack_GUI` — VAR (reuse), zero new call sites

The expensive one: 19 `properties` image swaps, each resolving a `*_influence_per` scripted loc, each of which evaluates up to three `*_has_more_than_N_influence_trigger`s that `OR` over ~10 `influence_array^N` / `influence_array_val^N` pairs.

Bound to `global.update_influence_ui`, which `update_dirty_influence_var` already bumps from the main influence path (`00_influence_scripted_effects.txt:1147`, right after `recalculate_influence`), the aid/puppet paths, the auto-influencer path, all ten click handlers in `01_influence_scripted_guis.txt` and `MD_on_actions.txt:440`. The panel reads *global* influence state, so a global influence counter is the correct binding — same as `countrydiplomacyview_influence`.

Drive-by: `update_dirty_influence_var` had **no rollover branch**. With a float variable and ~200 call sites an uncapped counter eventually reaches the range where `+1` stops registering and every bound panel freezes silently. It now carries the standard 10000 rollover.

## `israel_knesset_gui` — VAR

New `update_isr_knesset_dirty_variable`, same shape as `update_afg_toll_dirty_variable`. The panel's live content is the three `GFX_knes{3,5,6}_icon_visible` triggers plus `ISR_stats_desc_loc` = `[knes_status]` + `[knes_approvio]`.

Its `properties` block is deleted outright — `knesset_diagr` is not an element in `interface/MD_israel_gui.gui` and `knes_ico` is not a `defined_text` anywhere; the only two hits repo-wide were those two lines.

The three knesset-size flags are **not** collapsed into an indexed variable: 12 readers sit in decision `available` blocks and 22 more in `05_israel.txt`, where a bare `check_variable` is a `validate_variables.py` error and would need 34 `custom_trigger_tooltip` wrappers. The checklist's flag-collapse convention explicitly excludes that case.

## Unifying the mechanic

Per AngriestBird's review note on the issue. The 12 knesset bill decisions in `common/decisions/Israel.txt` had **byte-identical** `complete_effect` bodies below the `log` line and `remove_effect` bodies differing only in one reward effect. They now call `ISR_start_knesset_bill` / `ISR_end_knesset_bill` — **528 duplicated lines removed**, and 12 dirty write sites become one inside `ISR_end_knesset_bill`. Each decision keeps its own `log` line and its own reward; the reward stays inside `hidden_effect`, so tooltip output is unchanged.

Remaining bumps are 16 event options in `events/Israel_events.txt`: the three coalition events that resize the Knesset, the six knesset bill events, and the seven `israel_likud` options that move `bill_support`.

No `is_ai = no` guards — each is an ISR-only narrative option firing a few dozen times per campaign, the panel's category is already `allowed = { is_ai = no original_tag = ISR }`, and the shared effect carries no internal guard per `scripted-gui-rules.md`.

## Two deliberate behaviour changes

- `ISR_up_welfare` / `ISR_down_welfare` queued the `israel.knesset.7` scandal at `days = 10` where the other ten use `days = 40`. The shared effect uses 40. Both land inside the 50-day `days_remove` window, so the bill still resolves after the scandal either way.
- `israel.knesset.0` tested `ISR_good_size_knesset` / `ISR_big_size_knesset` before the two `has_idea = ISR_yesman_likud` variants in the same `else_if` chain, so the intended **+75 / +55** bonuses could never fire. The yesman branches are now tested first.

## Validation

`validate_scripted_gui.py`, `validate_set_variables.py`, `validate_variables.py`, `validate_scripted_localisation.py`, `validate_events.py` and `validate_common_mistakes.py` all clean. `tick_audit.py --spot-check` on the changed files reports no `gui-missing-dirty` and no new findings (the 5 pre-existing MEDIUM `every_country` hits in `israel.82` / `.88` / `israel_likud.22` / `.48` are untouched). Scoped `pre-commit` on the five files passes.

## Test plan

1. As Israel, open the Knesset panel; run a bureaucracy/military/welfare bill and confirm `[knes_approvio]` flips on the same tick a `bill_support` event resolves, and that the tooltip on each of the 12 bill decisions is unchanged.
2. Trigger a coalition event that resizes the Knesset; confirm the `GFX_knes{3,5,6}` icon and `[knes_status]` change on the same tick.
3. Open the bombing panel and confirm the 19 influence icons update on the tick an influence action lands.


common/scripted_guis/99_ISR_scripted_guis.txt
AngriestBird, line 18: This should have the same update treatment while we are in here and streamlined.

common/scripted_guis/99_ISR_scripted_guis.txt
```diff
@@ -58,7 +58,7 @@ scripted_gui = {
 		context_type = decision_category
 		parent_window_token = top_bar
 
-		dirty = global.md_gui_dirty
+		dirty = global.update_influence_ui
```
AngriestBird: Why is this using the influence dirty variable?

AngriestBird (changes requested):

common/scripted_guis/99_ISR_scripted_guis.txt
TemplarGeneral: Done in 5d4e8eb. The four click handlers now call a single `update_isr_screen_dirty_variable` over `global.ISR_screen_dirty_update_var`, matching the MD standard shape (`< 10000`, else `= 1`) and the `update_isr_knesset_dirty_variable` this branch already added — the inline copies were on a `< 200` cap with an `else` that set `0`. Also dropped the stray leading-space indentation in that block.

The bigger half of the "same update treatment": nothing outside this file ever bumped that variable, so the numbers the window actually renders — `ISR_palestine_situation`, `PAL_arab_popularity`, `PAL_war_support_in_isr`, `PAL_arab_parties_rulling`, all interpolated into `ISR_palestine_desc_loc` — only refreshed when the player clicked a region button. `ISR_update_palestine` is their sole writer and runs weekly from `on_weekly_ISR`, so the bump goes there behind `is_ai = no`.

common/scripted_guis/99_ISR_scripted_guis.txt
TemplarGeneral: Fair point — changed in 5d4e8eb. It is now `global.ISR_attack_dirty_update_var`.

Reasoning for the original choice, for the record: every dynamic element in `ISR_attack_decisions_GUI` is one of the 19 `GFX_*_no_influence_icon` sprites, whose images come from the `*_influence_per` defined_texts, which resolve through `iran_has_*_influence_trigger` to `influence_array` / `influence_array_val`. So the panel renders influence-system state and nothing else. But `global.update_influence_ui` is bumped from `change_influence_percentage`, i.e. by every influence change on the map by any country — an Israel-only panel does not need to be woken by Bolivia.

The dedicated counter is bumped from the panel's own sources instead: the two bombing decisions that move influence (`ISR_capture_iranian_officials`, `ISR_kill_souleimani` — six `change_influence_percentage` calls between them) at the end of their `remove_effect`, plus the weekly ISR catch-up in `on_weekly_ISR` for drift caused by other nations. Same cadence `global.md_gui_dirty` was giving, without the global coupling.

Two things I found in that properties block while in there:

- `[israel_influence_per]` had no `defined_text` anywhere in `common/`, so that icon never resolved. Added it in this commit on the four-tier `jordan_influence_per` shape scoped to `ISR`.
- `iran_influence_per` has a single `text` entry gated on `PER = { original_tag = PER }`, which is always true, so that icon is effectively hardcoded to `GFX_iran_high_influence` regardless of actual influence. Left alone as out of scope — happy to fix it here or file it separately, your call.

MrP0tter (approved):

MrP0tter (approved):

PR #3559, opened 2026-09-04
Unlock SRA Creating a Federal Government after USA collapse

AngriestBird:
## Bottom line

SRA (CSA) never got `USA_collapse_event.23` after the 2003 railroad-strike collapse, so `the_dust_settles` was never set and `CSA_focus_creating_a_federal_government` stayed locked. The event now waits 2 days, same as Cascadia and the Free States.

## Why

Event 9 creates CSA in `immediate`, then both options fired event 23 in the same hour. HOI4 drops country events sent to a nation that just spawned. Cascadia (`USA_collapse_event.30`, 2 days) and Free States (`USA_collapse_event.19`, 3 days) already delay their unlock events; CSA was the odd one out.

After the delay, CSA exists, "Southern Problems" can fire, and the first focus becomes available.

BLUF

### DIFF
--- a/events/United States.txt
+++ b/events/United States.txt
@@ -12738,7 +12738,10 @@ country_event = {
 		name = USA_collapse_event.9.o1
 		log = "[GetDateText]: [This.GetName]: USA_collapse_event.9.o1 executed"
 		CSA = {
-			country_event = USA_collapse_event.23
+			country_event = {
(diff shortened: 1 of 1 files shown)


2ndLTLuige2010 (approved):
I approve of you fixing strange shit.

PR #3567, opened 2026-09-05
Make the Myanmar Iron Fist availability escape hatch AI-only

JasonBreen:
## Summary

The Iron Fist game rule let human Myanmar bypass the Chinese-influence requirement on `BRM_crush_the_rebel_groups` and `BRM_turn_on_the_united_wa_state_army`.

Both availability blocks now use `if = { limit = { is_ai = yes } ... }` with a human `else` branch, addressing AngriestBird's review. AI Myanmar retains the existing choice of Chinese influence below 20 or the Iron Fist path; humans require influence below 20 regardless of the AI game rule. The United Wa focus still requires peace.

Merged current `main`, resolved both availability conflicts, and removed the unused English `BRM_IRON_FIST_PATH_REQUIREMENT` tooltip key. No new localisation is needed.

## Verification

- Parsed both updated availability blocks and checked 80 AI/player, path, influence, and war combinations, including influence 19/20/21: passed.
- Diff whitespace and script/localisation encoding checks: passed.
- [GitHub CI passed](https://github.com/MillenniumDawn/Millennium-Dawn/actions/runs/33992829457) on `e4c6046`: zero errors; focus-tree, style, localisation, and common-scripting checks passed.
- Advisory warnings: two `untooltipped-available-scripted-trigger` findings occur inside the `is_ai = yes` branches, which humans cannot enter. The validator does not exempt these conditional AI branches. The other 24 warnings concern `battleship` references in the unchanged British MIO file and are outside this two-file diff.
- In-game tooltip rendering and AI progression have not been tested.

## In-game checks

1. Human Myanmar, Iron Fist rule: both focuses require Chinese influence below 20; no AI path requirement should appear.
2. AI Myanmar, Iron Fist rule: the influence bypass remains available.
3. Other AI paths: influence below 20 is required.
4. At war: `BRM_turn_on_the_united_wa_state_army` remains unavailable.


TemplarGeneral:
No this isn't the solution. These should be ai only so no localisation is needed.

JasonBreen:
Changed the approach — dropped the tooltip and the loc key entirely. Both sites are now `hidden_trigger = { is_ai = yes  BRM_ai_iron_fist_path = yes }`, so the escape hatch is AI-only and renders nothing for a human. AI behaviour is unchanged from #3496; a human Myanmar goes back to the pre-#3496 `CHI = { influence_less_20 = yes }` gate, which is right — the hatch was only ever needed because `BRM_PRO_CHINA` keeps the AI above 20, and a player has the agency to push it down.

common/national_focus/05_myanmar.txt
```diff
@@ -1980,7 +1980,10 @@ focus_tree = {
 		available = {
 			OR = {
```
AngriestBird: we can fold this to if = { limit = { is_ai = yes } and then use those to control the pathing on this for visiual purposes.

AngriestBird (changes requested):

AngriestBird (approved):

PR #3571, opened 2026-09-05
Add AMD corporate history and extend the NVIDIA chain

JasonBreen:
Two pieces of the same gap. AMD is the only other company that designs x86 and it had no chain at all — the sole reference to it anywhere in the mod is Canada's ATI branch. NVIDIA had a chain that jumped 2006 → 2013 → 2016 → 2022, skipping both moments the company nearly did not survive.

## AMD, new chain

**2000** the gigahertz race — ship an Athlon at 1 GHz before Intel and be first in the history of the part, or hold for yield · **2003** extend x86 to sixty-four bits while Intel bets on Itanium, and make Intel license the architecture back · **2009** a leading-edge fab now costs more than AMD is worth, and Abu Dhabi will take the manufacturing as GlobalFoundries · **2015** Lisa Su spends four years and most of the cash on a core designed from nothing · **2020** thirty-five billion in stock for Xilinx.

**The 2006 ATI acquisition is deliberately absent.** `CAN_ati_amd_integrated` already decides it from Canada's side. Two chains must not resolve the same transaction with no shared state.

The permanent outcome is the counterfactual: AMD keeping its fabs leaves America with a leading-edge x86 line that is not Intel's, which is owned capacity. The historical route — Zen plus Xilinx — is a 365-day market position, because the parts are designed in America and built by TSMC.

## NVIDIA, two beats and a fifth outcome

**2008 Bumpgate** — the packaging on a generation of mobile parts fails under heat and the notebooks carrying them are dying in customers' hands. Take the charge and replace them, or contest the claims and let every laptop maker decide whether to design NVIDIA in again.

**2018** — Ethereum collapses, the channel is full of graphics cards, and shareholders are asking how mining revenue was reported as gaming. Write the inventory down, or keep the miners supplied and book revenue that will not repeat.

The capstone gains `usa_economic_events.335.e`, a disciplined graphics supplier that never reached scale and never shipped into a bubble either; the old fallback moves to `.f` with `NOT = { has_country_flag = USA_nvidia_channel_reset }` added. The four existing outcomes and their triggers are untouched, so no existing playthrough resolves differently. The chain header now reads seven milestones and five futures.

`USA_corporate_history_2000` is new — the United States had no year-2000 entry point — and is dispatched from `MD_event_on_startup_events` alongside ASML's.

## Verification

- `tools/precommit_validate.py` on all six files: 9 validators, zero failures.
- Reference integrity checked in full: every event, milestone, apply-effect, outcome idea, `outcome_settled` entry and loc key resolves; every capstone flag is set by a beat; every aggregator is dispatched exactly once.
- The NVIDIA capstone still presents exactly one option for any flag combination — the new `.e` and the moved `.f` split the old fallback's condition on one flag, so the five options remain mutually exclusive and exhaustive.
- Outcomes Only still resolves to `.a` for NVIDIA (the historical flags satisfy `cuda_platform` + `accelerator_scale`), so `USA_nvidia_milestone_2026` needed no change; AMD's resolves to `.a`, asserted rather than assumed.
- AI weighting: every new beat keeps exactly one option selectable while `bankruptcy_incoming_collapse` is active, and no option that spends carries the escape clause.
- Naming: all eight new descriptions name the company under discussion, none over three sentences.
- AMD uses event IDs 510–515 after conflict resolution with `main`; Anduril retains 500–505. Event calls and English localisation use the separate ranges, and AMD remains last in the event file.

Conflict resolution and event-ID collision repair by GPT-6 (OpenAI Codex); the merge commit includes model co-author attribution.


events/United States.txt
```diff
@@ -39782,3 +39917,361 @@ country_event = {
...
+		hidden_effect = { USA_amd_apply_integrated_silicon = yes }
+		ai_chance = { base = 100 }
+	}
+
+	option = {
+		name = usa_economic_events.505.b
+		trigger = {
+			has_country_flag = USA_amd_zen_bet
+			NOT = { has_country_flag = USA_amd_xilinx }
+		}
```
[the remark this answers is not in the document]
JasonBreen: Fixed in 8b9b560fed. You are right that the fab decision was doing nothing: a player who kept the fabs and then funded Zen was awarded Integrated Data Centre Silicon, whose own description says none of it is manufactured in America. The capstone now splits on `USA_amd_fabless` first — owning the manufacturing and designing a competitive core is the permanent integrated-device outcome, owning it without the core is a new fifth outcome (legacy fabrication holdout), and the three fabless leaves keep their existing conditions. All eight flag combinations resolve to exactly one option, verified by simulating every play-through against the triggers as written in the file, and the historical route still lands on `.a`, so the Outcomes Only wiring is unchanged.

localisation/english/MD_focus_USA_l_english.yml
```diff
@@ -6034,6 +6034,65 @@
...
+ usa_economic_events.500.a: "Ship first and take the headline"
+ usa_economic_events.500.b: "Hold for yield and margin"
+ usa_economic_events.500.a_bonus: "High Frequency Process"
+ usa_economic_events.500.b_bonus: "Process Yield Discipline"
+ usa_economic_events.501.t: "Opteron and Sixty-Four Bits"
+ usa_economic_events.501.d: "Intel has bet on Itanium, a new instruction set that runs no existing software well, and AMD can instead extend x86 to sixty-four bits so everything already written keeps running. If it works, Intel has to license AMD's architecture back. If it does not, AMD has spent its engineering on a fork of a standard it does not control."
+ usa_economic_events.501.a: "Extend x86 to sixty-four bits"
+ usa_economic_events.501.b: "Follow Intel onto the new architecture"
+ usa_economic_events.501.a_bonus: "Instruction Set Extension"
+ usa_economic_events.501.b_bonus: "Compatible Architecture Licensing"
```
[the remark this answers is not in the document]
JasonBreen: Fixed in 8b9b560fed — relabelled to "Licensed Architecture Adoption". The old text described AMD licensing its architecture out, which is the opposite of what that option does.

common/ideas/united_states.txt
AngriestBird: This is not a valid modifier

common/ideas/united_states.txt
```diff
@@ -3094,6 +3094,72 @@ ideas = {
...
+			allowed_civil_war = { always = yes }
+			modifier = {
+				research_speed_factor = 0.02
+				civilian_chip_consumption_modifier = 0.02
+			}
+			research_bonus = { CAT_computing_tech = 0.02 }
+		}
+
+		USA_amd_legacy_fabrication_holdout = {
+			picture = construction
```
AngriestBird, line 3142: We really need to explore this about the fact that we have so many of these for the USA now that they also probably should looked at for balance. I think honestly one thing we may want to do as well is look into also integrating these with the regulatory content for Wall Street as Wall Street is very happy if you de-regulate the tech industry and would play a good meaningful integration with existing content. 

Maybe look at the pacing of this so there's a bit less?

AngriestBird (changes requested):
Few comments, content addition looks good but mostly something I was thinking about after reviewing a handful of these.

common/ideas/united_states.txt
JasonBreen: Checking this one before changing it — I think `equipment_conversion_speed` is valid, and I want to make sure I am not misreading which line you mean.

It is documented at `resources/documentation/modifiers_documentation.md:2583` as **Categories: country, war_production**, so country scope, which is where it sits here. It is also already used in five other places on main: `common/country_leader/CHI_traits.txt:111`, `common/decisions/Czech_Republic.txt:19`, `common/dynamic_modifiers/99_GRE_dynamic_modifiers.txt:67`, `common/dynamic_modifiers/99_PER_dynamic_modifiers.txt:74` and `common/ideas/!_special_project_ideas.txt:162`.

That line is also not from this PR — it is `USA_anduril_autonomous_arsenal`, which came in with #3561 and is on main already. Nothing in this branch touches it.

I did check my own additions against the same source rather than take that as read: this PR adds 8 distinct modifier keys and all 8 resolve against the 722 vanilla keys in the docs dump plus the 177 in `common/modifier_definitions/` — `research_speed_factor`, `consumer_goods_factor`, `political_power_gain`, `production_factory_efficiency_gain_factor`, `microchip_export_multiplier_modifier`, `microchip_plant_productivity_modifier`, `civilian_chip_consumption_modifier`, `industry_chip_consumption_modifier`.

If you meant a different line, or if that modifier is known not to work in an idea despite the docs, say which and I will fix it.

common/ideas/united_states.txt
JasonBreen: Agreed, and I would rather settle the direction with you than guess at it, so I have not changed the pacing yet.

The numbers, so we are arguing about the same thing: there are now **51 corporate-history outcome ideas across 12 USA chains** — amd 5, nvidia 5, spacex 5, then four each for motorola, microsoft, ibm, apple, anduril, micron, ti, intel and google. This PR adds 6 of them (the five AMD outcomes and the fifth NVIDIA one).

One thing that softens it slightly: they are mutually exclusive within a chain and all but three are 365-day, so the practical ceiling is 12 active at once rather than 51, and in most runs fewer. But 12 simultaneous economy modifiers that were never balanced against each other is still your point, and it stands.

On Wall Street — that is a better idea than another parallel outcome idea. Deregulation making the tech chains cheaper or faster, and the tech outcomes feeding Wall Street sentiment, would tie two systems together instead of running them beside each other. It is also a bigger change than this PR, and it would want the existing 45 outcomes revisited too, not just the 6 here.

So: do you want me to (a) hold this PR while that integration is designed, (b) cut it down to the AMD chain alone and drop the NVIDIA fifth outcome, or (c) land it as is and do the balance and Wall Street pass across all 12 chains as its own piece of work? I am happy with any of those — (c) is my preference only because it treats all 51 the same way rather than singling out the newest.

JasonBreen: Agreed, and I would rather settle the direction with you than guess at it, so I have not changed the pacing yet.

The numbers, so we are arguing about the same thing: there are now 51 corporate-history outcome ideas across 12 USA chains - amd 5, nvidia 5, spacex 5, then four each for motorola, microsoft, ibm, apple, anduril, micron, ti, intel and google. This PR adds 6 of them (the five AMD outcomes and the fifth NVIDIA one).

One thing that softens it slightly: they are mutually exclusive within a chain and all but three are 365-day, so the practical ceiling is 12 active at once rather than 51, and in most runs fewer. But 12 simultaneous economy modifiers that were never balanced against each other is still your point, and it stands.

On Wall Street - that is a better idea than another parallel outcome idea. Deregulation making the tech chains cheaper or faster, and the tech outcomes feeding Wall Street sentiment, would tie two systems together instead of running them beside each other. It is also a bigger change than this PR, and it would want the existing 45 outcomes revisited too, not just the 6 here.

So: do you want me to (a) hold this PR while that integration is designed, (b) cut it down to the AMD chain alone and drop the NVIDIA fifth outcome, or (c) land it as is and do the balance and Wall Street pass across all 12 chains as its own piece of work? Happy with any of those - (c) only because it treats all 51 the same way rather than singling out the newest.

common/ideas/united_states.txt
JasonBreen: You were right and I was not - I miscounted the line when I first checked, and I owe you a correction.

Line 2713 is `industrial_capacity_military`, not `equipment_conversion_speed`, and it is indeed not a modifier. The military factory output one is `industrial_capacity_factory`; there is no `_military` variant, so it compiled silently and did nothing. `validate_modifiers` confirms it as an unknown modifier once the file group is in scope.

It is mine, from #3561, and it is in six places rather than one: `common/ideas/Norwegian.txt:8,22` (Kongsberg), `common/ideas/Swedish.txt:7,21` (Saab) and `common/ideas/united_states.txt:2713,2727` (Anduril). All six are fixed in #3604 (`2b8b3006ff`) rather than here, since they are all on main already and none of them are in this PR.

`equipment_conversion_speed`, the line below, is fine - country scope in the docs and used in five other places - so nothing changed there.

I also re-ran `validate_modifiers` across the whole repo afterwards, with a positive control first (reintroduced the bad name, confirmed it was reported, reverted): zero unknown modifiers remain.

JasonBreen: Correction to my own correction: this landed in #3606 rather than #3604. I fixed the same six sites in parallel and the two changes were identical, so after merging main they cancelled out and no longer appear in #3604 diff. Either way `industrial_capacity_military` is gone from the mod and validate_modifiers is clean repo-wide.

TemplarGeneral: You probably want industrial_capacity_factory

AngriestBird:
Closing due to internal discussion about extending this. Holding to post 2.0.

common/ideas/united_states.txt
JasonBreen: Exactly right, and it is already fixed on main. `industrial_capacity_military` does not exist, so all six uses compiled silently and did nothing.

It was in three files rather than one - `common/ideas/Norwegian.txt` (Kongsberg), `common/ideas/Swedish.txt` (Saab) and `common/ideas/united_states.txt` (Anduril), two occurrences each - all from #3561. They landed on main as `industrial_capacity_factory` in #3606. `validate_modifiers` is clean repo-wide now, checked with a positive control first so the clean result means something.

None of it is in this PR, so nothing to change here.

PR #3580, opened 2026-09-05
Make the Isratine one-state branch reachable for the Israeli AI (#3478)

### LINKED ISSUE #3478: [FOCUS TREE] ISR_binational_state is permanently unreachable: nothing removes the_military for Israel
`ISR_binational_state` (`common/national_focus/05_israel.txt:135`) heads the 26-focus Isratine one-state branch. Its `available` is:

```
available = {
	has_stability > 0.7
	has_country_flag = ISR_big_size_knesset
	NOT = {
		has_idea = the_military
	}
	OR = {
		emerging_anarchist_communism_are_in_power = yes
		emerging_communist_state_are_in_power = yes
		neutrality_neutral_communism_are_in_power = yes
		emerging_autocracy_are_in_power = yes
		neutrality_neutral_muslim_brotherhood_are_in_power = yes
	}
}
```

Israel starts with `the_military` (`history/countries/ISR - Israel.txt`, in the opening `add_ideas` block) and **nothing anywhere in the mod removes it for ISR**. Every `remove_idea = the_military` site belongs to another country (Algeria, Belarus, Greece, Iran, Iraq, Bashkiriya, and two Iraqi decisions). The gate is therefore unsatisfiable for the whole game and the entire branch — `ISR_return_refugees`, `ISR_palestinian_army`, `ISR_federalism`, `ISR_join_arab_league`, the identity fork, and the rest — is dead content for player and AI alike.

Compounding it, the focus sat at a live `ai_will_do = { base = 3 }` over a reward that:

- `annex_country` on PAL and HAM,
- `change_tag_from` twice,
- `add_stability = -0.3`,
- and a `random_list` with a 50% `start_civil_war = { ideology = nationalist size = 0.2 }`.

#3477 sets the focus to `base = 0` so that the AI can never take it even if the availability gate is later loosened, but that is a safety measure, not a fix — it leaves the branch unreachable and is the sole source of the 16 orphan focuses that `tools/analysis/ai_path_report.py --tag ISR` now reports under `PEACE_PROCESS` and `NO_PATH` + historical AI off.

**What a real fix needs**, in whichever direction the design intends:

1. Give Israel a legitimate, AI-reachable route to shed `the_military` — a demilitarisation focus or decision on the Isratine branch's approach — and then re-own the branch and tame the civil-war roll; or
2. Drop the `NOT = { has_idea = the_military }` clause if the power-group gate was never meant to be absolute, and rebalance the reward; or
3. Retire the branch if it is not intended to be reachable.

Whichever is chosen, the `annex_country` + `change_tag_from` + 50% civil-war reward should be revisited before the branch is made reachable again.

TemplarGeneral:
### Changes
- The Israeli AI on the Peace Camp path can now rally Hadash, seat it in government and sideline the military faction, so the Isratine one-state branch and its 16 follow-up focuses are reachable instead of dead content
- Focus "A Binational State" no longer tries to annex Palestine or Hamas when those countries no longer exist
- Restored the branch root's AI weighting so it matches the rest of the branch instead of being permanently skipped

Notes for review:
- The three new decisions are AI-only and live in the existing ISR AI path category, so they carry no localisation by design.
- The branch stays owned by the Peace Camp path flag, as its 16 descendant focuses already were. A Peace Camp run may therefore now end up with a Hadash government instead of the Meretz/Labour two-state line. Splitting it onto its own game rule option was considered and deliberately left out of this PR.
- Left untouched, spotted while investigating: the four Isratine identity ideas cancel on the same supermajority condition the branch root requires (looks inverted), the five-ideology check is copy-pasted into all 17 branch focuses although a scripted trigger for it already exists, and an AI strategy block reuses the name of the ISR_befriend_libya focus.

Closes #3478


common/national_focus/05_israel.txt
```diff
@@ -174,16 +174,22 @@ focus_tree = {
...
-				ISR = { change_tag_from = ISR }
-				annex_country = {
-					target = HAM
-					transfer_troops = yes
+				if = {
+					limit = { country_exists = HAM }
+					annex_country = {
+						target = HAM
+						transfer_troops = yes
+					}
```
AngriestBird, line 189: This change tag setup is bugged and unneeded.

AngriestBird (changes requested):

PR #3586, opened 2026-09-05
Rebuild Romania's AI path game rule (#3162)

TemplarGeneral:
Part of #3162.

## Problem

`ROM_ai_behavior` existed but steered nothing. It set **two disagreeing flag layers** per option: an ideology bucket (`ROM_CENTER` / `ROM_FASCIST` / `ROM_DEMOCRATIC`) consumed by `ROM_random_ai`, which force-flipped the ruling party, and a party path flag (`ROM_PSD_PATH`, …) read by the seven strategy plans.

| Rule option  | Path flag                                       | Branch root needs | `ROM_random_ai` installed | Match   |
| ------------ | ----------------------------------------------- | ----------------- | ------------------------- | ------- |
| `HISTORICAL` | `ROM_PSD_PATH`                                  | 3                 | 18 or 14                  | no      |
| `CENTER`     | `ROM_PSD_PATH` / `ROM_PEASANTS_PATH`            | 3 / 1             | 18 or 14                  | no      |
| `DEMOCRATIC` | `ROM_PNL_PATH` / `ROM_PD_PATH` / `..._EUROPEAN` | 2 / 14-18         | 2 or 3                    | partial |
| `FASCIST`    | `ROM_GRP_PATH`                                  | 20                | 20                        | yes     |

Every party branch root is gated `<party>_are_in_power = yes`, so on the historical option the AI could never open the PSD spine its own plan boosted.

On top of that: the tree had **zero path ownership** (373 focuses, 0 path flags), only two bare `factor = 0 is_historical_focus_on = yes` killswitches that stranded 46 focuses under *any* option; `NO_PATH` left the AI with no plan at all; `ROM_european_plan` zeroed `ROM_pd`, its own route in; and `romania_politics.21` had no `ai_chance`, so a Vadim-path AI had a coin-flip chance of ceding Transylvania to TRA and then being annexed into WLC with every ROM core stripped.

## Changes

**Rule and loc.** `HISTORICAL` + five alt paths + `RANDOM_PATH` + `NO_PATH` as the `default = { }` block. PD and the True European Party merge into one option (`ROM_the_true_european_party` is a descendant of `ROM_pd` and installs the same government); the monarchy gains an option it never had despite a live `ROM_monarchist_plan`. Two-sentence descs, `§8…§!` on party names, header normalised to `@ROM Romania`.

**Wiring.** Flags renamed to `ROM_<PATH>_FOCUS_PATH` and set **only** by the rule. `ROM_random_ai` and the three bucket flags are deleted, along with the `clr_global_flag`/`set_global_flag` blocks in the seven party focus completion rewards, which would otherwise have zeroed their own branch's weights mid-run. `RANDOM_PATH` draws from all six buckets including historical.

**Ownership.** New `common/scripted_triggers/99_ROM_scripted_triggers.txt` (alt-guarded `ROM_ai_historical_path`, since it is read from strategy-plan `enable`, a decision-category `visible` and an event `trigger`). 262 of 373 focuses re-owned across seven groups via `apply_ai_path_weights.py`: historical 50, restored monarchy 72, greater romania 47, democratic party 30, peasant party 28, national liberal 27, anti-corruption fork 8. Economy (62), military (46) and the political header stay path-neutral. **0 orphans in every option × historical-AI state.**

**Reaching power.** Romania already owns its election flow behind `generic_election_killswitch`: `romania_elections.1` seats whichever party `party_pop_array` makes largest, and every branch root is blocked until it resolves. So the ramp is popularity-only — a new AI-only `ROM_ai_path_category` with five `ROM_rally_the_*` decisions, no forced `change_ruling_party_effect`, no loc keys. `romania_politics.1` / `.2` gained `ai_chance` so the monarchy path can re-earn `ROM_cdr_reformed`, which the 2000 election clears and which its branch root requires.

**Plans.** `ROM_PSD_plan` → `ROM_HISTORICAL_plan` on the triggers; `ROM_european_plan` folded into `ROM_PD_plan`; `ROM_unscripted_plan` added so `NO_PATH` still plays.

**Hardening.** `ai_is_threatened` across the 46-focus military trunk; `strength_ratio` guards on the five `annex_everything` focuses and `ROM_monarch_of_all_romanians`, which previously had nothing but the branch root between them and wars with Hungary, Bulgaria, Ukraine and Serbia; `ROM_avoid_starting_wars`, `ROM_cancel_war_neighbours` and per-target readiness gates in `common/ai_strategy/ROM.txt`; `ROM_support_mlv`'s `allowed = { original_tag = SOV }` corrected to `ROM`. Eight mutex ties broken by base preference.

## Verification

- `ai_path_report.py --tag ROM`: rule and wiring **clean**, 0 orphans in all 14 states, no additive path modifiers, no unreferenced flags, `mechanics` **clean**, `government` still *undated successor roster / walker none* (no walker written, per the standard).
- `validate_focus_tree.py --path .`: **no issues**.
- `validate_decisions.py`: 438 warnings in 68 files, identical to the pre-branch baseline; the new AI-only decisions add none.

## Judgement calls to flag

- **`ROM_mineriad_legacy` is not a defect.** The report lists it under *nothing relieves*, but `history/countries/ROM - Romania.txt:630` adds it with `add_timed_idea = { days = 1095 }` — it lapses on its own. The other four entries are bonuses or carriers: `un_aid_subsidies_modifier` is the shared UN aid container, `ROM_armed_forces_modifier` / `ROM_romanian_economy_modifier` are Romania's own dynamic-modifier carriers whose negative backing variables each have 3–16 focus cures, and `ROM_corruption_modifier` is relieved through `ROM_corruption_decrease` (`99_ROM_scripted_effects.txt:3`), which the report cannot see through the scripted-effect indirection.
- **Three mutex ties remain by design.** `ROM_economic_patriotism` / `ROM_nordic_approach` and `ROM_king_paul` / `ROM_queen_margareta` are the losing pairs of three-way forks whose preferred sibling now sits at a higher base; `ROM_monarchist_victory` / `ROM_republican_victory` are force-completed by `romania_politics.9`, partitioned by mutually exclusive triggers, and the monarchist plan already zeroes the republican side.
- **`romania_politics.22` is left as-is.** Both of its options end at `.23`, which annexes Romania regardless, so the guard has to sit at `.21` — and does.
- **`ROM_destroy_the_magyars`, `ROM_retake_southern_dobruja` and `ROM_unite_banat` were dropped from `ROM_GRP_plan`'s `ai_national_focuses`.** They stay takeable at their tree weights behind the new strength guards; they are simply no longer prioritised, since both targets are NATO members from 2004.
- Non-English `.yml` files are out of scope per `AGENTS.md`; the retired `RULE_OPTION_ROM_FASCIST` / `_DEMOCRATIC` / `_CENTER` keys still exist there and degrade gracefully.


common/ai_strategy/ROM.txt
```diff
@@ -88,3 +88,117 @@ ROM_cascadia_help = {
 	ai_strategy = { type = send_volunteers_desire id = "CAS" value = 1000 }
 	ai_strategy = { type = send_lend_lease_desire id = "CAS" value = 400 }
 }
+
+ROM_avoid_starting_wars = {
+	allowed = { original_tag = ROM }
+	enable = {
+		has_war = yes
```
AngriestBird: Duplicate AI strategy with the global one

common/ai_strategy/ROM.txt
```diff
@@ -88,3 +88,117 @@ ROM_cascadia_help = {
...
+	}
+	abort_when_not_enabled = yes
+
+	ai_strategy = { type = avoid_starting_wars value = -200 }
+}
+
+ROM_cancel_war_neighbours = {
+	allowed = { original_tag = ROM }
+	enable = {
+		has_war = yes
```
AngriestBird: Unneeded with the global?

AngriestBird (changes requested):
Generally looks good. Duplicate AI strategies.

common/ai_strategy/ROM.txt
MrP0tter: Removed as duplicate

common/ai_strategy/ROM.txt
MrP0tter: resolved

AngriestBird (approved):

PR #3589, opened 2026-09-05
Use token tables for ruling-party drift

### LINKED ISSUE #3414: Pick ruling-party drift ideas from a token table
**Context**

`add_own_ideology_drift` in `common/scripted_effects/00_ideology_drifts.txt` is a 5-way `has_government` × major/minor dispatch onto `ruling_party_western` / `major_ruling_party_western` (and nationalist/emerging/non_alligned/salafist). Clear is ten separate `has_idea` removes. The ideas themselves are 0.02 vs 0.04 drift on that outlook.

Same token fold as #3405 / #3408. Keep the ten idea objects. Fascism still also adds `sunni` for non-NIG; that gate stays outside the table.

**What**

- Seed `global.ruling_party_idea^outlook` and `global.major_ruling_party_idea^outlook` at startup
- One helper: pick major vs minor, `add_ideas` from `var:token`, skip if already applied
- Clear previous token on government change
- No `meta_effect` on this path

**Acceptance**

- Same 0.02 / 0.04 drift per outlook and major flag as today
- `sunni` still applied on the fascism non-major non-NIG path
- No 5-way `has_government` in `add_own_ideology_drift`

Blazer135:
Closes #3414

**What**
Select ruling-party drift ideas from major/minor outlook token tables and skip unchanged ideas.

**Why**
Replace repeated government dispatch and ten separate removal checks while preserving 0.02/0.04 drift and the fascist, non-major, non-NIG Sunni exception.

**How**
Seed tables before country startup, reuse the numeric outlook helper, and track the applied token for replacement on government changes. The existing idea objects and modifiers remain intact. In-game token resolution has not been verified.

The unused-idea validator recognizes literal `token:` references, with regression coverage, so ideas stored in these tables are counted as used. Content CI uses this correction after it reaches `main`.


Blazer135:
Validators for the Ideas are False Positives

AngriestBird (approved):

Issue #3393, opened 2026-09-02
Add an outlook lookup for party-index range checks


AngriestBird:
**Context**

Outlook membership is five uneven ranges (democratic 0-3, communism 4-9, fascism 10-11, neutrality 12-19, nationalist 20-23). That mapping is copied as `< 4` / `< 10` / `< 12` / `< 20` ladders in `change_relative_party_popularity`, `steal_from_party`, `add_up_party_totals`, `set_coalition_drift`, and `change_ruling_party_effect`.

The older politics effects do not all resolve variable-carried ideology tokens. In particular, `set_politics` needs a literal ideology when changing to a different government. Keep ideology arguments static and share only the numeric lookup and calculations.

Do not regroup ideologies. Do not fold this into #3392. The lookup should be the single place that the later 1-index shift updates.

**What**

Calculate a numeric outlook ID and start/end bounds from a party index. Use numeric loops, arrays, and math for shared calculations. Keep one literal dispatch for each old effect that requires a static ideology.

**Acceptance**

- Same slot-to-outlook mapping as today
- `change_relative_party_popularity` has one remainder-rescale block
- `steal_from_party`, `add_up_party_totals`, `set_coalition_drift`, and `change_ruling_party_effect` use the numeric lookup
- No variable-carried ideology is passed to `set_politics`, `add_popularity`, or `party_popularity@...`
- Slot 10 remains fascism with elections enabled or disabled
- Party totals remain nonzero when their party arrays are nonzero

PR #3591, opened 2026-09-05
Select coalition drift ideas from token tables

### LINKED ISSUE #3407: Pick coalition outlook ideas from a token table
**Context**

Coalition partner share is already a scalar per outlook (`coalition_democratic`, `coalition_communism`, …) in `set_coalition_drift`. That scalar is then quantized into nine mutually exclusive ideas per outlook (`coalition_western_1`…`_9` and the emerging/salafist/nationalist/non-aligned copies) in `common/scripted_effects/00_ideology_drifts.txt` plus `common/ideas/generic_ideas.txt`.

The idea ladders are linear: level N is `*_drift = 0.02 * N` and `*_acceptance = 5 * N`. `clear_coalition_*_count_ideas` / `reset_*_coalition_ideas` are five copy-pasted 9-way `else_if`s.

Fold with tokens, same pattern as #3405. Keep the idea objects. Do not add `meta_effect` to build `coalition_western_[N]` on a hot path. A later dynamic-modifier pass is optional if we want to delete the 45 ideas.

**What**

- Seed five token arrays at startup (`global.coalition_western_idea^1` = `token:coalition_western_1`, …)
- `set_coalition_drift` maps each outlook scalar to a level, then `add_ideas` / `remove_ideas` from `var:` of that token
- Skip the write when the country already has that idea
- Delete the clear/reset `else_if` ladders
- After #3393 the range `if`s that fill the outlook scalars become the lookup

**Acceptance**

- Same drift and acceptance as the current 0.01 / 0.07 / 0.11 / … cut-points
- No 9-way `else_if` that `add_ideas = coalition_western_N`
- No `meta_effect` on the coalition-drift path

Blazer135:
Closes #3407

**What**
Select coalition drift ideas through five token tables and one shared threshold calculation, retaining all 45 ideas and their modifiers.

**Why**
Remove the repeated nine-level add/remove ladders and avoid replacing unchanged coalition ideas.

**How**
Preserve strict cut-points and the numeric outlook aggregation from #3393. Track applied tokens for replacement and election cleanup; remove redundant clears in coalition decisions, election events, and two Brazilian focuses. No runtime meta substitution is added. In-game token resolution has not been verified.

Pass explicit temporary inputs to the shared coalition helpers. The unused-idea validator recognizes literal `token:` references, with regression coverage, so ideas stored in these tables are counted as used. Content CI uses this correction after it reaches `main`.


Blazer135:
False Positive Validator for the 45 ideas

AngriestBird (approved):

PR #3594, opened 2026-09-05
Refactor spending-law cost ladders onto a shared helper (#3409)

### LINKED ISSUE #3409: [TASK] One helper for spending-law cost ladders
**Context**

`update_bureaucracy_rate`, `update_police_rate`, `update_health_rate`, and `update_social_rate` in `common/scripted_effects/00_money_system.txt` are the same loop: resize a cost array, `population_total * (1 + modifier) * common_money_modifier * per_capita * global.spending_ladder_* ^i`, then index by `*_law - 1`. Education and defence already diverge (research slots, unique ladders). Expected-spending clamps in `00_expected_spending_effects.txt` are the same war/offset/round/clamp shape six times.

This is monthly for every country. Fold the identical arithmetic, not the unique modifiers.

**What**

- One scripted effect parameterized by ladder array, per-capita constant, cost modifier, law index, output var, and array size
- Bureau / police / health / social call it
- Leave education, military, and infra as their own functions
- Optional same pass: one expected-spending clamp helper for the six `calculate_expected_*_spending` twins (war -1, add modifier and offset, round, clamp). Distinct from #3402 (GUI frames)

**Acceptance**

- Same daily costs at each law level as today, including police CT extra
- Bureau/police/health/social no longer duplicate the loop body
- Monthly money tick still writes `bureaucracy_gain` / `security_gain` / `health_gain` / `welfare_gain`

TemplarGeneral:
### Changes
- Bureaucracy, police, healthcare and welfare spending ladders now share one `calculate_spending_law_base_cost` helper: the `population_total * cost multiplier * common_money_modifier * per-capita` product is computed once per department instead of once per law level, so each ladder entry costs a single multiply on every economic tick
- Each of the four effects keeps only what actually differs: its own cost array, spending ladder, array size, law index and gain variable. Police keeps its counter-terror surcharge (`ct_police_cost_var`) in its multiplier
- The five identical expected-spending calculations (welfare, healthcare, education, police, administrative) now share `calculate_expected_law_spending` for the war / modifier / desire-offset / round / clamp shape

### Notes
- Parameterization goes through temp variables with a `# Parameters:` contract, matching the rest of the mod and picked up by `validate_scripted_params.py`, rather than the `$PARAM$` macro form the issue sketches: nothing in vanilla or MD uses macros in `common/scripted_effects/`
- The multiply order inside the helper is the order of all four originals, so costs at every law level are unchanged. Array names are untouched, so the per-level law tooltips in `MD_laws_l_english.yml` keep working
- Education, military and infrastructure keep their own bodies (research slots, facilities, unique ladders, war-scaled clamp) and are not touched
- One behaviour change, debug-only: the per-department `Expected ... spending` log now prints after the clamp instead of between the round and the clamp, so it reports the value actually used
- Possible follow-up, deliberately out of scope: the four `array_*_spend` arrays are read only by law tooltips, so filling all 5-6 entries for AI countries every tick is wasted work

Closes #3409

### Test plan

**Playthrough A: USA 2000 (covers #3409)**

This is a pure refactor, so every check is a comparison against a same-date `main` run. Take the `main` reference numbers first.

Game start, before unpausing:

- [ ] `tag USA`, open the economy law screen and hover all 5 bureaucracy, 5 police, 6 healthcare and 6 welfare levels, confirm each "$X billion is being spent" figure matches the `main` reference exactly (#3409).
- [ ] Confirm each of the four ladders is still monotonically increasing from level 1 to its top level, with no zeroed or duplicated entries, which would indicate the dropped `value = v` broke the per-level indexing (#3409).
- [ ] Confirm the current-level expense rows on the budget screen (bureaucracy, internal security, healthcare, welfare) match `main` to the same decimals (#3409).
- [ ] Confirm education, military and infrastructure spending are unchanged, they keep their own bodies and must not be affected (#3409).
- [ ] Check `error.log` for math-expression or unknown-variable errors from `00_money_system.txt` and `00_expected_spending_effects.txt`, there must be none (#3409).

Parameter isolation (the four ladders run back-to-back in one `ingame_update_setup` pass and share temp-variable names):

- [ ] Pick a country whose health and welfare cost multipliers differ, confirm the healthcare and welfare ladders keep their distinct per-capita bases rather than converging on whichever department ran last (#3409).
- [ ] Trigger the counter-terror path that raises `ct_police_cost_var`, confirm all 5 police levels get more expensive and bureaucracy, healthcare and welfare are untouched (#3409).
- [ ] Add a modifier that changes only `health_cost_multiplier_modifier`, advance one economic tick, confirm only the healthcare ladder moves (#3409).
- [ ] Advance one economic tick with the counter-terror surcharge active and then clear it, confirm the police ladder returns to its baseline values instead of keeping the surcharge (#3409).

Scale and edge cases:

- [ ] `tag CHI` (or IND), set the top law level in each of the four departments, confirm the ladder values still clamp at the same ceiling as `main` and no entry reads 0 or a garbage number (#3409).
- [ ] `tag` a micro nation (for example VAT or SEY), confirm all four ladders read near-zero, never negative, and the law screen renders without blanks (#3409).
- [ ] Trigger a civil war split and check the new tag, confirm the ladders populate correctly on the first economic tick after the split (#3409).

Expected spending (advance one month with `debug` enabled):

- [ ] Confirm the `Expected welfare / healthcare / education / police / administrative spending` log lines each appear once per tick and now print the post-clamp value inside 1 to 6 (welfare, healthcare) or 1 to 5 (education, police, administrative) (#3409).
- [ ] Confirm the persistent `expected_welfare_spending`, `expected_health_spending`, `expected_edu_spending`, `expected_police_spending` and `expected_adm_spending` variables actually change over time, the helper writes a temp variable and each wrapper copies it back, so a broken copy-back would freeze them (#3409).
- [ ] Change a desire offset (for example via a party shift affecting `bureau_desire_offset`) and confirm only the administrative expectation moves (#3409).
- [ ] Enter a war, confirm every one of the five expectations drops by 1 relative to the peacetime value and still clamps at a minimum of 1 (#3409).
- [ ] Compare all five expectation values against the `main` reference run on the same date, they must be identical (#3409).

Mid-game AI checks (observer, advance to 2005):

- [ ] Confirm AI countries still raise and lower the bureaucracy, police, healthcare, welfare and education laws, with no country frozen at level 1 or at its maximum (#3409).
- [ ] Spot-check three AI nations of different sizes, confirm their four ladder arrays and current expense rates match a `main` observer run at the same date (#3409).
- [ ] Confirm no new spending alerts fire that did not fire on `main`, `md_check_spending_alerts` reads the same variables (#3409).


TemplarGeneral:
The problem is the expected spending of any law is red and showing -
besides military for some reason

PR #3706, opened 2026-09-07
Clear outstanding event and style validator warnings

TemplarGeneral:
Clears every outstanding warning in the event and style validators. The reported warnings turned out to be a mix of real content gaps and validator false positives, so this fixes both sides.

## Result

| Category | Before | After |
| --- | ---: | ---: |
| `event-option-log-without-effect` | 7 | 0 |
| `unreferenced-triggered-only` | 3 | 0 |
| `event-log` | 73 | 0 |
| `focus-standards` | 3 | 0 |
| `spacing` | 1 | 0 |

Both validators now report `NO ISSUES FOUND`.

## Tooling

**Scope-keyed effects were invisible to the option log detector.** `_OPTION_STATEMENT_RE` only matched keys starting with a letter, so an option whose sole effect was a numeric state scope (`652 = { ... }`) or a quoted tag scope (`"LGN" = { ... }`) recorded no statement at all and read as effect-free. That produced all 7 `event-option-log-without-effect` warnings against options doing real work — damaging buildings, running the Moscow mayoral elections.

This one matters beyond the warning count: `tools/linting/fix_event_option_logs.py` shares the detector, so running it on `main` would have **deleted seven correct log lines**. It is now a verified no-op. The regex change is a strict reduction — 7 warnings to 0, none introduced.

**`lar_collab_gov.1` whiteflagged.** The check had no exemption mechanism; the only escape was the dynamic-namespace rule. It is the vanilla La Resistance event behind the live `operation_collaboration_government` system, dispatched engine-side with no script reference to find, so it now sits in a module-level exempt set.

**Focus-standards bucket stopped swallowing event warnings.** The filter tested for `"focus"` anywhere in the message, which also matches event options whose own name carries it — `LibyaFocus.104.a`, `SyriaFocus.54.b`, `TurkeyFocusEvents.56.a` were each reported twice. It now matches the `Focus ID ...` message the check actually emits. Kept as its own commit since it is an independent bug.

Each tooling commit carries regression tests.

## Content

**73 missing option logs** across 26 event files. Each log uses the option's own name, matching the convention in the surrounding files.

One correction worth flagging: the warning for `Czech Republic.txt` names `CZE_642_name`, which is not an option at all — the style check's name scrape re-matches on the later `set_state_name = CZE_642_name` line and overwrites the real name. The option is `CZE_czech.31.o1`, and that is what the log says. It was the only such mismatch among the 73.

**`russia.17` and `russia.63` removed.** Both triggered-only and fired by nothing. Their payloads were re-implemented inline in the focuses that are the live path: `SOV_special_o` carries the special-forces variable block, `SOV_pmc_gazproms` the Gazprom PMC template and its three `create_unit` blocks. The focus versions are supersets — they add the party popularity block, `treasury_change = -15` and the create tooltip — so nothing is lost. Their six English loc keys go with them; non-English files are left alone per AGENTS.md.

**One stray tab** before `=` in the Algerian welfare focus.

## Note on the warning counts

The report this started from listed 53 `event-log` warnings. `main` actually carries 73 — the per-file counts matched exactly, the list was just truncated. All 73 are fixed.

## Testing

- `python -m pytest` — 4442 passed, 22 skipped
- `validate_events.py` and `validate_style.py` — both clean
- `fix_event_option_logs.py --dry-run` — 0 lines to remove across 167 files
- `check_common_mistakes.py --mode diff` — passed

No changelog entry, per AGENTS.md.


common/national_focus/05_algeria.txt
```diff
@@ -11147,7 +11147,7 @@ focus_tree = {
 			newline = yes
 			set_temp_variable = { corp_change = -2 }
 			modify_corporate_tax_rate_effect = yes
-			newline	= yes
```
AngriestBird, line 11150: Create an issue for this so we can have the fix_styling actually runnable to fix this so we don't have to run it manually

AngriestBird (approved):

common/national_focus/05_algeria.txt
TemplarGeneral: Filed as #3720. Note it needs more than flipping the hook stage: `fix_styling --mode all --dry-run` would rewrite 11,897 lines across 1,191 files today (mostly collapsing aligned whitespace before inline comments) while `validate_style` reports exactly 1 spacing warning repo-wide, so it has to be narrowed to the validator's actual warning surface before it can run on commit without dragging whitespace churn into every diff.

PR #3711, opened 2026-09-07
Currency effect display changes

JohnsonvilleF221:
### What was broken

The reserve currency laws showed only the number of nations using each currency. The effects that number produces are delivered through inflation_dynamic_modifier, so they appear under the inflation tooltip with nothing connecting them to the currency picked. At a standard start the US Dollar is pinned at the +3% ROI cap — a 50% lift on the 0.06 base roi_rate — and none of that is visible where the choice is made.

reserve_roi_bonus also holds a single value, for whatever currency the country already holds, so the other seven entries had no figures of their own to show.

### The fix

calculate_reserve_projections (00_currency_effects.txt) precomputes return on investment, trade opinion and inflation cost for all eight currencies into globals, so any currency can be read before it is adopted. The figures depend only on issuer currency_strength and adoption count, both global, so it runs once per month from the existing on_monthly guard rather than per country, plus a startup seed once the issuers are set.
Each law entry now carries one tooltip: those live figures followed by that currency's flat modifiers.
The flat modifiers move into hidden_modifier and still apply. The engine sorts real modifiers itself and appends custom_modifier_tooltip last, so any visible modifier pushed the readout below it; neither declaration order nor the order of two separate custom_modifier_tooltip entries changes that.
Flats that overlap the displayed figures (EUR/CNY/JPY/NLG) are folded into the live totals rather than listed twice. The yuan reads its statics from CHI's focus progress, since its three tiers share one tooltip.

### Notes

No balance changes. Every existing modifier value is untouched.
English localisation only; non-English is left to Paratranz.
Cost is roughly 350 statements per month, against the ~26,000 calculate_currency_strength already spends across 393 countries on the same tick.
**_Trade opinion reads a flat 5.00% because reserve_trade_bonus uses bare min/max operators rather than clamp_** **_(00_currency_effects.txt:447-448), which pins it at 0.05. Left as-is here to keep this PR display-only._**

PS — I did notice while working on this that when focus "Yuan Internationalisation" is taken as China, the sprite for the yuan disappears and it also doesn't add all the nations that were on the yuan to it. I might have already made a bug report by the time you're reading this though.

### Before
 
Screenshot 2026-09-07 130811
Screenshot 2026-09-07 130830

### After

Screenshot 2026-09-07 203736
Screenshot 2026-09-07 203721

Oh, and I also added myself as an author 😊


common/scripted_effects/00_currency_effects.txt
```diff
@@ -458,3 +458,225 @@ adopt_chinese_yuan_reserve = {
...
+# pay a holder — so a law entry can be read before it is adopted.
+#
+# The result depends only on the issuer's currency_strength and that currency's
+# adoption count, both global, so the values are identical for every country and
+# are computed once per month rather than once per country.
+#
+# Everything here reads and writes globals only, so it is scope-independent: it
+# runs from the on_monthly global guard and from the once-only startup block.
+
+capture_issuer_strength = {
```
AngriestBird: inline the set variable. The indirection isn't needed here since it's a single line scripted effect caller. It's the same work either way.

common/scripted_effects/00_currency_effects.txt
```diff
@@ -458,3 +458,225 @@ adopt_chinese_yuan_reserve = {
...
+	set_temp_variable = { tmp_static_infl = 0 }
+	if = { limit = { check_variable = { global.nlg_issuer > 0 } }
+		var:global.nlg_issuer = { capture_issuer_strength = yes }
+		set_temp_variable = { tmp_roi_mult = @nlg_roi_curve }
+		set_temp_variable = { tmp_adopters = global.nlg_reserve_adoption_count }
+		apply_reserve_projection = yes
+	}
+	else = { zero_reserve_projection = yes }
+	add_reserve_statics = yes
+	set_variable = { global.proj_roi_nlg = global.proj_roi_tmp }
```
AngriestBird: Quite a bit of indirection here and I think with a few more passes we can try to reduce this. I recommend scripted effect parts of this and then rebuilding it into a temp variable system. Be leery of using global and set variables as it saves permanent data that we don't nesscarily need. 

I also think we can seed these values in the generalized currency and not sure we need to nesscarily do this in here. Have we done a pass here to ensure this is as simple as we can make it?

localisation/english/MD_currency_l_english.yml
```diff
@@ -29,14 +29,25 @@
...
- usd_reserve_users_tt: "Nations Using This Currency: [?global.usd_reserve_adoption_count|0Y]"
- eur_reserve_users_tt: "Nations Using This Currency: [?global.eur_reserve_adoption_count|0Y]"
- cny_reserve_users_tt: "Nations Using This Currency: [?global.cny_reserve_adoption_count|0Y]"
- jpy_reserve_users_tt: "Nations Using This Currency: [?global.jpy_reserve_adoption_count|0Y]"
- rub_reserve_users_tt: "Nations Using This Currency: [?global.rub_reserve_adoption_count|0Y]"
- gbp_reserve_users_tt: "Nations Using This Currency: [?global.gbp_reserve_adoption_count|0Y]"
- chf_reserve_users_tt: "Nations Using This Currency: [?global.chf_reserve_adoption_count|0Y]"
- nlg_reserve_users_tt: "Nations Using This Currency: [?global.nlg_reserve_adoption_count|0Y]"
+ # One tooltip per currency. It MUST be a single loc key: the engine ignores the order of
+ # custom_modifier_tooltip entries within a modifier block, so two keys cannot be ordered.
```
AngriestBird: Comment is a bit verbose and isn't needed here.

Changelog.txt
```diff
@@ -1583,3 +1583,4 @@ v2.0.0
   - Removed the two-page investment building selector in favor of a single grid showing all building types
   - Repositioned and decluttered the energy view and state view (productivity text, victory points, controller flag, population, resistance/compliance, building-modifier icon), and added a Build Enrichment Facility button to the energy view
   - Standardized buttonType casing to ButtonType across the politics view GUI
+  - Reserve currency laws now show what each currency gives you: its live return on investment, trade deal opinion and inflation cost, recalculated monthly for every currency rather than only the one you hold, followed by that currency's flat modifiers. Previously the tooltip listed only the number of nations using the currency, and its effects appeared under the inflation modifier with nothing connecting them to the choice
```
AngriestBird: Doesn't need a changelog -- Reserve currencies are new to stable which is more orl ess what should be in the changelog.

AngriestBird (changes requested):

AngriestBird (approved):

Issue #3777, opened 2026-09-08
[BUG] Game Rules | MD drops vanilla's engine-read Peace Score and Scorched Earth rules


TemplarGeneral:
Split off from #3769.

**What happens**

MD's `common/game_rules/00_game_rules.txt` carries vanilla's filename and replaces it wholesale, so every vanilla rule id MD does not redefine stops existing. Four of the 43 dropped ids are **engine-read** and still functional in a 2000-present setting, so the toggles are simply gone from MD's rules screen and the engine falls back to its hardcoded defaults:

- [ ] `allow_scorched_earth` (vanilla `common/game_rules/00_game_rules.txt:481`, group `RULE_GROUP_GAMEPLAY`, `required_dlc = "No Step Back"`)
- [ ] `peace_score_to_overlord`
- [ ] `peace_score_to_faction_leader`
- [ ] `peace_score_reset_low_scores`

The last three are the whole `RULE_GROUP_PEACE_SCORE` group ("Peace Score distribution"), so MD players have no control over how peace-conference score flows to overlords and faction leaders.

**Why these four and not the other 39**

The other dropped ids are WW2-only (`*_fragmentation_status`, `*_colonization_status`, vanilla ideology path selectors) and correctly dead in MD. These four have no `has_game_rule` reader anywhere in vanilla script, which is what marks them engine-read: the engine consumes them directly, so dropping the definition removes a working option rather than a dead one.

`git log -S` shows none of the four was ever in MD's file. They were not removed on purpose, MD's copy simply predates the NSB and peace-score patches.

**Fix**

Copy the four blocks verbatim from vanilla into MD's file:

- `allow_scorched_earth` into the `# MILITARY RULES` section (next to `allow_paratroopers`, same group).
- The three `peace_score_*` rules into the `# PEACE DEAL RULES` section, next to `allow_chaotic_PC` / `allow_player_led_PC` / `allowed_CPD`.

[...]

PR #3778, opened 2026-09-08
Add POR and LIT AI behavior game rules

### LINKED ISSUE #3769: [BUG] Game Rules | MD's game_rules override silently deletes vanilla rules
**What happens**

Loading a saved gamerule preset throws `invalid rule id` for rules that exist in vanilla but not in MD, so the preset silently loses those settings.

```
[persistent.cpp:67]: Error: "invalid rule id: POR_ai_behavior, near line: 14
Unexpected token: POR_ai_behavior, near line: 14" in file: "semi random.txt" near line: 14
```

**Where**

MD ships `common/game_rules/00_game_rules.txt` under the **same filename as vanilla's** `common/game_rules/00_game_rules.txt`, so it replaces the vanilla file wholesale rather than adding to it. Every rule id that only existed in vanilla ceases to exist, and any preset referencing one fails to parse.

- [ ] `POR_ai_behavior` — defined in vanilla at `<HOI4>/common/game_rules/00_game_rules.txt:1840`. MD has no `POR_`/`PRT_` rule at all.
- [ ] `LIT_ai_behavior` — defined in vanilla at `<HOI4>/common/game_rules/00_game_rules.txt:2068`. MD defines **zero** `LIT_*` rules.
- [ ] Decide the policy going forward: either give MD's file a distinct name so the vanilla rules survive alongside it, or deliberately re-add the country `*_ai_behavior` rules MD wants to keep. For reference, MD's only German rule is `GER_ai_behavior` (`common/game_rules/00_game_rules.txt:2627`).

**Additional context**

The same preset also failed on `GER_focus_tree_selection`, `LIT_focus_tree_selection` and `player_peace_setting`. Those are **not** MD's problem — none of the three has any trace in MD or in vanilla, so the preset was written with an extra mod in the playset. Recording them here only so the log lines are accounted for. MD's closest peace-related rule is `allow_player_led_PC` (`common/game_rules/00_game_rules.txt:153`); the others are `allow_chaotic_PC` (`:138`), `allowed_CPD` / `allowed_CPD_AI` (`:168` / `:185`) and `allow_forcing_a_white_peace` (`:1428`).

Full log output:

```
[persistent.cpp:67]: Error: "invalid rule id: GER_focus_tree_selection, near line: 5 ..."
[persistent.cpp:67]: Error: "invalid rule id: LIT_focus_tree_selection, near line: 6 ..."
[persistent.cpp:67]: Error: "invalid rule id: POR_ai_behavior, near line: 14 ..."
[persistent.cpp:67]: Error: "invalid rule id: player_peace_setting, near line: 18 ..."
```

**Versions**

Reported from a dev-branch playthrough (2000-01 → 2007-09).

**To Reproduce**

1. Save a gamerule preset in a playset where the vanilla `POR_ai_behavior` / `LIT_ai_behavior` rules are visible.
2. Load that preset with MD enabled — the rules fail to parse and their settings are dropped.

**Operating System**

- [x] Windows
- [ ] Linux
- [ ] Mac

TemplarGeneral:
Closes #3769

#### Bottom line

- `POR_ai_behavior` and `LIT_ai_behavior` exist again, so game-rule presets that reference them stop failing with `invalid rule id`.
- The override policy for `common/game_rules/` is written down in `.claude/docs/game-rules-reference.md`.

#### Why

- MD's `common/game_rules/00_game_rules.txt` carries vanilla's filename, so it replaces vanilla's file wholesale and 43 vanilla rule ids stop existing. A preset referencing one fails to parse and silently drops that setting.
- 39 of those ids are WW2-only (`*_fragmentation_status`, `*_colonization_status`, vanilla ideology path selectors) and correctly dead in MD, so the filename override stays. Renaming MD's file would reload vanilla's, duplicating the 43 ids MD overrides and putting the WW2 rules back on the rules screen.
- Nothing recorded which vanilla rules MD must keep mirroring, which is how these two went missing in the first place.

#### How

- `common/game_rules/00_game_rules.txt`: `LIT_ai_behavior` and `POR_ai_behavior` added in the `# AI FOCUS BEHAVIOUR` section, in its alphabetical-by-country-name order, each with a `HISTORICAL` option and the `NO_PATH` default.
- No `999_game_rules_on_actions.txt` wiring: those blocks set `TAG_*_FOCUS_PATH` flags read by a country's focus tree, and neither country has an MD tree yet. The wiring lands with the tree.
- `MD_game_rules_l_english.yml`: option keys only. Vanilla already names both tags, matching how POL, ROM and MNT are handled.
- `.claude/docs/game-rules-reference.md` (indexed in `documentation-references.md` and `AGENTS.md`): why the filename override is deliberate, which vanilla rules stay dropped, which must be mirrored, and the diff recipe to re-check after a HOI4 version bump.

#### Not fixed here

- The same log also shows `GER_focus_tree_selection`, `LIT_focus_tree_selection` and `player_peace_setting`. Those are The Road to 56's rule ids, not vanilla's and never MD's, so MD cannot define them without claiming another mod's rules. They clear by re-saving or deleting that preset with MD alone.
- #3777 tracks the four vanilla rules that are engine-read and still missing from MD: `allow_scorched_earth` and the three `peace_score_*` rules.


AngriestBird:
General did a fucky wucky and forget this was for Road to 56

PR #3819, opened 2026-09-08
Avoid redundant state scans in economy updates

AngriestBird:
## Bottom line
Remove three state scans from economy updates.

- Use the live owned-state count for the capital repair guard.
- Reuse freshly calculated missile-silo and anti-air totals instead of summing each twice.
- Leave other state checks intact where ownership, per-state conditions, or refresh timing differ from country totals.

BLUF


AngriestBird:
Ended up cost more then saved so far -- closing for now

PR #3832, opened 2026-09-08
Event Picture Format Validation (#3829)

### LINKED ISSUE #3829: [VALIDATION] Event Picture Validator
If possible we should expand the Event or GFX validator to check if a picture with the news format, is put to a news_event and one with a country_format is added to a country_event and that no picture is added to hidden = yes events

This will prevent having wrong picture types in events

TemplarGeneral:
### Changes
- Adds a validator check for event pictures using art authored for the other event window, so wide news art on a country event or narrow country art on a news event is caught before merge instead of showing up as an overflowing or under-filled picture frame in game
- Adds a validator check for pictures on hidden events, which open no window and never draw them
- Removes 17 dead picture fields from hidden events in Counter Terror, Iran, Space_GFXlaunches and United States

Closes #3829


TemplarGeneral:
Why is the validator not showing this event?

image

AngriestBird (approved):

Issue #3332, opened 2026-08-31
[TASK] Events | Remove the self-referencing events


AngriestBird:
**Description**
A clear and concise description of the task. The main goal of these tasks are general purpose improvements.

Self referencing events need to be converted into a new flow so they are not referencing themselves. These event types are very bad for performance and need to be redone. On actions are preferable. 

**Additional context**
Add any other context or screenshots here.

PR #3854, opened 2026-09-09
Remove self-referencing events and detect them in CI (#3332)

JonathanWoollett-Light:
No event may fire itself. A self-fire parks a re-arming entry in the engine's event queue: the recurrence lives in the save rather than in script, every country that ever entered the loop keeps paying for it, and nothing outside the event can stop it.

Detection. Self-reference is the length-1 case of a cycle in the event fire graph, so one pass finds both. `scan_event_fire_edges` builds parent-to-child edges with call-site lines (`scan_event_fire_graph` now delegates to it and keeps dropping self-edges for the date-gated scheduling check), and `_iter_fire_cycles` enumerates cycles through stdlib `graphlib.TopologicalSorter` — no new dependency. Two checks: `self-referencing-event` (ERROR, always on) and `event-fire-cycle` (WARNING, opt-in behind `--fire-cycles`, 57 on main). Multi-event cycles stay opt-in because they are usually deliberate — a paginated browse screen, a hub with a back option — and nothing separates those from a poll statically.

Conversions. All 49 self-referencing events (61 call sites) are gone: 19 recurrences moved to `on_<period>_TAG` polls, 18 two-sided events split into sender/recipient pairs, 7 redundant fires deleted, 3 restructured. uae_royal.1/.2 were `major = yes` news events re-firing themselves immediately; sov_economic.45 and Serbia.8 recursed the same way.

Every converted recurrence keeps a deterministic interval: a timed cooldown flag sized so the next poll tick lands where the event used to reschedule itself, never a per-tick probability roll. The flags are the carrier the queue entry used to be — one per loop, plus CHI_scs_fonop_campaign for membership in the SCS patrol campaign and ceasefire_negotiated, which keeps Kachin's start-state ceasefire on its old 1%/month instead of the negotiated 3%. Dropping `random_days` jitter means the five ENG inner-circle pickers now fire on their old `days = 150` base rather than the 150-200 spread.

Two chains were unreachable on main because a self-fire was their only reference, so they are deleted rather than wired up: GER_agrar.7/.8 (with the five ideas only they granted) and chi_bri.95 with its CHI_cai_suspended opinion modifier — the CAI-suspension gate read an idea nothing ever added. Note for whoever wants that content back: GER_agrar.5/.6/.9-.12 are already dispatched from a random_list in 99_GER_on_actions.txt that .7 and .8 were simply never added to.

### Changes
**Please describe the scope of the changes for this pull request.**

Closes #<issue number here>


events/Ivory_Coast.txt
```diff
@@ -213,6 +204,61 @@ country_event = {
...
+	title = ivory_coast_md.16.t
+	desc = ivory_coast_md.16.d
+	picture = GFX_ivory_coast_protests
+	is_triggered_only = yes
+
+	option = { #Suppress again
+		name = ivory_coast_md.16.o1
+		log = "[GetDateText]: [This.GetName]: ivory_coast_md.16.o1 executed"
+		add_stability = -0.10
+		country_event = {
```
Blazer135: [P1] Preserve the renewed-protest civil-war odds

Suppressing renewed northern protests now always queues ivory_coast_md.6, causing civil war. Previously, after CDI_1st_protests was set, this was a 50% civil-war roll and a 50% suppressed-protests outcome followed by another protest round. Please preserve that roll and move the continuing protest schedule into the on_action; removing the self-fire should not make civil war certain after the second crackdown.

BLUF

common/scripted_effects/00_antarctica_effects.txt
```diff
@@ -3407,6 +3416,20 @@ antarctica_process_blizzard_daily = {
 			log = "[GetDateText]: [Root.GetName]: Antarctica blizzard ended. Starting module repairs."
 		}
 	}
+	else = {
```
Blazer135: [P2] Decrement the next-storm timer during active blizzards

The next-start countdown is set when a blizzard begins, but this branch only decrements it while antarctica_blizzard_active is false. The removed event scheduled the next start 300-350 days after the previous start, so this adds the 18-35-day storm duration to every interval. Decrement the countdown daily in both states, while keeping the next-storm trigger guarded against an active blizzard.

BLUF

common/on_actions/99_GRE_on_actions.txt
```diff
@@ -43,6 +43,20 @@ on_actions = {
 				}
 			}
 
+			# greece.702 — Aegean incident recurrence. The 75d cooldown, armed by
```
Blazer135: [P2] Preserve the 90-day Aegean incident interval

This runs in on_weekly_GRE, so the 75-day cooldown makes greece.702 recur on the next weekly tick around day 77, rather than the previous 90-day interval. The focus and greece.702 both set that cooldown. Since each incident changes tension and may trigger an ultimatum, this accelerates escalation. Use a daily countdown initialized to 90 in both places if the refactor is meant to preserve the exact interval, and update the comment that assumes a monthly tick.

BLUF

Blazer135 (commented):
3 findings from the verified review are attached inline. BLUF

PR #3877, opened 2026-09-09
Religious Factions Exclusivity

Lynxivium:
Taking both ulemas pretty much doesn't make sense.

### DIFF
--- a/common/ideas/AA_law_internal_factions.txt
+++ b/common/ideas/AA_law_internal_factions.txt
@@ -483,6 +483,13 @@ ideas = {
 					has_idea = shinto
 					has_idea = cheondo
 				}
+				NOT = {
+					OR = {
+						has_idea = the_ulema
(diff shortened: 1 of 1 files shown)


Lynxivium (commented):
They are the same thing, but wahabi is for specific countries

common/ideas/AA_law_internal_factions.txt
```diff
@@ -517,6 +517,7 @@ ideas = {
 						NOT = { has_completed_focus = NIG_suppression_of_islam }
 					}
 				}
```
AngriestBird, line 533: please make is so you can't have any of the other religious factions. Preisthood, Clergy, Ulema, or WEahabbists shoudl all be mutually exclusive

AngriestBird (changes requested):

AngriestBird (approved):

Issue #3862, opened 2026-09-09
[BUG] Algeria, focus Maghreb Federation bugged


millenniumdawnmod-hub:
> *Reported by **discord-sync** via [Discord](https://discord.com/channels/210890013334831104/1547265398775226459) on 2026-09-09*
> *Tags: General Bug*

**What happens**
There is no Libya but i cannot go down the focus cause the requirement are not met. You should insert an "or the country doesn't exist".

**Where**
See Discord thread for details.

**Versions**
See Discord thread for details.

**To Reproduce**
See Discord thread for details.

**Operating System**
See Discord thread for details.

**Additional context**
Auto-created from a Discord forum post.

### Attachments
![image.png](https://github.com/MillenniumDawn/Millennium-Dawn/releases/download/bug-attachments-2026Q3/asset-1-1547265399266091048.png)

![image.png](https://github.com/MillenniumDawn/Millennium-Dawn/releases/download/bug-attachments-2026Q3/asset-1-1547265400004280461.png)

millenniumdawnmod-hub:
**Reply from Discord:**

https://klipy.com/gifs/on-the-case-search-3

AngriestBird:
Merged earlier #3926

PR #3926, opened 2026-09-10
Fix Maghreb focus requirement when Libya does not exist

Freshpriest1993:
Use this for the PR Changes description:
Updated Algeria’s Maghreb Union focus requirements so Libya no longer blocks progression when the country does not exist.
Previously, the focus required Libya to have the maghreb_union_member country flag in all cases. This caused the focus to remain unavailable if Libya had been annexed or otherwise removed from the map.
The requirement now passes when either:
- Libya exists and has the maghreb_union_member flag, or
- Libya does not exist.
Applied the fix to both affected focus-tree checks in common/national_focus/05_algeria.txt.


common/national_focus/05_algeria.txt
```diff
@@ -25515,7 +25515,8 @@ focus_tree = {
 		available = {
 			TUN = { has_country_flag = maghreb_union_member }
```
AngriestBird: The logic you did for LBA should be for all of them and not just LBA
```
if = {
    limit = { country_exists = LBA } 
    LBA = { has_country_flag = maghreb_union_member
}
if = {
    limit = { country_exists = TUN } 
    TUN = { has_country_flag = maghreb_union_member
}
```

AngriestBird (changes requested):

Freshpriest1993:
Fixed the indentation validation issues in common/national_focus/05_algeria.txt.  
The latest validation run now reports 0 new errors against the main baseline. The remaining 31 Event errors appear to be unrelated baseline errors from other files/countries, and this PR currently only changes 05_algeria.txt.  
I’ve left those unrelated errors untouched to keep this PR scoped to #3862.

AngriestBird (approved):

PR #3953, opened 2026-09-10
Enable redundant-focus-flag warnings in CI

### LINKED ISSUE #3939: [VALIDATION] has_country_flag validation
**Description**
A clear and concise description of the requested validation suite.

There are a lot of country flags added and check which can be replaced by using native triggers. The biggest one will be:

has_completed_focus

**Additional context**

AngriestBird:
## Bottom line

CI and `run_all_validators.py` now pass `--redundant-focus-flags`. Country flags that only latched a single focus completion now use `has_completed_focus`, except where that swap is not equivalent: USB self-tree wipe, IRQ/UIQ `keep_completed = no`, SYR `Insulted_Iran` read on other tags, and two EH flags whose formation events reload `event_horizon_generic_focus` without `keep_completed`.

## Why

Issue #3939 asked for a validator that finds `has_country_flag` uses replaceable by native triggers, especially `has_completed_focus`. That check already existed. It never ran in CI.

## How

`ValidatorSpec.args` for the core-batch variables job, and `_VALIDATOR_EXTRA_FLAGS` for the local suite. Content rewrite is short-form only. Loc keys were left in place. Standalone CLI stays opt-in. `--strict` still does not gate the remaining flags.

Closes #3939

BLUF


AngriestBird:
Checked the leftover 16 against actual `load_focus_tree` sites, not just the set-file caution.

**Kept as flags (not equivalent to `has_completed_focus`)**
- USB `FS_continued_election_ban` / `FS_established_elections`: `USB_america_the_proclamation` loads `generic_focus` on USB itself.
- `Insulted_Iran`: set by SYR focus `SYR_shadow_over_hezbollah`, read on *other* tags in reversed `PER_support_shias`.
- Eleven IRQ flags including `all_shias_iraq`: the caution text is a Syria tree load in `05_iraq.txt`, but `events/Iraq.txt` still does `keep_completed = no` on UIQ. A rebel that inherits the flag would diverge.

**Fixed in `b98a498da3`**
- Reverted the two EH swaps (`EH_nation_activated_electric_traps_flag`, `EH_research_sites_enabled`). Formation events load `event_horizon_generic_focus` without `keep_completed`.
- Replaced `UKR_vitrenko_power` and `UKR_union_state`. The caution was SOV/BLR rebel tree loads in the same file; UKR's own reloads use `keep_completed = yes`, and the reads are on UKR (including `UKR = { ... }` from SOV/Russia/USR loc).

IRQ/USB/`Insulted_Iran` stay WARNING and ungated.

AngriestBird:
Approving w/ bypass. I need to get this in here with the new cached values.

Issue #3895, opened 2026-09-09
[VALIDATOR] Party localisation standardization backlog


TemplarGeneral:
## Bottom line

`validate_party_loc.py` reports **7194 findings across 292 nations** in `localisation/english/MD_politics_view_parties_l_english.yml`. Greece (#3887) is the only worked example; every nation below still predates the standard in [.claude/docs/party-loc-reference.md](.claude/docs/party-loc-reference.md).

The validator is warning-only and scoped to the tags a branch touches, so none of this blocks CI. It is a backlog list, not a gate.

Tags that do not own a state on the 2000 start, and fictional tags, are tracked separately in #3996. Nations without a custom focus tree in `common/national_focus/` are tracked in #4280.

## How to work one nation

```bash
python tools/validation/validate_party_loc.py --tag GRE
```

Then run `/party-names <TAG>` and re-run until clean. One nation per PR is fine, a handful of small ones per PR is better.

## What is being checked

| Category | Findings | Meaning |
| --- | ---: | --- |
| `party-loc-name-format` | 2727 | Name is not `£sprite (ABBRV) - Party Name` |
| `party-loc-empty-desc` | 985 | `_desc` key is an empty string |
| `party-loc-desc-header` | 846 | Description does not open with `(Ideology Group) - Name (Native: ..., ABBRV)` |
| `party-loc-desc-body` | 1063 | Description has no `\n\n` between header and body |
| `party-loc-missing-hook` | 1496 | Loc key has no hook in the politics-view switch, so it never renders |
| `party-loc-orphan-hook` | 17 | Hook points at a loc key that does not exist |
| `party-loc-key-without-name` | 23 | `_desc`/`_icon` key with no matching name key |
| `party-loc-icon-sprite-mismatch` | 12 | `_icon` sprite disagrees with the name key's sprite |
| `party-loc-duplicate-hook` | 17 | Block gates the same `original_tag` twice; the second entry is dead |
| `party-loc-slot-case` | 8 | Subideology is miscased, so the key is dead on Linux |

Missing nations and missing slots are deliberately **not** listed: an unfilled slot is supposed to fall through to the generic label rather than have a party invented for it.

## Nations (80)

Counts are findings per nation, with the per-category breakdown in brackets.

- [x] **ABK** Abkhazia — 18 (name 5, desc-empty 6, hook 6, orphan-hook 1)
- [x] **ACE** Aceh — 3 (name 1, desc-empty 1, hook 1)
[...]

PR #3975, opened 2026-09-10
Standardize party localisation for ALG and ALG_FIS

TemplarGeneral:
### Changes
- Reworked the Algeria and FIS party entries in the politics view to the (ABBRV) - Name standard with full two-paragraph descriptions, and wired every name, description and icon key into the politics-view hooks
- Algeria's parties now sit on the slots that match their ideology: the RND under pro-Western autocrats, the RCD under liberals, the Party of Algerian Renewal (PRA, not "PAR") under libertarians, the FFS under social democrats, the FNA under right-wing populists and Ahd 54 under reactionaries, with starting popularity and leaders following each party
- Remapped the ALG focus, event and decision party slot references to the new slots
- The invented Algerian Green Party is replaced by the real National Movement for Nature and Development, Djaballah's Justice and Development Front appears from February 2012 as the successor to Islah, and the UDR no longer exists before its 2003 founding
- Fixed the FIS and Tamazgha Movement entries never showing because their hooks sat behind unconditional lines, the AQIM description showing Saudi text, and Boukrouh being created under the wrong ideology

Part of #3895


AngriestBird:
This has a large amount of diff so I cannot review. Can we reapply this fix/changes to a new branch with the new sorting so the new alphabetical is sorted?

SBeausoleil (commented):
Changes to the German ideas file will do a very large collision with PR https://github.com/MillenniumDawn/Millennium-Dawn/pull/4200 which not only fixes the indentation, but also adds content. Since your PR isn't linked to it and doesn't seem to alter the content of that file beyond formatting, I recommend removing it.

SBeausoleil (approved):

SBeausoleil:
Sorry, accidentally closed the pull request.

SBeausoleil (changes requested):
Remove changes formatting changes to the german file, they are fixed in another PR (4200) that also alter its content.

SBeausoleil (dismissed):

AngriestBird (changes requested):
See above comment

AngriestBird (approved):

PR #3990, opened 2026-09-10
Fix Algeria drug crisis spirit cleanup

### LINKED ISSUE #3981: [BUG] Drug Crisis in Algeria Bug
> *Reported by **discord-sync** via [Discord](https://discord.com/channels/210890013334831104/1547673986740125797) on 2026-09-10*
> *Tags: General Bug*

**What happens**
I tried completing the entire focus path related to the drug crisis, but it doesn’t go away. This is because the focus search for another spirit i think that is: Drug problem. Please Fix.

**Where**

**Versions**

**To Reproduce**

**Operating System**

**Additional context**

Freshpriest1993:
## Summary

Fixes #3981.

The Algeria drug crisis focus path was attempting to remove the wrong idea ID when completing either response branch.

`ALG_drug_problem` is the focus ID, while the national spirit applied by the focus is `ALG_drug_crisis`.

## Fix

Updated both Algeria drug-policy branches:

- `ALG_relax_restrictions`
- `ALG_tighten_punishments`

Changed:

```txt
remove_idea = ALG_drug_problem

### CODE REVIEW

File: common/national_focus/05_algeria.txt [hunk 1]
```diff
@@ -14806,7 +14806,7 @@ focus_tree = {
 			log = "[GetDateText]: [Root.GetName]: Focus ALG_relax_restrictions"
 			custom_effect_tooltip = ALG_relax_restrictions_tt
 			swap_ideas = {
-				remove_idea = ALG_drug_problem
+			remove_idea = ALG_drug_crisis
```
AngriestBird: Does this focus also account for ALG_drug_problem? I feel like these might be two different ideas handling the same thing.

AngriestBird (changes requested):

common/national_focus/05_algeria.txt
Freshpriest1993: Good catch. I traced the event flow and confirmed ALG_drug_problem is a separate idea added by ALG_flavor_event.7, while the ALG_drug_problem focus later adds ALG_drug_crisis. Both therefore need to be handled. I’ll preserve the existing swap_ideas removal of ALG_drug_problem and separately remove ALG_drug_crisis in both follow-up focuses.

Will fix within 2 hours.

common/national_focus/05_algeria.txt
```diff
@@ -14805,6 +14805,7 @@ focus_tree = {
 		completion_reward = {
 			log = "[GetDateText]: [Root.GetName]: Focus ALG_relax_restrictions"
 			custom_effect_tooltip = ALG_relax_restrictions_tt
+			remove_ideas = ALG_drug_crisis
```
AngriestBird: Let's do one more quick change.
```
if = { limit  = { has_idea = ALG_drug_crisis } remove_ideas = ALG_drug_crisis } 
```

AngriestBird (changes requested):

common/national_focus/05_algeria.txt
Freshpriest1993: Updated as requested. Both ALG_relax_restrictions and ALG_tighten_punishments now check has_idea = ALG_drug_crisis before removing the spirit, while preserving the existing ALG_drug_problem swap logic. I also traced ALG_flavor_event.7 and confirmed it is gated by the ALG_drugs flag, which the event sets, with no clear path found that removes that flag later. Thanks for catching the edge case.

AngriestBird (approved):

PR #4056, opened 2026-09-11
Split system keys out of party loc and drop merge duplicates

AngriestBird:
## Bottom line
System and generic party strings live in 0_political_system. MD_politics_view_parties is country keys only, one copy each, sorted by tag. The merge that duplicated about 7000 keys is cleaned up.

## Why
Merges from main concatenated the old parties file onto the reorganized one, so system keys and whole country blocks appeared twice.

BLUF


AngriestBird:
Approving this w/ Dread. We handled this together.

PR #4191, opened 2026-09-13
Remove unused SAF placeholder scripted effects

TemplarGeneral:
## Bottom line
Deletes the 5 unused SAF scripted effects flagged by the validator and their orphaned tooltip keys.

## Why
- `99_SAF_scripted_effects.txt` came in with #4056 as WIP placeholders; nothing calls any of the 5 effects.
- The 4 `_party_founded_TT` keys in `MD_focus_SAF_l_english.yml` were referenced only by those effects.
- `SAF_error_stopper` set a flag whose history line is already commented out.

## How
- `git rm` both files (each would otherwise be empty).
- Politics-view scripted loc flag branches left as-is; the flags were never set before either, so behaviour is unchanged.

BLUF


AngriestBird:
Approved by moi. This is going to be closed since we need that for something else. We can ignore these scripted effects if we really want to.

PR #4312, opened 2026-09-14
R3D sp agtm reassignment

VitorFox82:
- Mechanized and Armored infantry no longer use mounted atgm
 - Heavy Motorised, Airborine Motorised, Light Engineers, Amphib Landing Detachment all use mounted atgms in some capacity
 - Heavy Engineers are now 2 width, cannot be air dropped and use tanks for support to represent heavy demolition and bridging vehicles


AngriestBird (changes requested):
Unit changes look good, changelog is messed up though and should be restored back to normal and then changes reapplied

AngriestBird (approved):

PR #4336, opened 2026-09-15
Stop the Iraqi civil war GUI firing events at every country

JasonBreen:
## Bottom line

`Iraq_civil_war_gui` lets every AI country in the game run Iraq's civil war buttons, so all ~225 countries mail themselves `iraq_civil_war.2`-`.5` and lose equipment to the Iraqi insurgency. Gating `ai_enabled` on `original_tag = IRQ` stops it.

## The bug

`common/scripted_guis/99_IRQ_scripted_guis.txt` declares `ai_enabled = { always = yes }` and `ai_check = { always = yes }` with no `visible` block.

`IRQ_civil_war_mechanic` is correctly restricted (`allowed = { original_tag = IRQ }`), but a decision category governs the player's UI only. It does not stop the AI evaluating the scripted GUI's effects.

The button guards do not scope to Iraq either:

- `NOT = { IRQ_is_selected_state_iraq = yes }` is true for any country holding none of the `IRQ_*_selected` flags, i.e. every country except Iraq.
- `168 = { has_state_flag = IRQ_high_insurgent_presence }` is a global state check, independent of who is evaluating it.

So every AI country ran `IRQ_conduct_attack`, read `IRQ_government_readiness` as `0` in its own scope, fell into the below-50 branch, and fired one of the four events at itself.

## Impact

Measured in a 2002 campaign: **297 events across 225 distinct countries in two game months** (Vatican City, Tuvalu, San Marino, Samoa, Sweden, Zimbabwe...). Each one runs `add_equipment_to_stockpile` for **-100 `infantry_weapons_1`** and **-50 `cnc_equipment_1`** on the receiving country and writes `IRQ_*` insurgency variables into its scope. Plus the per-tick cost of every AI country evaluating the GUI.

## The fix

`ai_enabled = { original_tag = IRQ }`, matching the category's own `allowed` block and the existing pattern in the EU GUIs. `ai_check` is left as-is; it now only ever runs for Iraq.

`original_tag` rather than `tag` is deliberate, so Iraq's own occupation-zone tags keep participating.

## Verification

Same save, same stretch of game time, before and after:

| | before | after |
|---|---|---|
| events | 297 | 8 |
| distinct countries | 225 | 0 non-Iraq |
| recipients | every country | Iraq + its occupation zones |

AI-controlled Iraq still runs its civil war normally.

## Notes

A scan of all scripted GUIs found no other case of `ai_enabled = { always = yes }` combined with effects and no `visible` gate, so this looks isolated. PR #3616 audited GUI AI config and fixed the sibling `american_playing_cards` in this same file, but skipped this one because it has effects and so did not read as a config mismatch.

BLUF


AngriestBird (dismissed):
Changelog required for this

JasonBreen:
Added, under Bugfix in v2.0.1.

TemplarGeneral (approved):

PR #4342, opened 2026-09-15
Standardise Syria AI path game rule

TemplarGeneral:
### Changes
- Syria's AI behaviour rule now offers Historical, Iron Lady of Syria, Syrian Savage, New al-Assad, Secular Coalition, Muslim Brotherhood, National Progressive Front, Random Path and No Path
- Every Syrian focus branch follows the chosen path, including the Damascus Spring election outcomes, Lebanon, the post-war alignment and the diplomatic and military trees
- The separate "Syria behaviour during the Arab Spring" rule is folded into the Historical path, which now guarantees the civil war and makes the Islamic State far likelier to rise
- Non-Syrian Arab Spring regimes no longer get the Syrian civil-war weighting on refusing to step down
- AI Syria can no longer strand its Lebanon withdrawal on a wrongly flagged focus, and stops opening new wars while losing one

Part of #3162


common/ai_strategy/SYR.txt
```diff
@@ -803,3 +803,26 @@ SYR_jerusalem_pact = {
 	ai_strategy = { type = contain id = "ISR" value = 50 }
 	ai_strategy = { type = antagonize id = "ISR" value = 50 }
 }
+
+SYR_hold_new_wars_while_losing = {
```
AngriestBird: Uneeded. Please update the skill so we aren't adding duplication to the general war goal declaration suite.

AngriestBird (changes requested):

common/ai_strategy/SYR.txt
TemplarGeneral: Removed

AngriestBird (approved):

Issue #4168, opened 2026-09-13
[BUG] There are defects in the focal tree of the Pattani issue in Thailand


millenniumdawnmod-hub:
> *Reported by **discord-sync** via [Discord](https://discord.com/channels/210890013334831104/1548569263936241735) on 2026-09-13*
> *Tags: General Bug*

**What happens**
After a player completes any one of the routes, the debuff of 'The Deep South Insurgency' will be removed, resulting in the initial focuses of the other two routes being skipped. Subsequently, the following focuses have no ruling party restriction, which effectively allows players to complete all three routes and obtain all the benefits, contrary to the original design intention.
version: 2.0.0
mod checksum:442a

**Where**
See Discord thread for details.

**Versions**
See Discord thread for details.

**To Reproduce**
See Discord thread for details.

**Operating System**
See Discord thread for details.

**Additional context**
Auto-created from a Discord forum post.

### Attachments
![image.png](https://github.com/MillenniumDawn/Millennium-Dawn/releases/download/bug-attachments-2026Q3/asset-1-1548569264569458769.png)

millenniumdawnmod-hub:
**AI Related Issues**

Pattani focal tree defect in Thailand: completing one route removes the Deep South Insurgency debuff, skipping other routes' initial focuses and removing ruling party restrictions, letting players complete all three routes against design intent.

Severity: :orange_circle: High · Suggested labels: `bug`, `exploit`, `focus-trees`, `needs-verified`

**Possibly related issues:**
- #4168 — Duplicate (high) — [BUG] There are defects in the focal tree of the Pattani issue in Thailand

millenniumdawnmod-hub:
**Reply from Discord:**

Thanks for raising this. I'll take a look to see why ruling party restrictions could be skipped and could be taken without initial focus

AngriestBird:
Closed in #4362

Issue #3882, opened 2026-09-09
[BUG] Thailand | Political Party paths can be permanently unlocked if the ruling party is no longer in power


AngriestBird:
**What happens**
What you saw, and what you expected instead.

While playtesting if Thailand elects another party via the regular election systems you can end up where you have flavored content for other political parties. 

Example: Start with the Thai Rak Thai party and then elect the Communist Party then you can do both full trees and keep rotating through that. 

The main issue largely due to the country_flags here.  

**Where**
The file and line if you know it (e.g. `common/decisions/05_ENG_decisions.txt:3256`), plus the focus, decision, event, idea or GUI ID involved. Paste the matching `error.log` line if there is one.

**Versions**
Game version and checksum, MD version (release number, or the dev branch date you pulled).

**Additional context**
Screenshots, a save game, or anything else that helps.

cal-chu:
PR linked with fix https://github.com/MillenniumDawn/Millennium-Dawn/pull/4362

cal-chu:
Should be now resolved in https://github.com/MillenniumDawn/Millennium-Dawn/pull/4362

AngriestBird:
Closed with #4362

PR #4362, opened 2026-09-15
Thailand: fix multi-trunk political path exploit and rework the red road

cal-chu:
## Bottom line

Thailand's political trunks could all be completed in a single game, and the Deep South's three
doctrines could all be taken once any one of them resolved. This closes both exploits, rebuilds
the branch around the party that actually governs, and reworks the communist road so it can be
entered and left cleanly.

### Political paths

- Each trunk now checks who is in power right now, not a flag set once at the start. Lose an
  election and the trunk closes; win it back and it reopens
- Leaving a trunk tears down everything it left running in one place: the strike, the organising
  drive, timed spirits and stale decision categories
- Ghosts of the Jungle and Seeds of Reform commit you to a trunk before that party holds power,
  so they no longer strip the sitting government's spirits. That happens when the party wins
- Every realignment now has a notification event, including the communist one

### The red road

- Protracted People's War launches the Communist Party's uprising, and winning that war grants
  the same rewards as a negotiated handover instead of skipping them
- The General Strike runs for a year with five instruments, each costing a different resource,
  and a live countdown in the decision category
- If the Communist Party ever wins an election, it takes power automatically. A lost election is
  now a delay rather than a dead end
- The republic can only be proclaimed once the uprising has been won, and extraordinary elections
  are blocked while the road runs

### The palace

- The palace policy focuses need a government that is not communist and not Move Forward.
  Thailand is a constitutional monarchy, so the crown and an elected government coexist
- The 2006 coup can be taken while Thai Rak Thai governs, and both coups bypass if a junta is
  already in power
- The Yellow Shirts and the Quiet Word now apply a daily nationalist drift instead of a flat
  political power grant

### The Deep South

- The three doctrines are mutually exclusive. Finishing one no longer skips the other two and
  hands over every reward. Each is open to the trunks that would plausibly pursue it
- The whole sub-branch retires itself once the insurgency ends
- The three intensity tiers have their own names and descriptions instead of reading as one idea,
  and the category explains the grievance meter and the floor it sets

### Economy

- The Sufficiency Economy focus was a duplicate of Sufficiency and the Crown. Removed, with its
  decision migrated across
- Sufficiency now costs consumer goods rather than saving them, matching its own description
- Repay the IMF Early no longer drops its reward when the deadline lands on a focus already more
  than half finished

### Campaigns and AI

- Focuses that granted a lump of party support now run a campaign that builds it week by week
  over a year, so the payoff can be contested
- The alt-history AI can follow the communist road through to the end instead of stalling after
  the first focus

### Localisation

- Tooltips name the party a focus requires, and requirements that list several parties render one
  line each instead of one run-on sentence
- Support gains state the rate and the duration rather than a figure the focus never applied
- Country flags that rendered as raw identifiers now read as sentences

Closes #[3882](https://github.com/MillenniumDawn/Millennium-Dawn/issues/3882) and #[4168](https://github.com/MillenniumDawn/Millennium-Dawn/issues/4168)


cal-chu:
image
image
image

Fixing and tested root cause issue. Now when you start the communist political path, you are unable to continue down the path for any democratic/NCPO party political tree

TemplarGeneral:
I found some stuff:

1. common/national_focus/05_thailand.txt:8465 — SIA_people_seize_power bypass (ruling_party = 4 + has_civil_war = no) matches the post-victory state of the new PPW civil war, so the armed road skips its reward (barami, pillars, SIA_kings_moral_authority, SIA_peoples_constitution removal, thailand.251).

2. common/national_focus/05_thailand.txt:8262 — Ghosts of the Jungle calls SIA_realign_to_communist at 5% communists on any trunk, stripping the sitting government's ideas and locking its trunk while it still rules.

3. common/scripted_triggers/SIA_thailand_triggers.txt:9 — hidden_trigger { NOT SIA_path_communist } on all four trunk_open triggers plus the detector guard at common/on_actions/99_SIA_on_actions.txt:403 means no later election clears the flag or reopens a trunk.

4. common/national_focus/05_thailand.txt:2918 — Seeds of Reform clears every path flag and strips populist/royalist/democrat ideas without changing ruling party, so the detector never re-fires and the player loses flag-gated decisions and weekly systems.

cal-chu:
@TemplarGeneral I've reviewed your comment and made the following fix and committed:
>  1. The bypass skipped the armed road's payoff because protracted_peoples_war sets
> ruling_party to 4 before it starts the civil war, so winning the war satisfied both bypass
> conditions at once. I pulled the reward out into SIA_communist_seizure_effect and call it
> from both completion_reward and a new bypass_effect, so the armed road keeps it.
> 
> 2. Ghosts of the Jungle now only commits to the road. The four SIA_strip_*_ideas calls
> moved into that same seizure effect, so a sitting government loses its ideas when it
> actually falls rather than when the focus completes.
> 
> 3. Removed the four hidden_trigger lines. They were dead weight once the trunks moved onto
> live ruling_party, since ruling_party 18, 22, 23, 1 and 3 are all false while the
> communists govern. I also dropped the flag from the detector guard, so an election can
> realign off the road again.
> 
> 4. Seeds of Reform now calls SIA_realign_to_reform, which is the same set of effects plus
> the SIA_path_reform flag it was never setting.
> 
> Because the detector now reacts to elections, Ghosts sets generic_election_killswitch. It
> lifts when the revolution succeeds, when it fails, or at the next scheduled election.

Also made a few changes so that elections don't happen when players start kicking off the communist path:
image


I've also mixed in the changes from this issue into this PR: https://github.com/MillenniumDawn/Millennium-Dawn/issues/4168
image

TemplarGeneral (changes requested):
Found some more stuff:

1. [SIA_thailand_effects.txt:558](vscode-webview://01ukn97jr7jkja19t4u0tnnfa5jvujauoivgucin34omlh9nl6h4/common/scripted_effects/SIA_thailand_effects.txt#L558) — SIA_communist_seizure_effect never calls SIA_strike_cleanup_effect. Communist election win mid-strike: SIA_people_seize_power auto-bypasses, the strike keeps draining stability, and thailand.259 later installs a junta over the communist government.
2. [05_thailand.txt:2927](vscode-webview://01ukn97jr7jkja19t4u0tnnfa5jvujauoivgucin34omlh9nl6h4/common/national_focus/05_thailand.txt#L2927) — SIA_seeds_of_reform calls SIA_realign_to_reform, which clears every SIA_path_* flag and strips populist/royalist/democrat spirits (incl. SIA_junta_government, SIA_democrat_machine_active) while that party still governs. The monthly detector only re-fires on a ruling-party change, so nothing restores them. AI picks it at base 35.
3. [SIA_thailand_effects.txt:514](vscode-webview://01ukn97jr7jkja19t4u0tnnfa5jvujauoivgucin34omlh9nl6h4/common/scripted_effects/SIA_thailand_effects.txt#L514) — SIA_clear_path_flags (any election realignment) clears SIA_path_communist during an active strike without cleanup: SIA_red_path_category hides the momentum decisions, the weekly drain continues, and thailand.259 coups the newly elected government.
4. [05_thailand.txt:7043](vscode-webview://01ukn97jr7jkja19t4u0tnnfa5jvujauoivgucin34omlh9nl6h4/common/national_focus/05_thailand.txt#L7043) — SIA_the_southern_question seeds intensity 30 but adds SIA_deep_south_insurgency_1; SIA_ds_update_tier treats > 29 as tier 2, so the tooltip understates the spirit until the weekly tick swaps it.
5. [05_thailand.txt:183](vscode-webview://01ukn97jr7jkja19t4u0tnnfa5jvujauoivgucin34omlh9nl6h4/common/national_focus/05_thailand.txt#L183) — bypass = { date > 2005.1.1 } on SIA_repay_the_imf also fires on an in-progress focus started late 2004, dropping the reward and showing the lapsed thailand.51 text.

cal-chu:
@TemplarGeneral Fixed with latest few commits:

> 1. Fixed. The seizure now shuts the strike and the organising drive down before it installs the new government, so that government doesn't inherit the stability drain, and thailand.259 can't coup it on day 365. A new SIA_communist_seized flag means the focus, its bypass and the weekly election check all run the same code once. SIA_thailand_effects.txt:582
> 
> 2. Fixed, by removing the call. Seeds of Reform no longer runs SIA_realign_to_reform at all. It just commits you to the trunk and leaves the sitting government's spirits alone, like Ghosts of the Jungle does. The strip happens later, when Move Forward actually wins. I also dropped the detector's reform guard so that win still registers. 05_thailand.txt:2920, 99_SIA_on_actions.txt:481
> 
> 3. Fixed. Shutting the strike down is now part of SIA_clear_path_flags instead of something every caller has to remember, along with the cadre spirit and any running drive. One exception: if the drive was pushing the party that just took power, it did its job, so it's left to finish. SIA_thailand_effects.txt:521
> 
> 4. Fixed by naming SIA_deep_south_insurgency_2 directly. I tried deriving it through SIA_ds_update_tier first, but that blanks the tooltip: tooltips check if conditions against the state right now, and the intensity variable isn't set yet when the tooltip is drawn. 05_thailand.txt:7054
> 
> 5. Fixed with a progress check. The bypass now also requires that the focus isn't more than half finished. Past halfway you still get paid; below that it lapses, and thailand.51 has a separate branch for that case. 05_thailand.txt:187, Thailand.txt:977

Also I've added a few more changes to other unrelated parts of the focus tree as I kept finding new things to finetune and adjust from testing playthroughs. It should be more robust and consistency. Realized the politcal trees were really complicated. Now the royal focuses could also be taken along side the TRT and democrat party trees which is intentional (path to soft coup or so forth) which is realistic as thailand is a consitutional monarchy. Also did some balances to ideas and focuses, removed redundant concepts. I've run audit skill a few times if you could review again to make sure is all good

TemplarGeneral (approved):

PR #4386, opened 2026-09-15
Add three Portuguese defence manufacturers

JasonBreen:
## Bottom line

UMM, Tekever and UAVision join Portugal's six existing manufacturers. Script and localisation are complete; icons use the existing Generic placeholders pending artwork.

## The companies

| Organization | Founded | Availability | Template |
|---|---|---|---|
| `POR_umm_tank_manufacturer` | 1977 | from start | `generic_AFV_equipment_organization` |
| `POR_tekever_aircraft_manufacturer` | 2001 | `date > 2001.01.01` | `generic_light_aircraft_equipment_organization` |
| `POR_uavision_aircraft_manufacturer` | 2005 | `date > 2005.01.01` | `generic_light_aircraft_equipment_organization` |

Each entry follows the include-a-generic-template shape the rest of `MD_POR_organizations.txt` already uses, so no bespoke trait trees are introduced.

Tekever and UAVision post-date the 2000 start, so they gate on `available = { date > … }` exactly as the Brazilian, French and German organizations already do. Both are drone builders, and `generic_light_aircraft_equipment_organization` already covers `mio_cat_eq_only_uav`, so a bespoke organization is unnecessary.

UMM predates the start and needs no gate. Worth noting for a follow-up: Portuguese production wound down around 2006, so it may eventually want an end-date or decline treatment rather than running to 2050.

## Icons

The three sprites point at the existing `gfx/interface/ideas/companies/Generic/` placeholders, so nothing renders broken in the MIO list. Substituting real artwork later changes one `texturefile` line per sprite and touches neither the organizations nor the localisation.

## Verification

`validate_mios` reports no issues. That run was confirmed with a positive control — deliberately pointing one organization at a non-existent sprite produced `mio-icon-unresolved` at the expected line, and restoring it cleared the finding, so the clean result is not a silent pass.

BLUF


common/military_industrial_organization/organizations/MD_POR_organizations.txt
```diff
@@ -39,3 +39,26 @@ POR_ogma_aircraft_manufacturer = {
 	name = POR_ogma_aircraft_manufacturer
 	include = generic_fixed_wing_equipment_organization
 }
+
+POR_umm_tank_manufacturer = {
+	allowed = { original_tag = POR }
```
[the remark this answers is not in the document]
JasonBreen: Not changing this one, and I do not think it should change here.

That trigger locks a country out of the generic MIO roster, per the note at the top of MD_generic_organizations.txt. Every tag currently in it carries thousands of lines of bespoke organizations with no generic includes at all: BRA 6684, JAP 9445, GER 18745, POL 14869, and so on. Portugal's file is 66 lines and eight of those are include = generic_*_organization, so all nine of its organizations are thin naming wrappers around the same generic templates the catalog would offer.

Adding original_tag = POR would therefore lock Portugal out of the generic roster and leave it with nine wrappers that only re-expose a subset of it, which is a net loss of equipment coverage rather than an integration fix.

It is also pre-existing rather than something this PR introduces: Portugal already had six organizations in this shape before these three were added.

Happy to be overruled if the intent is for Portugal to graduate to bespoke organizations, but that is a content decision and a much larger piece of work than this PR.

VitorFox82 (changes requested):
Would you be okay with changing UMM to generic_utility_vehicle_manufacturer and UAVision to GENERIC_skyforge_autonomous_systems, this one may need some loc changes?

JasonBreen:
Done, both changed and pushed.

UMM now uses `generic_utility_vehicle_manufacturer`, which is the better fit anyway — UMM built light off-road utility vehicles rather than armour, so the AFV template was overreaching.

UAVision now uses `GENERIC_skyforge_autonomous_systems`, following how `UKR_DeViRo_drone_manufacturer` consumes it: own `allowed`, `name` and `icon`, with the 2005 availability kept as an override the way DeViRo overrides its own date gate.

No localisation change turned out to be needed. The organization sets `name = POR_uavision_aircraft_manufacturer` explicitly, which overrides the template's `GENERIC_skyforge_autonomous_systems_name`, and that key already exists.

Three things worth your call, none of which I have changed:

1. The skyforge template's `visible` block requires `has_country_flag = skyforge_autonomous_systems_unlocked` or `GENERIC_skyforge_autonomous_systems_allowed`. DeViRo overrides `available` but not `visible`, so I matched that exactly — but it does mean UAVision stays hidden until Portugal satisfies one of those. Tell me if it should override `visible` too.

2. Tekever is also a drone builder and is still on `generic_light_aircraft_equipment_organization`. Happy to move it to skyforge as well if you want the two treated consistently — I left it alone since you only named UAVision.

3. The UMM token is still `POR_umm_tank_manufacturer`, which now reads oddly against a utility-vehicle template. Renaming it touches the organization id, its `name`, and the localisation line, so I left it rather than widen the diff unasked.

Separately: the validation report on this branch now has a working baseline and reports no new findings against main.

Changelog.txt
```diff
@@ -13,6 +13,8 @@ Content:
  - [ENG] Reworked the service economy and industrial collapse paths so the second one completed has halved rewards instead of being locked out
  - [ENG] Moved the Overseas Territories focuses to spread across the economic tree, combined the productivity focus branch into the economic tree and added the APOLLO Programme and Made Smarter Programme focuses
  - Custom MIO nations can now spend Military Industrial Organization Contract Tokens on generic catalog companies covering equipment their domestic MIOs lack. Also added UKR to this group (Issue #3973)
+ - Custom MIO nations can now spend Military Industrial Organization Contract Tokens on generic catalog companies covering equipment their domestic MIOs lack (Issue #3973)
```
TemplarGeneral: It created a duplicate line

TemplarGeneral (dismissed):

Changelog.txt
JasonBreen: You are right, removed.

Upstream amended that entry to mention UKR while this branch was open. I have been resolving the repeated Changelog.txt conflicts by keeping both sides, which is correct when each side is a genuinely new entry but wrong here: the two lines were the same entry before and after the amendment, so keeping both preserved the stale one.

The branch now carries only upstream's current wording. I also swept the whole file for near-duplicates rather than exact matches, since these two differed only by the trailing clause and an exact-match check missed them: no other groups found, and the same sweep is clean on #4391.

JasonBreen:
Same story as #4391: the failing `Mod tests (core)` is not from this branch.

The report reads `Duplicated Localisation Keys | 0 new | 1 error`, and states 1 pre-existing error remains. It is `CAT_agriculture_tech_research`, defined twice on consecutive lines in `MD_traits_l_english.yml` with identical values, inherited from main. This PR touches a localisation file, which pulls that group into scope and so inherits the error.

#4413 removes it.

This branch's total is 0 new errors.

JasonBreen:
> Would you be okay with changing UMM to generic_utility_vehicle_manufacturer and UAVision to GENERIC_skyforge_autonomous_systems, this one may need some loc changes?

ye, im okay with it

VitorFox82 (approved):

VitorFox82 (approved):

Issue #4310, opened 2026-09-14
[BUG] Balance issues (v 2.0.0)


millenniumdawnmod-hub:
> *Reported by **discord-sync** via [Discord](https://discord.com/channels/210890013334831104/1549112087157014670) on 2026-09-14*
> *Tags: Game Breaking*

**What happens**
By the 2020s, all of Europe unites into a single state, but the problem is that the AI ​​starts behaving overly aggressively. In my game, Europe declared war on half the world by 2026 and generated over 6400% world tension. It makes it impossible to play or develop properly; no single nation can stand up to a united Europe in the 2020s. Moreover, the USA does absolutely nothing to stop it—and even helps—turning the world into a madhouse.
Hoi 1.19
Md 2.0.0
Checksum 442a

**Where**
See Discord thread for details.

**Versions**
See Discord thread for details.

**To Reproduce**
See Discord thread for details.

**Operating System**
See Discord thread for details.

**Additional context**
Auto-created from a Discord forum post.

millenniumdawnmod-hub:
**Reply from Discord:**

Then disable the USOE end state rules?

millenniumdawnmod-hub:
**Reply from Discord:**

Yes, that’s possible, but I’m not talking about the unification of Europe itself—I’m talking about the AI’s excessive aggression. 6400% world tension—surely that shouldn't be happening.

millenniumdawnmod-hub:
**Reply from Discord:**

Without any other information this report is practically useless

millenniumdawnmod-hub:
**Reply from Discord:**

I would need to see how they started doing it and what path they are on otherwise

millenniumdawnmod-hub:
**Reply from Discord:**

What can I do for this?

millenniumdawnmod-hub:
**Reply from Discord:**

You need to show me actual information about what they did and how it started

millenniumdawnmod-hub:
**Reply from Discord:**

It started when Europe declared war on me (the USSR), but I signed a white peace with them; after that, Europe began declaring war on all the countries in the Middle East (Iran, Iraq, Saudi Arabia, etc.) and then moved on to Africa. Then came a spam of "aggressive action" notifications, driving world tension up to 6400%.

millenniumdawnmod-hub:
**Reply from Discord:**



[...]

PR #4391, opened 2026-09-15
Restore diplomatic AI anchoring to the formed United States of Europe

JasonBreen:
## Bottom line

The formed United States of Europe matches no `ai_strategy` in the mod and runs on bare engine defaults with no diplomatic anchoring. Relates to #4310, but does **not** claim to fix the aggression in that report.

## The bug

`common/ai_strategy/EUU.txt` holds Europe's only two strategy blocks, and both gate identically:

```
allowed = { EU_is_potential = yes }
enable  = { has_idea = EU_member }
```

`has_idea = EU_member` is a membership idea. A single unified state is not a member of itself, and `leaving_EU` removes it regardless. Both blocks carry `abort_when_not_enabled = yes`, so they switch off at unification, and nothing catches the fall: every `ai_strategy` in the mod is tag-gated and there are no `allowed = { always = yes }` blocks.

At unification the state therefore loses `befriend USA / CAN / AST / NZL`, `antagonize SOV`, and any conquer/contain targeting.

## The fix

Adds one strategy block restoring the befriend weighting, gated on:

```
allowed = { EU_is_potential = yes }
enable  = { is_USoE = yes }
```

`is_USoE` is a new trigger covering **all seventeen** USoE cosmetic tags. Forming the United States of Europe does not tag switch — `MD_USoE_decisions.txt` applies a cosmetic tag to whichever country forms it — and focuses later swap the base tag for ideological and dynastic variants (`USoE_com`, `USoE_green`, `USoE_air`, `USoE_army`, `USoE_navy`, `USoE_HAB`, `USoE_BOR`, …). A gate on the base tag alone would silently stop applying the moment one of those fired.

`antagonize SOV` is deliberately **not** restored. Spread across members that weighting is mild; concentrated on one superstate it plausibly contributes to the first war the reporter describes.

## What this does not do

It does not address the wargoal volume driving the aggression in #4310. The reporter's `Threat from actions` accrual at 8.00% per action with `Changing by 0.0% per month` compounds without decay, but the cause of the wargoal volume is a separate question from the missing AI strategy, and bounding a superstate's expansion properly needs the wider USOE AI review discussed on that ticket.

## Revision note

An earlier version of this PR gated on `allowed = { original_tag = EUU }`, which never matches, since the end state does not tag switch. That version was inert. This one is gated on the cosmetic tag instead.

BLUF


AngriestBird (changes requested):
This fixes only part of the problem. I am pretty sure the AI is getting given a ton of wargoals and as such it pusues a hyper aggressive foreign policy and then declares and gives a ton of world tension similar to the Roman one you solved. 

The conditionals you set for the AI strategy should not be allowed set to EUU and instead EUU potential and then has_country_flag USOE (you do not tag switch to EUU). 

I also think that you should have it where the USOE should be a little more passive in some scenarios so we  likely need to do a larger review over the AI for USOE. This ticket can serve as the justification for that work.

JasonBreen:
You are right, and it is worse than partial — as written my change is inert for the case in the report.

I checked the formation path and there is no tag switch anywhere. The United States of Europe is a **cosmetic tag** applied to whichever country forms it (`MD_USoE_decisions.txt:12`, `set_cosmetic_tag = USoE`), so `allowed = { original_tag = EUU }` never matches the superstate the reporter is describing. It would only ever apply to the EUU country itself, which is not what forms.

One correction to the suggested condition: there is no `USOE` country flag. Searching for `set_country_flag = USOE` and `has_country_flag = USOE` returns nothing. What actually exists is:

- `has_cosmetic_tag = USoE`, plus sixteen variants — `USoE_com`, `USoE_green`, `USoE_air`, `USoE_army`, `USoE_navy`, `USoE_HAB`, `USoE_BOR`, `USoE_HAN`, `USoE_HOH`, `USoE_NAP`, `USoE_ORA`, `USoE_SAV`, `USoE_WIN`, `USoE_WIT`, `USoE_GLU`, `USoE_BOR_ANJ`
- `has_country_flag = USoE_member` (`EU_USoE_decisions.txt:153`), which marks members rather than the formed state

So a working gate is `allowed = { EU_is_potential = yes }` with an enable on the cosmetic tag, and it needs to cover the variants rather than just the base tag, since the ideological and dynastic swaps are applied by focuses.

On the wargoals point, I have not verified that and would not want to guess at it. It is consistent with what the reporter describes — the "threat from actions" spam at 8% per action with no monthly decay — but the cause of the wargoal volume is a separate question from the missing AI strategy.

Given you want a broader USOE AI review and this PR does not do what it claims, I would rather not leave it open looking like a fix. Two options:

1. I rewrite it now as the corrected gate, scoped as "restore diplomatic anchoring to the formed USoE", explicitly not claiming to solve the aggression.
2. I close it and the ticket carries the whole piece of work.

Your call. I have not pushed anything further.

JasonBreen:
Rewritten and pushed.

The gate is now `allowed = { EU_is_potential = yes }` with `enable = { is_USoE = yes }`, where `is_USoE` is a new trigger in `99_EU_scripted_triggers.txt` covering all seventeen cosmetic tags rather than just the base one — the focus-driven ideological and dynastic swaps would otherwise silently drop the gate.

I did not use `has_country_flag = USOE` because no such flag exists; the cosmetic tag is the only thing that marks the formed state.

Title and description updated so the PR no longer reads as a fix for the aggression. It restores the anchoring only, and I have said explicitly that the wargoal volume needs the wider review.

JasonBreen:
The failing `Mod tests (core)` here is not from this branch.

The validation report on this PR says it plainly: **0 new errors**, 1 pre-existing error remains. The error is a duplicated localisation key, `CAT_agriculture_tech_research`, defined twice on consecutive lines in `MD_traits_l_english.yml` with identical values. It is inherited from main, and any PR whose changed files pull the localisation group into scope picks it up.

Opened #4413 to remove it, since it will keep tripping other PRs otherwise.

The two new warnings this report does attribute to the branch are both Japanese focus flags in `common/national_focus/05_japan.txt`, which this branch does not touch — they arrived through merging main in to clear the earlier conflicts.

This branch's own changes are `common/ai_strategy/EUU.txt`, `common/scripted_triggers/99_EU_scripted_triggers.txt` and `Changelog.txt`. Nothing to fix here.

AngriestBird (approved):

PR #4394, opened 2026-09-15
Restore the rescue-a-captured-general raid

JasonBreen:
## Bottom line

`descriptor.mod` replaces `common/raids` wholesale, and the mod ships nine of vanilla's ten raid files. The missing one defines `rescue_captured_general`, so the game logs a dangling-token error and players cannot rescue a captured general.

## The bug

Caught live in a 2002 campaign:

```
lexer.cpp:251: Tried to get dynamic token that does not exist: rescue_captured_general
```

`rescue_captured_general` is a vanilla raid defined in `common/raids/land_infiltration_raids.txt`. Because `descriptor.mod` carries `replace_path = "common/raids"`, any vanilla file the mod does not carry over is dropped entirely. That file is the only one missing.

Vanilla's `interface/lar_operations.gfx` and `localisation/english/military_raids_l_english.yml` still reference the raid, so the engine tries to resolve a token that no longer exists.

Player impact: with a general captured there is no rescue raid available, and the error re-fires whenever the game attempts to resolve it.

## Why this reads as an oversight rather than a removal

- The file has never existed in this repository, so it was not deleted at some point.
- The mod blanks raids it wants gone rather than omitting them — `land_infiltration_custom.txt` is a single line reading `# Intentionally left blank`. There is no such placeholder here.
- It is the only vanilla raid file absent; the other nine are all carried over.

## The fix

Adds `common/raids/land_infiltration_raids.txt` with vanilla's definition, unmodified. The raid keeps its `has_dlc = "Thunder at Our Gates"` gate, so it only appears for players who own that DLC, and it introduces no new art or localisation — both already ship in the base game.

BLUF


common/raids/land_infiltration_raids.txt
```diff
@@ -0,0 +1,220 @@
+types = {
+
+	rescue_captured_general = {
+
+		days_to_prepare = 60
+		days_re_enable = 60
+		speed_multiplier = 10
+		max_distance = 20
+		category = land_infiltration_raids
```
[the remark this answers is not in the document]
JasonBreen: Good catch, the fix was incomplete. Fixed, though not quite the way you suggested.

You are right that common/raids/categories is a replace_path and that land_infiltration_raids is undeclared, so the raid had an unresolved category.

Importing vanilla's declaration would have been the wrong repair, because the mod has not merely dropped that one category, it has replaced the whole set. It declares infastructure_strikes, military_resource_raids, naval_raids, special_forces_raids, drone_strike_raids and nuclear_raids, and keeps none of vanilla's land_infiltration_raids, air_raids, paratrooper_raids or naval_commando_raids. Adding vanilla's back would bolt a parallel scheme onto the mod's own.

The raid now uses special_forces_raids instead. It fits a commando rescue, shares the same intel_source = army, and already hosts the mod's own special forces raids. The raid keeps its own has_dlc = "Thunder at Our Gates" gate, so DLC gating is unchanged.

Also checked the other external references while I was in there: GFX_raid_target_icon_captured_general is present in tools/validation/vanilla_sprites.txt, so it resolves from the base game.

common/raids/land_infiltration_raids.txt
```diff
@@ -0,0 +1,220 @@
+types = {
+
```
AngriestBird, line 2: It's fine to include, but do we even have this feature enabled?

AngriestBird (dismissed):

common/raids/land_infiltration_raids.txt
JasonBreen: Yes, raids are very much live. The mod ships 370 raids across nine files in common/raids, has its own redesigned category set (infastructure_strikes, military_resource_raids, naval_raids, special_forces_raids, drone_strike_raids, nuclear_raids), and carries two raid game rules, allow_carrier_only_raids and rule_raid_doctrine_reach.

What was missing was only this one vanilla file. descriptor.mod replaces common/raids wholesale and nine of vanilla's ten files were carried over, so rescue_captured_general was the single orphan. The error is what surfaced it in a live 2002 game:

lexer.cpp:251: Tried to get dynamic token that does not exist: rescue_captured_general

It is gated behind has_dlc = "Thunder at Our Gates", so it only appears for players who own that DLC, and I put it in special_forces_raids rather than importing vanilla's category.

Confirmed fixed in game: after loading with this file present, the token error no longer appears and no raid or category errors replaced it.

Happy to drop it if you would rather the mod simply not have that raid — in that case the tidier alternative is a blank land_infiltration_raids.txt, matching how land_infiltration_custom.txt is handled, though that would leave the token error in place since vanilla's GUI still references it.

TemplarGeneral (approved):

PR #4361, opened 2026-09-15
Changed designs to BEL, POR, GRE, GUA, HOL, HUN, IDH, GLC, FLN, FLA

Qwerty483LOL:
Added V400 Icon and vehicle with the required technology to Portugal
Changed designs to: Portugal, Belgium (again) Greece, Guatemala, Netherlands (Holland), Hungary and US breakaway states

PR #4400, opened 2026-09-15
Share USA starting equipment variants with breakaways

### LINKED ISSUE #4392: [TASK] USA | Reduce code duplication in US breakaway histories
**Description**
A clear and concise description of the task. The main goal of these tasks are general purpose improvements.

Notable the USA and the breakaways all inherit the USA design variants for building purposes. We need to de-duplicate this. 

**Additional context**

AngriestBird:
## Bottom line

Share the 194 USA starting equipment definitions through shared scripted effects and reuse the DLC-appropriate catalogs from USA and all 17 breakaway histories. Preserve the breakaway-only M1126 Stryker and UH-60A designs while removing 182 duplicated inline variant blocks from 14 histories.

Closes #4392

## How

Technology inheritance and country-specific OOB selection remain unchanged. USA-only guards preserve the original production-license ordering without granting those licenses to breakaways.

USA MIOs now allow the USA and breakaway tags. Each organization is available only to a country that owns its assigned company state, so regional company access follows state ownership. The MIO validator now resolves direct scripted-trigger tag lists and revalidates dependent organizations and references when a scripted trigger changes.

Breakaway manufacturer assignments still need in-game verification before this draft is marked ready.

BLUF


AngriestBird:
## Bottom line

Branch updated from main (merge 6d1b54b913). Conflicts resolved in favor of the shared common library: all USA breakaway starting equipment stays in the shared catalog in common/scripted_effects/99_USA_main_scripted_effects.txt, and fixes on this branch are the priority.

Details:

- The four merge conflicts (FLA, GLC, HWI, IDH) came from #4361 editing the suspension lines inside the copied variant blocks that this PR removes. Kept the branch's removal. All three affected designs that exist in the shared catalog already carry #4361's updated values there (M2A3 Bradley, M270, M3A3 Bradley CFV all use tank_torsion_bar_suspension_light in the catalog), so no design change is lost.
- One heads-up: M1126 Stryker Combat Vehicle is a breakaway-only design (medium_tank_amphibious_chassis_2, not in USA's catalog). It is removed along with the copies, which is consistent with moving everything into the shared catalog. #4361's suspension change to it no longer has a target. If we want breakaways to keep an amphibious chassis 2 design, it would need to be added to the shared catalog explicitly.
- Remaining open item before marking this ready: in-game verification of breakaway manufacturer assignments (per the PR description). All CI gates pass with no new findings against the main baseline.

TemplarGeneral (approved):

PR #4406, opened 2026-09-15
Fix UK Commonwealth Economic Hegemony GDP Check (#4352)

### LINKED ISSUE #4352: [BUG] UK - Commonwealth Economic Hegemony Bug
> *Reported by **discord-sync** via [Discord](https://discord.com/channels/210890013334831104/1549385726422950028) on 2026-09-15*
> *Tags: General Bug*

**What happens**
I am playing as the UK.

The issue is around the attached with the hegemony for the commonwealth. It states that I need for economic more GDP than the other 3 combined. However, it is not doing GDP but GDP per capita. I need as the UK more GDP per capita than the 3 combined which is insane and not doable. since it is 373k as it is Australia, New Zealand and Canada combined.

I have the subscription on the store so I should be playing with all the DLC's, it is 2009 but this issue didn't exactly occur then just always there.

I am using windows 11 and playing version 2.0.0 of the mod. 

Checksum - I am not exactly sure where to find this so I hope it is right. Operation Postern v1.19.2.0a729 (442a)

**Where**

**Versions**

**To Reproduce**

**Operating System**

**Additional context**

TemplarGeneral:
### Changes
- The Uniting the Realm Economic Hegemony requirement now compares the United Kingdom's total GDP against the combined total GDP of Canada, Australia and New Zealand instead of GDP per capita, so the decision is actually reachable
- The Economic Hegemony checklist tooltip now shows both GDP figures in billions

Closes #4352


AngriestBird (approved):

AngriestBird (dismissed):
Long term we should probably move this into a scripted effect since it's duplicated code.

AngriestBird (approved):

PR #4409, opened 2026-09-15
Replace zombie spawn-delay missions with a hidden event

### LINKED ISSUE #4174: [BUG] Zombies
> *Reported by **discord-sync** via [Discord](https://discord.com/channels/210890013334831104/1548566762944208947) on 2026-09-13*
> *Tags: General Bug*

**What happens**
Why is zombie mode still not working? When I set the infection to start in a year...

**Where**

**Versions**

**To Reproduce**

**Operating System**

**Additional context**

AngriestBird:
## Bottom line

Delayed zombie spawn no longer depends on player countdown missions. A hidden `zombie.16` event on USA fires after 365/730/1095 days for the 1/2/3 year game rules, then runs the existing spawn effect.

## Why

The 1-3 year delay only armed missions on `is_ai = no` countries at `on_startup`. If that failed, ZOM never existed and weekly/monthly zombie on_actions never ran, so the horde never appeared.

## How

Removed the countdown missions and category. Startup now schedules one hidden event per delay option (same event, different `days`) or spawns immediately when delay is none.

Closes #4174

BLUF


common/on_actions/99_ZOM_on_actions.txt
TemplarGeneral, line 1: Why not move these to the standard yearly effects?

TemplarGeneral (changes requested):

TemplarGeneral (approved):

PR #4417, opened 2026-09-15
Clearing Validator Errors

MrP0tter:
Clearing some of the validator error spam to making using them eaiser


common/scripted_effects/00_antarctica_effects.txt
```diff
@@ -3028,7 +3028,7 @@ antarctica_apply_single_laboratory_iteration_reward = {
 
 		if = {
 			limit = { check_variable = { station_illegal_activities > 0 } }
-			add_tech_bonus = { name = antarctica_lab_tech_bonus bonus = 0.15 uses = 1 category = CAT_nuclear_weapons }
+			add_tech_bonus = { name = antarctica_lab_tech_bonus bonus = 0.15 uses = 1 }
```
Blazer135: [P1] Restore the target of the nuclear research rewards

This reward now has neither a category nor a technology selector, so it no longer targets nuclear research. CAT_nuclear_weapons is still declared in common/technology_tags/00_technology.txt and used by ballistic missile technologies. Please restore the category here and in the seven other add_tech_bonus blocks where this PR removes their only selector: Germany, Iran, North Korea (two), Poland, Thailand, and the generic atomic MIO.

BLUF

Blazer135 (commented):
1 findings from the verified review are attached inline. BLUF

PR #4423, opened 2026-09-16
Stop Iraq's Most Wanted from starting twice

JasonBreen:
## Bottom line

`iraqi_MNF.170` starts Iraq's Most Wanted twice when two of its four dispatch
routes both resolve. Adding `fire_only_once = yes` stops the second run.

## What goes wrong

`iraqi_MNF.170` is the opening event for the fugitive deck. Its option sets
`IRQ_hunt_baathists`, activates the `USA_fugitive_countdown` mission and queues
the six initialisation events that place each fugitive in a state.

It is dispatched from four places, all on a 400 day delay:

- `events/Iraq.txt:1452` (`iraq_war.1`)
- `events/Iraq.txt:1598` (`iraq_war.3`)
- `events/Iraq.txt:2238` (`iraq_war.9`)
- `events/Middle East Peace Plan.txt:519` (`iraqi_MNF.2`)

It carried no one-shot guard, so a campaign that reaches two of those routes
fires the whole opening sequence twice.

## Why it corrupts the minigame

The initialisation events are not idempotent. `iraqi_MNF.18` picks a
`random_owned_state` whose `limit` excludes states already flagged
`IRQ_qusay_here`, `IRQ_uday_here` or `IRQ_abid_here`, but not one already
flagged `IRQ_saddam_here`. A second pass therefore places the same fugitive in
a second state, and re-runs `TOP_import_target_location` and a second
`TOP_add_target_lead` of 40.

`activate_mission = USA_fugitive_countdown` also runs twice.

## Verification

Reproduced in a 2003 USA campaign. `game.log` records the option executing
twice in the same in-game hour:

```
8:00, 12 May, 2003: United States: iraqi_MNF.170.a executed
8:00, 12 May, 2003: United States: iraqi_MNF.170.a executed
```

The player sees the event popup twice and the card grid opens in a
double-initialised state.

Not changed: the four dispatch sites are left alone, and the asymmetric `limit`
in `iraqi_MNF.18` is left alone. With a single dispatch neither is reachable.

BLUF


JasonBreen:
Closing: this belongs in the fork, not here.

PR #4450, opened 2026-09-16
Let the Algerian arms MIOs build what their traits improve

### LINKED ISSUE #4436: [BUG] Algerian MIO's giving incorrect boosts
> *Reported by **discord-sync** via [Discord](https://discord.com/channels/210890013334831104/1549711676188925972) on 2026-09-16*
> *Tags: General Bug*

**What happens**
- current 2.0 version of MD 
- Checksum 442a
- Own all the DLC
- tag: ALG
- Start of the game
- Windows 11

there are 2 MIO's for Algeria
1. Khenchela Arms Manufacturing complex
2. Light Arnaments Manufacturer

Both of these MIO's have a lot of bonusses for different things like C&C equipment, ATGM's and MANPADS. Even tho you cant assign these companies to any of the the above, only to small arms.

And an additional question: is Algeria not supposed to be able to buy additional MIO's? Especially because they don't have a lot of really good companies

**Where**

**Versions**

**To Reproduce**

**Operating System**

**Additional context**

### Attachments

> *⚠ mirror failed, link expires in ~24h*

JasonBreen:
## Bottom line

Fixes #4436. Both Algerian MIOs inherit a trait tree covering six equipment
types but could only be assigned to small arms, so about half of each tree
could be paid for and never apply.

## What was wrong

`ALG_khenchela_arms_manufacturer` and `ALG_seriana_small_arms` take
`include = generic_infantry_equipment_organization`. That template declares:

```
equipment_type = {
    infantry_weapons_type  cnc_equipment_type  L_AT_Equipment
    AA_Equipment           artillery_equipment mio_cat_all_utils
}
```

and carries traits gated to each of them. Both Algerian MIOs then override
`equipment_type` with `{ infantry_weapons_type }`.

The override narrows what the MIO can be assigned to. It does not remove the
inherited traits. So the traits limited to command and control, ATGMs, MANPADS,
artillery and utilities were unreachable, which is what the report describes.

Both are widened to the six types the tree covers.

## One trait needed a limit of its own

Widening makes previously dead traits live, and that exposed
`ALG_seriana_trait_kalashnikov_expertise`, which grants `soft_attack` with no
`limit_to_equipment_type`. `soft_attack` has no base value on MANPADS or
command and control equipment, so the bonus would be inert there. It is now
limited to `infantry_weapons_type`, which is what Kalashnikov expertise should
affect anyway.

`validate_mios` reports no issues. Without that limit it reports thirteen inert
bonuses, and `upstream/main` reports none, so the check is meaningful here
rather than inherited.

## Not included

`VEN_cavim_manufacturer` has the same narrowing and was found by the same scan.
It is not fixed here: its own trait tree grants `hard_attack`, `soft_attack`,
`defense` and `breakthrough` across twelve traits with no equipment limits, so
widening it sets off twelve inert bonus errors that need those traits revisited
first. It was not the reported bug and deserves its own change.

Scanning every MIO in the mod, those three were the only ones with this shape.

BLUF


common/military_industrial_organization/organizations/MD_ALG_organizations.txt
```diff
@@ -17,7 +17,14 @@ ALG_khenchela_arms_manufacturer = {
 		}
 	}
 
-	equipment_type = { infantry_weapons_type }
+	equipment_type = {
```
TemplarGeneral: As copilot tells. If you want the MIO to just behave exactly as the include one, only put in that line and nothing else. If you put in a other block (like equipment_type) it will override the include

TemplarGeneral (dismissed):

common/military_industrial_organization/organizations/MD_ALG_organizations.txt
```diff
@@ -17,7 +17,14 @@ ALG_khenchela_arms_manufacturer = {
...
+		infantry_weapons_type
+		cnc_equipment_type
+		L_AT_Equipment
+		AA_Equipment
+		artillery_equipment
+		mio_cat_all_utils
+	}
```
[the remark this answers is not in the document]
JasonBreen: Deleted. Both local `equipment_type` blocks are gone, so the roster comes from `generic_infantry_equipment_organization` alone — the template block is byte-identical to what the override had grown to, so behaviour is unchanged and there is nothing left to drift. The trait limit stays: the inherited roster is still six types, so the Kalashnikov bonus has to remain pinned to small arms or it leaks onto the other five.

common/military_industrial_organization/organizations/MD_ALG_organizations.txt
JasonBreen: Done — removed the `equipment_type` block from both Algerian MIOs so the `include` is the only source. Net change is now two deletions plus one `limit_to_equipment_type` on the Kalashnikov trait, which is still needed to keep that bonus on small arms.

AngriestBird (approved):

PR #4451, opened 2026-09-16
Stop a Swiss party switch from destroying the political system

### LINKED ISSUE #4292: [BUG] Political System of Switzerland national spirit disappears when switching parties
> *Reported by **discord-sync** via [Discord](https://discord.com/channels/210890013334831104/1549056494828392488) on 2026-09-14*
> *Tags: BETA, General Bug*

**What happens**
When switching a party as Switzerland the national spirit "Political System of Switzerland" disappears whilst doing the path strengthen the federation. As you can see in the first and second images the highlighted national spirit disappears in the second image after switching the parties. Went from SVP to SP

Game version: v1.19.2
MD version: Beta v2.0.0
Checksum: 442a

**Where**

**Versions**

**To Reproduce**

**Operating System**

**Additional context**

JasonBreen:
## Bottom line

Fixes #4292. Switching party as Switzerland could permanently destroy the
Political System of Switzerland spirit, with no way to get it back.

## The race

`on_government_change` in `common/on_actions/MD_event_on_actions.txt` removes
all four `swiss_political_system` tiers and sets
`end_of_the_swiss_political_system`, a permanent flag, when:

```
NOT = { has_government = democratic }      # engine ideology group
NOT = { has_government = neutrality }      # engine ideology group
NOT = { nationalist_right_wing_populists_are_in_power = yes }   # ruling_party = 20, a script variable
```

Two of those read the engine's ideology group and the third reads MD's
`ruling_party` variable, which is set from script in
`00_MD_politicsview_scripted_effects.txt`. During a switch there is a window
where the variable already holds the new party while the engine still reports
the old group, and in that window all three clauses are true.

The reported case is SVP to SP. SVP is `nationalist` with `ruling_party = 20`;
SP is `SWI.socialism`, which `00_ideologies.txt` puts in `democratic`. Once
settled the condition cannot pass, which is why the end state looks correct.

The same condition is the `cancel` on all four idea tiers, and HOI4
re-evaluates `cancel` continuously, so the tiers were usually already stripped
before the on_action ran. The comment above that block half acknowledged it:
"tiers cancel daily on this same condition, so never gate on has_idea here
(usually already stripped)".

## The fix

`SWI_keeps_political_system` states the rule in `ruling_party` alone:

| Group | `ruling_party` |
|---|---|
| democratic | 0-3 |
| neutrality | 12-19 |
| national populism | 20 |

The set is exactly what the old condition intended, so behaviour on settled
governments is unchanged.

**Scope of the trigger, corrected after review.** It is used by the four idea
cancels only. The on_action stays on `has_government`, because `set_politics`
fires `on_government_change` before `ruling_party` is written, and reading the
variable there sees the old party: the tiers still cancel, but the flag is
never set and the news event waits for an unrelated change. An earlier revision
of this PR moved the on_action onto the trigger and introduced exactly that
regression; thanks to @TemplarGeneral for catching it.

The two read different sources on purpose. The on_action fires once, at the
instant of the change, and needs state that is current then. The cancels
re-evaluate every tick and settle on the variable. That is also why the race
bit the cancels rather than the on_action.

## Verification

`validate_common_mistakes` is clean. `validate_unused_scripted` does not flag
the new trigger, so all four call sites resolve. Encodings and tabs unchanged.

Not verified in game: reproducing the original needs a party switch mid-path.

BLUF


common/national_focus/05_switzerland.txt
```diff
@@ -2344,6 +2344,16 @@ focus_tree = {
 					add_idea = swiss_political_system2
 				}
 			}
+			else_if = {
+				# Stripped by the old government-change race. Give the tier back
+				# rather than quietly paying political power for nothing.
+				limit = {
+					NOT = { has_idea = swiss_political_system2 }
+					SWI_keeps_political_system = yes
+				}
```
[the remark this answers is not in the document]
JasonBreen: Correct, and it was a real hole in the recovery. The idea cancels on `is_subject = yes` OR failing `SWI_keeps_political_system`, but the recovery gates only checked the second half, so a Swiss subject on an allowed party would be granted a tier, have the end flag cleared, and lose it again on the same evaluation — while also forgoing the political power the else branch pays. All three gates now carry `NOT = { is_subject = yes }` so they mirror the cancel exactly.

tools/validation/validate_tech_categories.py
```diff
@@ -236,7 +236,7 @@ def validate_tag_definitions(self, known: FrozenSet[str]):
             ],
             "All technology categories are CAT_lowercase",
             "Technology categories not named CAT_lowercase:",
-            severity=Severity.ERROR,
```
AngriestBird, line 286: This is an erroneous check. These should be errors.

common/scripted_triggers/99_SWI_scripted_triggers.txt
```diff
@@ -1,3 +1,19 @@
+SWI_keeps_political_system = {
+	# The consensus system survives under the democratic parties (0 to 3), the
+	# neutral ones (12 to 19) and national populism (20). Both the idea cancels
```
AngriestBird: Comment doesn't need to be so verbose here.

common/ideas/05_switzerland_ideas.txt
```diff
@@ -6,11 +6,7 @@ ideas = {
 			cancel = {
 				OR = {
 					is_subject = yes
-					AND = {
-						NOT = { has_government = democratic }
-						NOT = { has_government = neutrality }
-						NOT = { nationalist_right_wing_populists_are_in_power = yes }
-					}
+					NOT = { SWI_keeps_political_system = yes }
```
AngriestBird, line 9: Can you update the tooltips for the ideas (descriptions) and explain the cancel conditions for players?

AngriestBird (dismissed):

tools/validation/validate_tech_categories.py
JasonBreen: Agreed, and the change is gone — both files are back to main. The severity downgrade was the wrong response to those findings: the 39 they report are real and #4467 fixes them at the source, so nothing needs relaxing. It also had no business in a Switzerland PR.

common/scripted_triggers/99_SWI_scripted_triggers.txt
JasonBreen: Trimmed to one line covering the part the code cannot say for itself, the party index ranges. The race-condition explanation is in the PR body instead.

common/ideas/05_switzerland_ideas.txt
JasonBreen: Added. All four ideas share the swiss_political_system loc key, so the cancel conditions are now spelled out once in swiss_political_system_desc: the consensus holds only while Switzerland is not a subject and government stays with the democratic parties, the neutral bloc or the national populists.

TemplarGeneral (changes requested):
I found three things:

1. Regression, medium — [MD_event_on_actions.txt:578](vscode-webview://01ukn97jr7jkja19t4u0tnnfa5jvujauoivgucin34omlh9nl6h4/common/on_actions/MD_event_on_actions.txt#L578): paths that call set_politics before writing ruling_party (generic extremist civil-war decisions, e.g. [00_political_decisions.txt:398-408](vscode-webview://01ukn97jr7jkja19t4u0tnnfa5jvujauoivgucin34omlh9nl6h4/common/decisions/00_political_decisions.txt#L398-L408)) hit on_government_change with the old ruling_party, so the on_action skips. Tiers get stripped by the cancel, but switzerland.2 never fires and the end flag is never set; the news event then fires late on the next unrelated group change. The old has_government clauses covered this path.
2. Design, medium — [05_switzerland.txt:2347](vscode-webview://01ukn97jr7jkja19t4u0tnnfa5jvujauoivgucin34omlh9nl6h4/common/national_focus/05_switzerland.txt#L2347) (also else_if siblings at 2397, 4137): the recovery can't distinguish a race-stripped save from a legitimate loss. After a real switch to communism (news fired, flag set), returning to an allowed party and completing the focus re-grants the tier and clears the flag, undoing the permanent-loss design. The comment claims a narrower scope than the code implements.
3. Loc, low — [MD_focus_SWI_l_english.yml:474](vscode-webview://01ukn97jr7jkja19t4u0tnnfa5jvujauoivgucin34omlh9nl6h4/localisation/english/MD_focus_SWI_l_english.yml#L474): footer says government rests with "the chancellery"; Switzerland's executive is the Federal Council. Tier 4's _desc also lacks the footer.

JasonBreen:
## Bottom line

All three fixed in `960be7587d`. You were right on the regression, and the cause is worse than the decision path you cited.

**1. Regression.** Confirmed at `00_political_decisions.txt:398-408`: `set_politics` runs first and `set_variable = { ruling_party = 4 }` is nine lines later inside `hidden_effect`, so `on_government_change` fires with the old value. The cancels still stripped the tiers because they re-evaluate continuously, but the flag was never set and `switzerland.2` waited for an unrelated change. The on_action is back on `has_government` and is now byte-identical to main apart from a comment saying why it must not read the variable there.

Worth recording: `nationalist_right_wing_populists_are_in_power` is itself `check_variable = { ruling_party = 20 }`, so main's on_action was already mixing engine state with the variable and had a narrow version of this same bug for a national-populist predecessor. My change widened it to every democratic and neutral predecessor. I left the narrow case alone rather than widen this PR — an all-engine rewrite would need `has_government = nationalist`, which covers 20 to 23 while the rule is 20 only, so it would change behaviour.

**2. Recovery.** Removed all three blocks. You are right that they cannot tell a race-stripped save from a legitimate loss, and `AGENTS.md` settles it anyway: "Development builds may invalidate saves. Do not add legacy migration support." That is what they were. `05_switzerland.txt` is now identical to main.

**3. Localisation.** Federal Council, not chancellery. `swiss_political_system4_desc` gets the same footer — it is one of only two tiers with its own description key, tiers 2 and 3 fall through to the base one.

## What is left

Just the actual fix: `SWI_keeps_political_system` in `99_SWI_scripted_triggers.txt`, used by the four idea cancels in `05_switzerland_ideas.txt`, plus the two description footers. The cancels are where #4292 bit, because they re-evaluate every tick and the mixed engine and variable clauses disagreed mid-switch.

The on_action and the cancels now read different sources on purpose: the on_action fires once at the instant of change and needs engine state that is current then, while the cancels settle a tick later on the variable. The PR body claimed they could not disagree; that was wrong and I have corrected it.

BLUF

TemplarGeneral (approved):

Issue #4426, opened 2026-09-16
validate_tech_categories fails on main: 20 categories carried by no technology, 233 live references


JasonBreen:
## Bottom line

`Mod tests (targeted-b)` is red on `main`. `validate_tech_categories` reports 57
errors, all in `common/technology_tags/00_technology.txt`, and the validator
scans repo-wide rather than changed files only, so every open PR touching
`common/` inherits the failure.

20 of the 57 are not stale declarations. They are categories that **no
technology carries** while 233 references across the mod still point at them,
so those research bonuses and AI weights currently grant nothing. That is the
exact failure the validator was written to catch.

## Reproducing

```
python tools/validation/validate_tech_categories.py
✗ VALIDATION COMPLETE - 57 ERROR(S) in 1 file
```

Byte-identical result on a clean `upstream/main` checkout, and the same 57 on
unrelated PRs #4386, #4391 and #4421, whose diffs do not touch technology.

Introduced by the category rework, #4343 and #4355.

## Breakdown

| Category | Count | Notes |
|---|---:|---|
| `tech-category-missing-loc` | 35 | additive fix, no balance effect |
| `tech-category-unused` | 20 | blocking, see below |
| `tech-category-name-format` | 2 | `CAT_OLV`, `special_forces_doctrine` |

## The blocking 20

MD ships one doctrine technology file, `common/technologies/land_doctrine.txt`,
and it uses `CAT_land_doctrine` and `CAT_legacy_doctrines`. There is no air or
naval doctrine technology file, but both categories are still declared and
still referenced.

References outside `common/technology_tags/`:

| Category | refs | ai_focuses | decisions | ideas |
|---|---:|---:|---:|---:|
| `CAT_air_doctrine` | 75 | 4 | 3 | 0 |
| `CAT_naval_doctrine` | 63 | 0 | 3 | 0 |
| `CAT_nuclear_weapons` | 19 | 0 | 1 | 5 |
| `CAT_western` | 10 | 2 | 1 | 0 |
| `CAT_unconventional` | 8 | 3 | 0 | 0 |
| `CAT_green_water_navy` | 8 | 1 | 0 | 0 |
| `CAT_abm` | 8 | 0 | 0 | 0 |
| `CAT_equipment_doctrines` | 7 | 3 | 0 | 0 |
| `CAT_eastern` | 6 | 3 | 0 | 0 |
| `CAT_tech_doctrines` | 5 | 3 | 0 | 0 |
[...]

PR #4458, opened 2026-09-16
Name the technology categories that had none

JasonBreen:
## Bottom line

`validate_tech_categories` reports 57 errors on `main`, and because it scans
repo-wide rather than changed files, it fails `Mod tests (targeted-b)` on every
PR that touches `common/`. This clears the 35 that can be cleared without a
content decision, taking the file to 22.

## What this fixes

35 categories had no English name, so the tooltip showed the raw token: a
research bonus naming one of them read as gibberish. Adds the 46 keys the
validator names and nothing else.

Names come from the authoring comments in `00_technology.txt` where they exist
(`CAT_art_ammo # Artillery Ammo`, `CAT_m_sub # Nuclear Missile Submarine`), and
Title Case of the token otherwise. Categories that already had a base key, such
as `CAT_jeune_ecole_two` and `CAT_m_sub`, only gain the `_research` form.

## What is left, and why it is not here

**20 categories used by no technology.** These are not stale declarations: 233
references across `ai_focuses`, `decisions`, `ideas`, `national_focus` and
`scripted_effects` point at them, so those research bonuses and AI weights
currently grant nothing. That is the failure this validator exists to catch.

Classifying them against the categories technologies actually carry:

- `CAT_nuclear_weapons` is a real mis-pointing. MD's nuclear technologies use
  `CAT_nuclear`, and live idea modifiers name the dead one, for example
  `common/ideas/05_united_kingdom.txt:2149` granting 0.06 and
  `common/ideas/05_netherlands.txt:2583` applying -10. Repointing it would fix
  a bonus that silently does nothing, but it also turns an inert modifier live,
  which is a balance change and not mine to make in a CI fix.
- The other 19 are doctrine categories for systems the mod no longer ships.
  `common/technologies/` contains only `land_doctrine.txt`, using
  `CAT_land_doctrine` and `CAT_legacy_doctrines`; there is no air or naval
  doctrine file at all. Deleting them and their dead references is
  behaviour-preserving, since nothing carries them today, but it removes
  authored content such as `cat_jeune_ecole_two_desc` and should be a
  maintainer's call.

**2 tokens that are not `CAT_lowercase`.** `special_forces_doctrine` is also in
the unused 20, so renaming it alone changes nothing. `CAT_OLV` is a live
category with 74 references, and renaming it is the trap: eight non-English
locales define `CAT_OLV_research`, and AGENTS.md says not to edit non-English
localisation, so the rename would orphan the key in all eight with no way for
me to repair them. That needs a decision about whether the naming rule should
admit acronyms.

Tracked in #4426.

BLUF


TemplarGeneral:
This is (going to) completly reworked in an PR thats already open. So this isn't needed. Sorry mate

PR #4452, opened 2026-09-16
Rework technology categories: all remaining tabs and final sweep (#4250)

TemplarGeneral:
## Bottom line
- Closes out #4250: every remaining technology tab is on the new category layering and the tag file is regrouped by folder.

## Why
- Abbreviated, near-duplicate and unused tags made research bonuses and MIO categories unreliable; several focus and MIO bonuses pointed at tags no tech carried.
- Missile and artillery ammunition categories only exist with Gotterdammerung / No Step Back, so bonuses on them need a DLC guard.

## How
- Armor, artillery/AA/AT, aircraft, naval, naval modules, industry, electronics, missiles and space: every tech carries `CAT_military`/`CAT_civilian`, one folder tag, one branch tag and at most one group tag; sibling tags with identical tech sets are merged; remaining tags renamed to lowercase full words with loc consolidated in `MD_techs_l_english.yml`.
- Legacy no-DLC files (armor, artillery, anti_air, fixed_wing, bombers, non_got_missiles) carry the same tags as their DLC counterparts.
- Final sweep: shared group tags added, `00_technology.txt` regrouped by folder, tags no tech carries deleted, MIO `research_categories` and `ai_focuses` weights retargeted.
- 22 `add_tech_bonus` sites on DLC-only categories wrapped in `has_dlc`; POLSA, Perun and Pivdenne retargeted to `CAT_orbital_launch_vehicles`, Polonez and PULS to `CAT_self_propelled_artillery`; stale `$CAT_missile$` placeholder fixed.
- Validator remainder: 39 errors, all on doctrine tags in `00_technology.txt`, which were not part of the #4250 backlog and belong to a separate doctrine pass.

Closes #4250

BLUF

Issue #4445, opened 2026-09-16
[TASK] Cleanup old Doctrine tech tags


TemplarGeneral:
**Description**
A clear and concise description of the task. The main goal of these tasks are general purpose improvements.

As an followup from #4250 :

Categories are used for doctrines. So we can't remove them. We need to rework them so every grand_doctrine and sub_doctrine based on track.

TL;DR: We need to rework the doctrine categories like i did the tech categories.

**Additional context**
Add any other context or screenshots here.

PR #4459, opened 2026-09-16
Remove rewards for doctrines the mod no longer has

JasonBreen:
## Bottom line

Stacked on #4458. Together they take `validate_tech_categories` from **57
errors to 1**, which is what has been failing `Mod tests (targeted-b)` on every
PR touching `common/`.

This half removes 164 reward blocks that grant nothing, and repoints one that
should have worked all along.

## Rewards for doctrines the mod does not have

19 categories are carried by no technology, yet 164 blocks still name them:

| Dead category | Rewards naming it |
|---|---:|
| `CAT_air_doctrine` | 61 |
| `CAT_naval_doctrine` | 58 |
| 15 others | 45 |

`common/technologies/` ships one doctrine file, `land_doctrine.txt`, carrying
`CAT_land_doctrine` and `CAT_legacy_doctrines`. There is no air or naval
doctrine tree at all, so 119 of the 164 have nothing to point at.

A concrete case: `CZE_joint_air_exercises` offers three countries a 20 percent
air doctrine cost reduction with two uses, and delivers nothing.

**Removing them changes nothing in play**, because nothing is granted today
either way. What changes is that a focus stops displaying a reward that never
worked.

## Blocks that mix dead and live categories

14 blocks name a live category alongside a dead one. Those keep the block and
lose only the dead lines. Three carry `CAT_land_doctrine` and one carries
`CAT_sam`; those bonuses do work and are untouched. Verified by listing every
category line the diff removes and checking none is a live category.

## CAT_nuclear_weapons is repointed, not removed

MD's nuclear technologies carry `CAT_nuclear`. 19 references named the empty
`CAT_nuclear_weapons`, including North Korean, Iranian, Iraqi and German
weapons focuses, an Antarctic lab bonus, two MIO research category lists, and
idea modifiers in `05_united_kingdom.txt` (0.06) and `05_netherlands.txt`
(-10). Those now grant what they always claimed to.

This one is a behaviour change: inert modifiers become live. It is the only
dead category with a real equivalent, which is why it is treated differently
from the 19.

## Verification

- `validate_tech_categories`: 57 to 1
- `validate_common_mistakes`, `validate_focus_tree`, `validate_mios`: no issues
- `validate_events`: 12457 warnings before and after, unchanged
- Brace balance asserted per file before and after every edit

## The remaining 1

`CAT_OLV` is not `CAT_lowercase`. It is a live category with 74 references, and
eight non-English locales define `CAT_OLV_research`. AGENTS.md says not to edit
non-English localisation, so renaming would orphan the key in all eight with no
sanctioned repair. Worth deciding whether the rule should admit acronyms
instead. Tracked in #4426.

BLUF


TemplarGeneral:
Wait until the technology changes are merged. You then need to recheck. Also i think this PR closes this task then: https://github.com/MillenniumDawn/Millennium-Dawn/issues/4445

JasonBreen:
Superseded by #4467. Both of these were built before the category rework (#4452) landed, and rechecking against current main changed two answers: CAT_nuclear_weapons now carries 28 technologies, and CAT_abm is gone from the tree entirely. Repointing nuclear and stripping abm would have broken live references. #4467 is rebuilt on current main and takes the validator from 39 errors to 0.

Issue #4250, opened 2026-09-14
[TASK] Rewrite Technology Categories


TemplarGeneral:
**Description**
A clear and concise description of the task. The main goal of these tasks are general purpose improvements.

I want to rewrite the whole tech tags. In the basics we need to have the following:

- one tag per folder
- one tag per branch
- then sparse some around for categories (like combination of Naval Attack Modules, Naval Modules, Air Weapons, etc.)
- Helicopters is like already done in the new style
- every tag needs a _research variant and we need to make a validator for that 
- we standardize the category id's with using full names for clarity

**Additional context**
Add any other context or screenshots here.

## Backlog (one PR per folder)

Rules for every tab: each tech carries `CAT_military`/`CAT_civilian` + one folder tag + one branch tag (+ at most one group tag where a consumer needs it). Tag ids are `CAT_` + lowercase full words. Every tag has `CAT_x` and `CAT_x_research` English loc keys in `MD_techs_l_english.yml`. Sibling tags merge into one wherever no consumer needs the split. Legacy no-DLC files get the same tags as their DLC counterpart in the same PR. `validate_tech_categories.py` gates all of this.

- [x] Validator: `CAT_lowercase`, `_research` loc, unused tags, tech-file / MIO / ai_focuses references (#4343)
- [x] Global `CAT_Military` / `CAT_Civilian` rename (#4343)
- [x] Helicopters `air_techs_folder` (#4343)
- [x] Infantry `infantry_folder` (infantry.txt) (#4343)
- [x] Armor `nsb_armour_folder` + legacy `armour_folder` (NSB_armor.txt, armor.txt)
- [x] Artillery / AA / AT `artillery_folder` (NSB_artillery.txt, NSB_anti_air.txt, anti_tank.txt, artillery.txt, anti_air.txt)
- [x] Aircraft `bba_aircraft_folder` + legacy `fixed_wing_folder` / `bomber_folder` (BBA_aircraft.txt, fixed_wing.txt, bombers.txt)
- [x] Naval ships `naval_folder` (naval.txt)
- [x] Naval modules `naval_module_folder` (naval_modules.txt)
- [x] Missiles `cruise_missiles_folder` + legacy `missile_non_got_folder` (ballistic_missiles.txt, cruise_missiles.txt, missile_defense.txt, non_got_missiles.txt); includes `CAT_OLV` rename
- [x] Space `space_folder` (space.txt)
[...]

PR #4487, opened 2026-09-17
Map legacy tech categories in the validator and remove unsupported doctrine categories (#4472, #4445)

### LINKED ISSUE #4472: [VALIDATION] Map Old tech categories to new ones
**Description**
A clear and concise description of the requested validation suite.

The validator suggests the replacement for old technology categories but these are way off. We should introduce a table with old > new so the validator can suggest proper suggestions that make sense

**Additional context**

TemplarGeneral:
## Bottom line
- The tech category validator now names the real replacement when a PR references a tag renamed in #4250, instead of a string-similarity guess.
- The doctrine research categories from #4491 are removed: NCNS doctrines cannot carry a `categories` block, so every `add_doctrine_cost_reduction` now targets its doctrine folder and doctrine-targeted `add_tech_bonus` blocks became doctrine cost reductions.

## Why
- After the rework, an old tag like `CAT_inf_wep` or `CAT_OLV` still fails CI, but the `did you mean` hint pointed at an unrelated tag because the new names share few characters with the old ones.
- #4491 added `categories = { CAT_x }` to every grand doctrine and subdoctrine and 35 `CAT_*_doctrine` tags, but doctrines only support `mastery = { categories }` (sub-unit categories). None of it did anything in game, and `add_doctrine_cost_reduction` only accepts the doctrine folder token (`land_doctrine`, `equipment_doctrine`, `naval_doctrine`, `air_doctrine`, `special_forces_doctrine`).

## How
- Added an old-to-new table (122 rows) in `validate_tech_categories.py`, derived from the retargets the rework itself applied and cross-checked against what each old tag's techs carry today.
- Hint order: case-insensitive exact match (`CAT_Military` -> `CAT_military`), then the legacy table when its target exists, then the previous difflib fallback.
- Tests: legacy hint, mixed-case hint, and a real-repo check that every table target is a current tag and no key still is.
- Stripped the 182 doctrine-level `categories` blocks from `common/doctrines/`; `mastery` sub-unit categories are untouched.
- `add_doctrine_cost_reduction`: 306 `category =` lines now use the folder token (track tags map to their folder; `CAT_special_forces_equipment` in 9 doctrine reductions maps to `special_forces_doctrine`).
- 44 `add_tech_bonus` blocks that pointed at a doctrine tag are now `add_doctrine_cost_reduction` (`bonus` -> `cost_reduction`, same name and uses).
- Removed the 35 doctrine tags from `00_technology.txt`, their loc keys, the `categories` block on the 6 legacy `land_doctrine.txt` stub techs, and 23 dead doctrine weights in `ai_focuses`.
- Reverted the validator's doctrine-carrier logic and tests from #4491; the legacy-hint table stays.
- Fixed a Windows-only `_collect_files` dedupe bug (mixed separators from overlapping globs) that made `validate_simplifications` report every finding twice and failed the Windows tools job on this PR and on #4491's. `os.path.normpath` before the seen-check, plus a regression test.
- Not touched: 8 `add_doctrine_cost_reduction` blocks in `05_singapore.txt` that use equipment tags (`CAT_corvettes`, `CAT_naval`, `CAT_frigates`, `CAT_infantry`, `CAT_helicopters`, `CAT_aircraft`, `CAT_air_drones`); they predate #4491 and probably should be `add_tech_bonus`.

Closes #4472
Reverts the doctrine part of #4491 (#4445)

BLUF


tools/tests/validation/tech_categories_test.py
```diff
@@ -1,10 +1,12 @@
 """Tests for validate_tech_categories."""
 
+from shared.paths import REPO_ROOT
 from validate_tech_categories import (
+    _LEGACY_CATEGORIES,
     Validator,
     _brace_span,
-    _doctrine_category_bodies,
```
Blazer135, line 6: [P1] Restore the import still used by the doctrine test

This import edit removes _doctrine_category_bodies, but test_doctrine_category_bodies_skips_nested_mastery_block still calls it at line 326. The PR's CI fails with NameError on that test; Ruff also reports F821 for the same missing name. Restore the import so the existing regression test can execute.

BLUF

Blazer135 (changes requested):
Restore the import removed by this PR: an existing test still uses it, causing the confirmed NameError and F821 failures. The inline comment identifies the deletion. This is a PR-introduced failure; unrelated CI warnings are excluded.

**Automated tests must pass before merge.**

BLUF

tools/tests/validation/tech_categories_test.py
TemplarGeneral: Fixed

Blazer135 (approved):

PR #4489, opened 2026-09-17
Standardise Venezuela AI path game rule

TemplarGeneral:
### Changes
- The Venezuela AI behaviour rule now offers Historical, Export the Revolution, Democratic Opposition, Corporate State, Free-Market Republic, Military Junta, Random Path and No Path; the old Chavismo Continues option is folded into Historical
- Historical AI keeps Chavismo in power, takes the peaceful Bolivarian focuses, holds and survives the recall referendum, and never forms the Gran Colombia faction, launches the world revolution or hands the country to the Cartel of the Suns; it still dismantles the FARC, since that is the only cure for the FARC spirit
- The opposition and junta paths reach power through an AI-only popularity ramp, so the democratic, PDVSA and officer branches are actually playable by the AI
- Venezuela's war focuses no longer fire while it is already at war, Curacao is never demanded from a NATO Netherlands and a war on the United States needs superior strength; the Dutch refusal event has a stand-down option
- Combat-capacity and army-reform focuses are weighted up when the country is threatened or still carries its outdated army
- No leader walker: the roster is an undated successor list and the existing 2013 event already seats Maduro

Part of #3162


common/decisions/Venezuela.txt
```diff
@@ -375,3 +375,120 @@ VEN_hydroelectric_decision_category = {
...
+				has_global_flag = VEN_DEMOCRATIC_OPPOSITION_FOCUS_PATH
+				VEN_ai_pdvsa_path = yes
+			}
+		}
+
+		available = {
+			has_civil_war = no
+			has_completed_focus = VEN_our_president
+			western_liberals_are_in_power = no
+			check_variable = { party_pop_array^2 > 0.5 }
```
Blazer135, line 432: [P1] Stop opposition decisions after the PDVSA takeover

Both PDVSA rules keep this repeatable decision enabled after VEN_pdvsa_takeover installs party 14, or VEN_establish_free_market installs party 16. The preceding rally also keeps boosting party 2 above 50%, so the AI can spend another 150 PP to reinstall the western liberals. change_ruling_party_effect clears the coalition, making the corporate/libertarian branch requirements fail; the completed takeover cannot restore its government. Disable both opposition decisions for PDVSA paths once VEN_pdvsa_takeover is complete.

BLUF

common/national_focus/05_venezuela.txt
```diff
@@ -6658,7 +6993,19 @@ focus_tree = {
...
-		ai_will_do = { base = 10 }
+		ai_will_do = {
+			base = 10
+			modifier = {
+				factor = 0
+				has_war = yes
+			}
+			modifier = {
+				factor = 0
+				HOL = { has_idea = NATO_member }
```
Blazer135: [P2] Preserve junta progression when skipping NATO Curaçao

This prevents AI Venezuela from completing VEN_curacao_question while the Netherlands remains in NATO, but the focus has no bypass. Both VEN_northern_expansion and VEN_nearby_islands_control require it, and the later junta regime branches depend on those focuses. The normal NATO scenario therefore stalls the remaining junta tree. Add an AI-only bypass for the NATO condition, or let downstream prerequisites accommodate the skipped demand.

BLUF

events/Venezuela.txt
```diff
@@ -70,6 +80,11 @@ country_event = {
 					factor = 0.75
 					check_variable = { party_pop_array^2 > 0.20 }
 				}
+				modifier = {
+					add = 200
```
Blazer135: [P2] Make historical recall survival deterministic

Adding 200 to the survival weight does not ensure the historical outcome: the recalled branch still has weight 40 or more. Even with Mission Florentino and no opposition modifiers, historical AI has a 40 / (300 + 40), or 11.8%, chance of being recalled and triggering MD4_election_campaign.0. This contradicts the PR's promised historical behavior. Zero the recalled branch's weight for historical AI, preserving randomness for other cases.

BLUF

Blazer135 (changes requested):
Changes are needed to preserve the selected PDVSA and junta paths and make the promised historical recall outcome reliable. Three defects are documented inline. Reviewed the diff and shared political effects; in-game behavior was not tested.

**In-game testing required before merge.**

BLUF

Blazer135 (approved):

PR #4493, opened 2026-09-17
Standardise Yugoslavia AI path game rule

TemplarGeneral:
### Changes
- The Yugoslavia AI behaviour rule now offers Historical, Greater Serbia, Balkan Federation, Socialist Yugoslavia, Kingdom of Serbia, Random Path and No Path; the misspelled Milocevic options and the old Default are gone
- Historical AI lets the Bulldozer Revolution topple Milošević, then seats Đinđić, Koštunica, Tadić, Nikolić and Vučić on their real dates, never holds the castle, never restores the crown and neither claims nor recognises Kosovo
- The three Milošević paths keep him in power through the protests and hold the Socialist Party against the next election until their fork focus disables elections
- The kingdom path grows the monarchists after the revolution and seats them, so the Karađorđević branch is actually reachable by the AI
- War and ultimatum focuses no longer fire while Serbia is already at war, and army-building focuses weigh up while the country is threatened
- The report still lists International Isolation and the Special Unit spirit as uncured on some paths: the first is cancelled by any neutral or democratic government, which the revolution installs, and the second is a bonus Milošević keeps by design

Part of #3162


events/Serbia.txt
```diff
@@ -4897,3 +4897,55 @@ country_event = {
...
+		log = "[GetDateText]: [Root.GetName]: event Serbia.400"
+		if = {
+			limit = { date > 2017.5.30 }
+			set_variable = { Neutral_Autocracy_leader = 2 }
+			set_temp_variable = { rul_party_temp = 13 }
+		}
+		else_if = {
+			limit = { date > 2012.5.30 }
+			set_variable = { Neutral_Autocracy_leader = 1 }
+			set_temp_variable = { rul_party_temp = 13 }
```
Blazer135, line 4925: [P1] Allow historical leader handovers after the radical split

If ser_dem.5.a has set SER_radical_divided, these 2012/2017 handovers cannot install Nikolić or Vučić. Both corresponding branches in set_leader_SER require that flag to be absent, while set_leader first kills the current leader. The reachable historical handover therefore produces a generated leader instead of the intended one. Make these dated handovers install their intended leaders even after the radical split.

BLUF

events/Serbia.txt
```diff
@@ -4897,3 +4897,55 @@ country_event = {
...
+			set_variable = { Neutral_conservatism_leader = 0 }
+			set_temp_variable = { rul_party_temp = 14 }
+		}
+		else = {
+			set_variable = { liberalism_leader = 0 }
+			set_temp_variable = { rul_party_temp = 2 }
+		}
+
+		if = {
+			limit = { NOT = { is_in_array = { ruling_party = rul_party_temp } } }
```
Blazer135: [P2] Compare the ruling party as a scalar

ruling_party is a scalar variable, not an array, so this membership test cannot recognize that the selected party already rules. The event consequently calls change_ruling_party_effect, which clears gov_coalition_array, and resets elections even for a same-party succession. Use check_variable = { ruling_party = rul_party_temp } so the existing leader-only branch can run.

BLUF

common/decisions/Serbia.txt
```diff
@@ -2164,3 +2164,118 @@ KOS_int_recognition = {
...
+					has_completed_focus = SER_turn_back_to_communist_ideals
+				}
+			}
+			check_variable = { party_pop_array^20 > 0.25 }
+		}
+
+		complete_effect = {
+			log = "[GetDateText]: [Root.GetName]: Decision SER_socialists_retake_the_country"
+			set_temp_variable = { rul_party_temp = 20 }
+			change_ruling_party_effect = yes
```
Blazer135, line 2276: [P2] Restore Milošević when the recovery decision retakes power

When an election removes Milošević, this recovery decision restores party slot 20 but does not restore him. change_ruling_party_effect calls set_leader, whose slot-20 Serbian branch creates Boško Obradovic once Milošević is no longer the ruling leader. The recovery mechanism therefore abandons the advertised Milošević path when it activates. Explicitly restore Milošević here, or preserve his leadership through the election transition.

BLUF

Blazer135 (changes requested):
Changes are needed in the historical leader handovers and Milošević recovery decision. Three concrete defects are documented inline. Reviewed the diff and shared political effects; in-game behavior was not tested.

**In-game testing required before merge.**

BLUF

events/Serbia.txt
TemplarGeneral: fixed

events/Serbia.txt
TemplarGeneral: fixed

common/decisions/Serbia.txt
TemplarGeneral: fixed

Blazer135 (approved):

PR #4523, opened 2026-09-17
Share per-file reads across variables validator sections (#4521)

### LINKED ISSUE #4521: Share per-file reads across variables validator sections
## Context

#4518 baseline (7 cold samples via validation_timing_report) leaves the core batch as the span driver: validate_variables runs 211.9s single-threaded, spread across sections (check_variable 33s, redundant focus flags 30s, dynamic modifiers 25s, treasury 22s, math precision 19s, scripted-trigger 17s, clamp 17s). Each section runs its own pool pass that re-reads, comment-strips, and blank-strings the same files. A needle pre-filter on the shared available scan measured ~zero (skipped files are small) and was reverted; the cost is repeated full Python-level passes over the large files, not file count.

## What

- Read each candidate file once per validator run and share the stripped/blanked text across the section scans (math, orphan-money, treasury, clamp, available, tooltips, dynamic modifiers, flag syntax), instead of one read plus strip plus blank pass per section.
- Keep worker and pool structure, CPU split, cache namespaces, strictness, and every finding identical. No worker-default, batch-rebalance, scheduling, or checkout change bundled in.
- Prove findings parity on the fixture suites plus a clean-tree run (empty findings diff) before claiming anything.

## Acceptance

- Before/after section timings from compatible cold CI samples (median/range, same workload); no speedup claimed without a measured comparison.
- Fixture suites plus full tools tests green; clean-tree findings unchanged.
- One focused PR; follow-up validators (events spread, simplifications single-scan) stay separate issues.

AngriestBird:
## Bottom line

Closes #4521. One pool pass over the union file set reads each candidate file once and shares the stripped and blanked text across all nine variable sections, instead of one read plus strip plus blank per section. Findings are unchanged on a clean tree (3 warnings without redundant flags, 95 with). Full tools suite green (5140 passed, 3 skipped).

## Why

The core batch span driver was repeated full Python-level passes over the same large files. Each section ran its own pool pass with its own read, comment strip, and blank pass.

## How

The fused worker calls the same `_scan_*_text` helpers the standalone workers use, gated by a per-file mask so each file runs exactly its own sections. Flag syntax keeps its naive strip. Clamp ranges harvest in the same pass and resolve parent-side. Worker count, pool reuse, cache namespaces, and strictness are untouched. Section timings collapse into one shared-scan timing by design; compare totals before and after (13.3s to 9.9s on 4 workers, 58.9s to 48.5s single-threaded with redundant flags, local `--no-cache` runs). Other validators stay separate issues per the task scope.

BLUF


TemplarGeneral (approved):

tools/validation/validate_variables.py
```diff
@@ -629,6 +621,66 @@ def collect_clamp_ranges(
...
+    except Exception:
+        return [], [], []
+    return _scan_clamp_harvest_text(cleaned)
+
+
+def _extract_clamp_checks(cleaned: str, rel: str) -> List[Tuple[str, str, int, int]]:
+    checks: List[Tuple[str, str, int, int]] = []
+    for pattern in (_CHECKVAR_SHORT_RE, _CHECKVAR_LONG_RE):
+        for m in pattern.finditer(cleaned):
+            line = cleaned[: m.start()].count("\n") + 1
```
Blazer135, line 645: [P2] Avoid rescanning the file prefix for every clamp comparison

The shared worker now calls `_extract_clamp_checks` for every common/events file, and this line slices and counts the entire preceding text for every `check_variable`, including variables that have no harvested clamp. Previously those variables were rejected by the range lookup before line counting. This adds quadratic work to the shared scan: in a synthetic fixture, extraction took about 0.10s for 1,500 checks, 0.43s for 3,000, and 2.21s for 6,000. Please compute line offsets once with the existing `compute_line_offsets` helper and use `line_for_offset` for each match, so the read-sharing optimization does not add repeated full-prefix scans. BLUF

Blazer135 (commented):
One performance follow-up from the review, recorded inline after merge. BLUF

PR #4558, opened 2026-09-18
Fix microchip import typo and weekly production oscillation

### LINKED ISSUE #4367: [BUG] Microchip Production dissapearing
> *Reported by **discord-sync** via [Discord](https://discord.com/channels/210890013334831104/1549436187884920893) on 2026-09-15*
> *Tags: General Bug*

**What happens**
Playing as Iran The microchip production goes to 0 half a month and then it recovers for the other half. I try to have and extended technological metal surplus but that doesn´t solve it. Its clearly, as shown in the screenshot, that I have production of microchips.
Operation Postern v1.19.2.0.a729 (d245)
Millenium Dawn v2.0.0
All DLCs owned and no other mods active

**Where**

**Versions**

**To Reproduce**

**Operating System**

**Additional context**

### LINKED ISSUE #4556: [BUG] Microchip Typo
Bug: Microchip production always shows deficit even when resources are sufficient — typo in 00_microchip_effects.txt

In the microchip_update block, there is a typo in the variable name resource_improted@tungsten — it should be resource_imported@tungsten. This causes the import check for tungsten to always evaluate to 0, even when tungsten is being imported.

The affected code (in microchip_update):

text
AND = {
    check_variable = { resource_produced@tungsten = 0 }
    check_variable = { resource_improted@tungsten = 0 }
}
Should be:

text
AND = {
    check_variable = { resource_produced@tungsten = 0 }
    check_variable = { resource_imported@tungsten = 0 }
}
Impact: When a country does not produce tungsten domestically but imports it, the game incorrectly treats imported tungsten as zero. This triggers the "no tungsten available" condition, which sets country_microchip_production_var to the full amount of microchips produced (resource_produced@microchips). As a result, the advanced_production_modifier consumes 100% of microchip output, creating an artificial deficit even though all required resources are available.

AngriestBird:
## Bottom line

Microchip production counted imported tungsten as zero (`resource_improted`), which also made output flip off every other week when plants had no domestic tech metals. Composite tech AI had the same class of typo (`resources@composites`). The loc validator now errors on unwritten `[?name]` reads and on `name@target` prefixes that are neither documented dynamic variables nor written by script. `check_variable` / `has_variable` reads of unwritten names are reported too.

## Why

The weekly update gated on current microchip output. A full restriction drove output to zero, so the next pulse cleared it. The controller now runs while plants exist, uses nominal plant output for a full halt, and skips Taiwan's chip-race ratio at zero production. Kabylia resistance loc now reads state 385.

Closes #4556
Closes #4367

BLUF


tools/validation/validate_localisation.py
```diff
@@ -1177,6 +1222,34 @@ def validate_variable_references(self):
             category="loc-unwritten-variable",
         )
 
+    def validate_targeted_dynamic_variables(self):
+        """name@target in script must be a documented dynamic var or a write."""
+        self._log_section("Checking name@target dynamic variables...")
+
+        known = self._script_written_variables() | _engine_loc_vars(self.mod_path)
```
Blazer135, line 1385: [P2] Normalize written targeted variables without localisation exclusions

The new error-level check reuses a write index built through `_loc_var_name`, which deliberately drops names shorter than three characters and names resembling three-letter country tags. A valid `set_variable = { pp@ROOT = 1 }` followed by `check_variable = { pp@ROOT > 0 }` therefore records only `pp@ROOT` as written, while the scanner checks `pp`, and reports both lines as unknown dynamic variables. I reproduced this with the PR's validator; `pressure@ROOT` passes the same fixture. Extract the prefix of script writes independently of localisation-specific filtering and cover short targeted names in a regression test.

BLUF

localisation/english/MD_focus_ALG_l_english.yml
```diff
@@ -1299,8 +1299,8 @@
  algeria_tebboune.1.c: "Elect Abdelkader Bengrina"
  ALG_benflis_tt: "§YAli Benflis§! becomes President of Algeria, leading the liberal reformists."
  ALG_bengrina_tt: "§YAbdelkader Bengrina§! becomes President of Algeria, representing the Islamist opposition."
- kabyle_resistance_info: "The Kabyle region is experiencing unrest. Resistance level: [?state_resistance@438|%Y]"
- kabyle_resistance_effects: "Monitor Kabylia for rising separatist support. Current resistance: [?state_resistance@385|%Y]"
+ kabyle_resistance_info: "The Kabyle region is experiencing unrest. Resistance level: [?385.resistance|%Y]"
+ kabyle_resistance_effects: "Monitor Kabylia for rising separatist support. Current resistance: [?385.resistance|%Y]"
```
Blazer135, line 1303: [P2] Display resistance without multiplying its value by 100

`385.resistance` is the state's resistance value on the 0–100 scale (the generated dynamic-variable documentation uses `resistance > 50`), but `|%Y` formats a 0–1 fraction as a percentage. A state at 25 resistance will consequently display 2500% in this tooltip. Use ordinary numeric formatting with a literal percent sign, or the non-scaling percentage formatter, in both changed strings.

BLUF

tools/validation/validate_localisation.py
```diff
@@ -580,6 +583,41 @@ def process_txt_for_var_writes(args: Tuple[str]) -> Set[str]:
...
+)
+
+
+def _scan_targeted_var_text(text: str, basename: str) -> List[Tuple[str, str, int]]:
+    """(prefix, file, line) for every name@target that is not a flag token."""
+    if "@" not in text:
+        return []
+    out: List[Tuple[str, str, int]] = []
+    for number, line in enumerate(text.split("\n"), 1):
+        if "@" not in line or _FLAG_LINE_RE.search(line):
```
Blazer135, line 642: [P2] Exclude flag tokens instead of skipping their entire line

This skips every targeted-variable read on a line containing any flag operation. For example, `if = { limit = { has_country_flag = chip_policy check_variable = { resource_improted@tungsten = 0 } } }` produces no error, whereas placing the flag and variable check on separate lines correctly reports the typo. I reproduced both cases with the PR's validator. Exclude only the flag's value so compact trigger blocks still receive the new typo check, and add a mixed flag/variable regression case.

BLUF

Blazer135 (commented):
## Bottom line

Three issues to address: valid short targeted-variable names fail validation, Kabylia resistance is displayed at 100 times its value, and flag operations suppress adjacent targeted-variable checks. Details and suggested fixes are inline.

Both validator issues were reproduced with the PR's code. Existing CI passed; gameplay and tooltip rendering were not tested in-game.

BLUF

Blazer135 (approved):

PR #4560, opened 2026-09-18
Fix Arab Spring lottery deadlock blocking Syria civil war

AngriestBird:
## Bottom line

Player Syria can sit at 300+ Arab Spring strength for a decade with the revolt chain never firing. The monthly lottery sets arab_spring_active even on silent (90-branch) months, consuming the fire-only-once protest mission; a later re-pick then leaves the flag stuck and blocks every future pick for all Arab nations.

## Why

- Lottery picks now skip nations already active or already resolved.
- The protest flag is set only when the trigger event actually fires.
- The trigger-nation mission is repeatable, so a fizzled start retries instead of deadlocking.
- A high-unrest player Syria (strength above 249, post-2011, at peace, pre-global wave) starts the chain directly instead of waiting on the lottery.

## How

- common/on_actions/MD_on_actions.txt: pick exclusions, flag moved into the event branch.
- common/decisions/Arab Spring.txt: removed fire_only_once from arab_spring_protests_1.
- common/on_actions/99_SYR_on_actions.txt: player-Syria fallback trigger.
- Changelog.txt: one Bugfix line under v2.0.1.

Pre-commit and MD content validators pass on the changed files.

BLUF


Blazer135 (commented):
## Bottom line

The Syria fallback can start overlapping protest missions. The inline comment includes a suggested fix that preserves the existing single-trigger-nation check and sets the active flag when scheduling the event.

BLUF

common/on_actions/99_SYR_on_actions.txt
```diff
@@ -202,6 +202,20 @@ on_actions = {
...
+					is_ai = no
+					date > 2011.1.1
+					check_variable = { arab_spring > 249 }
+					NOT = { check_variable = { ruling_party = 12 } }
+					NOT = { has_country_flag = arab_spring_active }
+					NOT = { has_country_flag = AS_arab_spring_completed }
+					NOT = { has_global_flag = global_arab_spring_started }
+					has_war = no
+				}
+				country_event = { id = arab_spring.0 days = 1 }
```
Blazer135, line 224: [P2] Preserve the single active trigger nation in the Syria fallback

This checks only Syria's active flag, whereas the global lottery refuses to start while any country in `global.arabic_countries` has `arab_spring_active`. If Tunisia is already running its initial protests, high-unrest player Syria can start `arab_spring_protests_1` alongside it. When Tunisia accepts reform and `arab_spring.12` sets `global_arab_spring_started`, Syria also satisfies `arab_spring_protests_2`'s activation conditions. The existing 60-day mission has no cancellation condition, so both missions can run and the first can force the boiling-point event before the new 90-day response period expires.

Apply the same array-wide active-country guard to this fallback, and set the active flag when scheduling the event so the two start paths cannot select competing trigger nations on the same tick:

```suggestion
					NOT = { has_country_flag = arab_spring_active }
					NOT = {
						any_of_scopes = {
							array = global.arabic_countries
							has_country_flag = arab_spring_active
						}
					}
					NOT = { has_country_flag = AS_arab_spring_completed }
					NOT = { has_global_flag = global_arab_spring_started }
					has_war = no
				}
				set_country_flag = arab_spring_active
				country_event = { id = arab_spring.0 days = 1 }
```

BLUF

Blazer135 (approved):

PR #4607, opened 2026-09-19
Rename Communist-State identifiers to communist_state

AngriestBird:
## Bottom line

Renames the `Communist-State` subideology and every identifier derived from it to `communist_state`, per the snake_case naming standard in the code stylization guide. Pure rename, no gameplay changes.

## What changed

345 files, 2349 insertions / 2351 deletions:

- Subideology token in `common/ideologies/00_ideologies.txt` (slot 4, communism)
- Leader trait `emerging_Communist-State` → `emerging_communist_state` (definition in `common/country_leader/00_traits.txt` plus all `create_country_leader` blocks in 169 history files)
- Leader-rotation variable `Communist-State_leader` → `communist_state_leader` (451 hits in 89 `*_political_leaders.txt` files; this also unifies the snake_case spelling PHI already used since PR #41)
- Coalition decision IDs `Communist-State_invite` / `Communist-State_remove` → `communist_state_invite` / `communist_state_remove`
- Scripted localisation hooks and defined-text names (`Communist-State_L`, `_L_desc`, `_L_icon`, `_valid`, `_pop_election`, `Gov_Communist-State_loc/_icon`, `Communist-State_cannot_invite`)
- English loc keys for all of the above (626 occurrences in 27 files)
- One prose fix: `MD_focus_SUB_l_english.yml` now reads "Requires Armenia's Communist party..." so the player-facing text did not become a raw identifier
- Tools: `PARTY_SLOT_NAMES` slot 4 in `tools/shared_utils.py`, stale examples in validator comments, and hyphen-regression test fixtures switched to synthetic `Test-State` names (the charset keeps `-`; hyphenated focus IDs and namelists still justify it)
- Docs: `.claude/docs/party-loc-reference.md` (slot table, case note, icon exception) and `docs/src/content/resources/scripted-effects-reference.md` (slot table)

Not touched: gfx sprite names (`GFX_generic_Communist_State_small`, `GFX_TAG_Communist_State` etc. were already underscored, so no .dds renames), `party_pop_array` numeric slots, historical changelogs, and `docs/dist` build output.

## Known follow-ups

- **Non-English localisation keeps the old keys.** Per the localisation guardrail this PR touches English only, so the 9 other languages still carry `TAG.Communist-State`-style keys until the next translation sync ("Regular update") picks up the renames. Party names in those locales will show raw keys in the meantime.
- The leader-rotation counter and the leader trait are not preserved in existing saves (unset counter resets rotation; saved characters lose the trait). Consistent with the policy that development builds may invalidate saves; no migration support added.
- The same subideology family still has other non-snake_case names (`Nat_Populism`, `Western_Autocracy`, `neutral_Social`, ...). Same mechanical treatment would apply if wanted.

## Validation

- `validate_party_loc.py --all`: 6597 warnings on this branch vs 6598 on a pristine main worktree. All category counts identical except slot-case (8 → 7, one pre-existing case mismatch resolved by the rename); zero new warnings, and no pairing-category warning mentions `communist_state`.
- `python -m pytest` in `tools/`: all tests pass except `config_drift_test.py::test_tools_quality_checks_are_wired_in_precommit_and_ci`, which fails identically on clean main in this environment (installed coverage config default), so it is not a regression of this change.
- Commit hooks (MD Fix Styling, MD Validate) passed.
- Repo-wide grep: `Communist-State` remains only in non-English localisation (untouched by design), historical changelogs, `docs/dist`, and the new Changelog line that describes the rename.

BLUF


.claude/docs/party-loc-reference.md
```diff
@@ -47,8 +47,8 @@ documented at `MD_politics_view_parties_l_english.yml:38-61`.
 | 22   | `Nat_Autocracy`              | `nationalist` | Military Junta                |
 | 23   | `Monarchist`                 | `nationalist` | Monarchists                   |
 
-Case matters: `Communist-State` is hyphenated, `neutral_Social` and `oligarchism` are the
-only two with a lowercase first letter.
+Case matters: `neutral_Social` and `oligarchism` are the only two with a lowercase first
```
TemplarGeneral, line 50: Is this needed and like how it is? Because more start with lowercase like anarchist

TemplarGeneral (approved):

PR #4624, opened 2026-09-19
Ukraine Decisions/Events/OOBs update

hopekeeper:
### Changes
- Added 7 new Events for BTR-3 UAE tender procurement;
- Added 3 new Decisions for T-72 modernizations;
- Updated a bunch of Decisions for Equipment modernization and development;
- Updated description and triggers for "Mriya burnt" event;
- Cleaned some stylization errors in "MD_regimental_support.txt" and "MD_units_l_english.yml" files;
- Added 2 more desctext images for BTR-3 Decisons tooltips, 1 for desctext images for Su-25M1K Decisons tooltip and 3 more for dmal levents;
- Added specific "tempalte_counter" for 4 Ukrainian templates, that used incorrect ones per default;
- Updated Ukrainian "Airborne Brigade" template to be capable for parachute-dropping (as it is in reality);
- Added 2 techs for start of Ukraine game;
- Changed the order of Decisions Categories for Ukraine;
- Moved shared EU focus for Ukraine to not overlap.


AngriestBird (changes requested):
## Bottom line

Requesting changes. The new BTR-3 tender chain has a chain-ending softlock pattern (clickable options whose effects silently no-op on a `fire_only_once` event), a conditional double-fire bug in the factory logic, and the reworked Mriya event is a timing-fragile one-shot. Validators are clean, but none of these are validator-covered. Details with repro paths below.

### High

**1. BTR-3 tender options dead-click and permanently brick the chain (`events/Ukraine.txt`, `ukraine_develop_btr3.1`)**

Option A wraps its whole effect in `if = { limit = { UAE opinion > 39 ... } }` and option B in `if = { limit = { opinion < 40 ... } }`, but neither option has a `trigger =`, so both are always clickable and the event is `fire_only_once`. Any wrong pick consumes the event with zero effect: no variant, no win/lose flag. Since `UKR_develop_btr3e1` visibility requires the win OR lose flag (`common/decisions/Ukraine.txt`, `UKR_develop_btr3e1` visible block), the chain stalls permanently with the 180-day decision already spent. Concrete trap: the decision is available at UAE opinion > 19 but the win needs > 39, so taking it at 20-39 and clicking "And we won!" is a guaranteed no-op. Same pattern in `.4` option B (license path needs opinion > 49 inside an `if`; a click below that eats the license opportunity forever). Suggested fix: move the conditions to option `trigger =` blocks so invalid options are hidden (and AI-visible) instead of silent no-ops.

**2. Win path destroys a factory, then can double-fire progress + failure events (`ukraine_develop_btr3.1.a`, `.5.a`, `.6.a`, `.7.a`)**

In `.1.a` the two follow-up `if`s are sequential with a `remove_building = { arms_factory }` between them. If Kharkiv has exactly 1 arms factory: the first `if (arms > 0)` queues `.2` and removes the factory, then the second `if (arms < 1)` is also true and queues `.5`. Both fire. Same shape in `.5.a` (queues `.2` + self-loops `.5`), `.6.a` (`.3` + `.6`), `.7.a` (`.4` + `.7`). At game start Kharkiv has 2 (`history/states/696-Kharkiv.txt`), so the standard path is safe, but the failure mode is deterministic once reached, and the factory destruction itself is never disclosed in any tooltip/desc. Suggested fix: `else_if`/`else` (or snapshot the count first), and document the factory cost.

### Medium

**3. Mriya event is a timing-fragile one-shot (`ukraine_tech.306`)**

Now requires SOV to control province 16514 or 3543 at fire time, but it fires once (57 days after the 2022 milestone). If the frontline is elsewhere that day, or SOV doesn't exist, the event fizzles silently: no retry, no flag, choice and industry bonus lost. The new desc also says the Russians *failed* at Hostomel, which sits oddly with gating on SOV holding it. Suggested fix: retry scheduling while neither flag is set, or gate the options instead of the event.

**4. No `ai_chance` on any of `ukraine_develop_btr3.1`–`.7`**

AI UKR takes the decision (`ai_will_do = 10`) then picks event options blind. On `.1` roughly half of AI picks do nothing (see #1); the `.5`–`.7` renegotiate/terminate loops are also coin flips. Fix with `ai_chance` plus the `trigger =` gating from #1.

**5. Broken tooltip sprite name (`interface/MD_decisions_desc.gfx`)**

`GFX_UKR_desctext_modernize_sov_btr70_kba2_for_tooltipp` (triple-p) never matches the loc reference `£UKR_desctext_modernize_sov_btr70_kba2_for_tooltip`. The .dds file exists; only the sprite name is wrong, so the BTR-70M-KBA-2 decision shows a missing tooltip image. Note: `validate_gfx_references.py` passes anyway, so this pattern isn't covered.

### Low / nits

- New `UKR_develop_btr3rk` is visible to 2012 but available only to 2010 (dead 2-year window); license-flag line left commented out in `.4.b` (write-only flag anyway).
- `.6.a`/`.7.a` reuse `name = ukraine_develop_btr3.5.a` (renders fine, identical text, but copy-paste).
- Typos in new strings: "embarassing" + "lower then 50" (`.5`/`.6`/`.7` descs), "lossses / avantpost / cresult / trully / aupleasant" (Mriya desc).
- Unmentioned balance changes: Vilkha cost 1.0→0.3, LIT opinion 29→39, MiG-29MU1 capped <2015, Su-27 window narrowed to 2008–2020, new Yak-130/stealth prerequisite flags (no save migration; in-flight chains lose access), T-72 diesel-tech gate moved UA1→120 (sensible), Skif→"UAT Skif" rename, Crimea `add_core_of = UKR` x4 (looks correct, just undocumented).

### PR-body accuracy: partially accurate

True: 7 BTR-3 events, Mriya rework, template counters (4 confirmed), 2 starting techs (both IDs valid), category reorder, shared-focus move. Wrong: "3 new Decisions for T-72 modernizations" — the three T-72 blocks are reordered in place, no new T-72 decision; the actually-new decision (BTR-3RK) isn't mentioned. "Airborne Brigade now capable for parachute-dropping" — not effected by this diff: line battalions (`Mech_Air_Inf_Bat`, `can_be_parachuted = yes`) and all regimental support units were already parachutable. Missing entirely: Crimea cores, Simferopol rename + `fire_only_once`, all plane-decision gating/date caps, Vilkha/LIT changes.

### Verification

Reviewed in a clean worktree at PR head `ac69057`. `validate_decisions` (1 pre-existing unrelated warning), `validate_events`, `validate_focus_tree`, `validate_gfx_references`, `validate_oob_units`, `validate_history`, `validate_common_mistakes` clean; `validate_localisation` has 90 warnings, none in UKR files. All 15 new .dds files present; H_AT upgrades, `nsb_barrel_conversions`, `special_forces_tech_3` all exist; category priorities (116–120) collision-free. T-72 ID rotation is reference-safe.

BLUF

Blazer135 (commented):
6 findings from the verified review are attached inline. BLUF

common/national_focus/05_ukraine.txt
```diff
@@ -9351,7 +9351,8 @@ focus_tree = {
 
 		cost = 5
 
-		prerequisite = { focus = UKR_upa focus = UKR_suppress_western_and_bander }
+		prerequisite = { focus = UKR_upa }
+		prerequisite = { focus = UKR_suppress_western_and_bander }
```
Blazer135: [P2] Restore the either-path Arestovich prerequisite

These separate prerequisite blocks require both UKR_upa from the fascist path and UKR_suppress_western_and_bander from the Vitrenko path. A normal playthrough of either route no longer unlocks UKR_recruit_arestovich, despite its availability explicitly admitting both routes. Restore the original single prerequisite block containing both focus IDs so either one satisfies the gate.

BLUF

events/Ukraine.txt
```diff
@@ -8092,6 +8098,1637 @@ country_event = {
...
+
+	option = {
+		name = ukraine_modernize_sov_equipment_brdm_2_modernization.1.a
+		log = "[GetDateText]: [This.GetName]: ukraine_modernize_sov_equipment_brdm_2_modernization.1.a executed"
+
+		ai_chance = { base = 58 }
+
+		if = {
+			limit = {
+				date < 2007.01.01
```
Blazer135: [P2] Hide ineligible BRDM bids before they can consume the chain

All four bid choices are clickable, but this Mykolayiv option does nothing after 2007 or without control of Mykolayiv. The other two bids likewise keep date/postponement conditions inside their effects, so they do nothing at the 2000 start. The initiating decision is one-use, and only choosing the separate postponement option schedules another opportunity. Move these conditions to each option's trigger so choosing an ineligible bid cannot silently end the chain; retain postponement when no bid qualifies.

BLUF

events/Ukraine.txt
```diff
@@ -8092,6 +8098,1637 @@ country_event = {
...
+country_event = {
+	id = ukraine_modernize_sov_equipment_brdm_2_modernization.2
+	title = ukraine_modernize_sov_equipment_brdm_2_modernization.2.name
+	desc = ukraine_modernize_sov_equipment_brdm_2_modernization.2.desc
+	picture = GFX_ukraine_btr3u_uae_tender_presentation #TO-DO
+
+	is_triggered_only = yes
+	trigger = {
+		original_tag = UKR
+		control = 1086 #Mykolayiv
```
Blazer135: [P2] Use the supported state-control trigger

This country-event guard uses control = 1086, while the supported trigger is controls_state = 1086, already used in the preceding bid option. The same unsupported spelling occurs in the .3, .4, and .5 continuation guards. Change all four to controls_state so the Mykolayiv control requirement is actually expressed with a valid trigger. The exact parser behavior of the current entries has not been tested in game.

BLUF

events/Ukraine.txt
```diff
@@ -8092,6 +8098,1637 @@ country_event = {
...
+	id = ukraine_develop_btr3.2
+	title = ukraine_develop_btr3.2.name
+	desc = ukraine_develop_btr3.2.desc
+	picture = GFX_ukraine_btr3u_uae_tender_procurement
+	is_triggered_only = yes
+	fire_only_once = yes
+
+	trigger = {
+		original_tag = UKR
+		has_country_flag = {
```
Blazer135: [P2] Test the BTR-3 winning flag with valid syntax

The block form of has_country_flag requires flag = <name>; this bare token does not supply that required field. The .1.a win path sets UKR_develop_btr3_uae_tender_win_flag and schedules this procurement event, but the malformed check prevents a valid test of that flag. Use the simple scalar form: has_country_flag = UKR_develop_btr3_uae_tender_win_flag.

BLUF

common/decisions/Ukraine.txt
```diff
@@ -8307,6 +8305,48 @@ UKR_modernize_sov_category = {
...
+			industrial_capacity_factory = -0.004
+		}
+
+		days_remove = 120
+
+		complete_effect = {
+			log = "[GetDateText]: [Root.GetName]: Decision ukraine_modernize_sov_equipment_brdm_2_modernization"
+
+			hidden_effect = {
+				set_temp_variable = { treasury_change = -0.01 }
```
Blazer135: [P3] Deduct the displayed political-power cost

The new decision checks for more than 5 political power and displays cost_0_01_pp_5, but complete_effect only deducts treasury. ai_hint_pp_cost does not spend political power. Add add_political_power = -5 beside this treasury deduction so the player pays the advertised one-time cost.

BLUF

Changelog.txt
```diff
@@ -39,9 +39,12 @@ v2.0.1
...
+ - [UKR] Added more equipment decisions for Ukraine
+ - [UKR] Reworked starting BTR-3 development decision - made an UAE tender procurement an extensive Event chain
  - Added more equipment decisions for Ukraine
  - [USA] Operation Iraqi Freedom now puts the USA at war with Iraq when the decision completes, so the invasion and its phased war events can run (Issue #4663)
  - [JAP] Reworked Japan's earthquakes, electronics war and zombie firms: quakes now repeat across the seven earthquake-prone regions until each is retrofitted and Tohoku is rebuilt, the East Asian Electronics Competition tracks Competitive Pressure that rises over time and with China's semiconductor progress, the Galapagos Question grants a two-year premium, each zombie firm case resolves on its own terms, Article Nine decisions cost and grant different things, the consumption tax cut is temporary again, and Demographic Pressure cuts the starting workforce by 12% so Japan no longer opens at 10% unemployment (Issue #2743)
+>>>>>>>>> Temporary merge branch 2
```
Blazer135: [P3] Remove the temporary merge markers

This adds a literal temporary merge marker to the published changelog, with two more near lines 212 and 214. Please remove all three markers and consolidate this PR's added UKR content/bugfix notes into exactly one BLUF line under the current version, as AGENTS.md requires. Preserve unrelated existing entries.

BLUF

common/decisions/Ukraine.txt
hopekeeper: Thanks, implemented it.

events/Ukraine.txt
hopekeeper: This one resolved

events/Ukraine.txt
hopekeeper: This one fixed

common/national_focus/05_ukraine.txt
hopekeeper: Ok, reverted this change.

